PR #70, opened 2022-11-30
Add models for the `/_remotestore/_restore` API endpoint

### LINKED ISSUE #69: [FEATURE] Add models related to Remote-backed Storage
There is only a single very simple operation needing modelling:
- [ ] Restore `/_remotestore/_restore`: https://opensearch.org/docs/latest/opensearch/remote/#restoring-from-a-backup

It must be communicated via the model and any derivative generated clients, that this API is still currently experimental and as such may be subject to change and is not recommended for production use.

As this is a truly "missing" API, in that it does not exist in the clients yet, this will enable us to spike and experiment on generating only missing subsets of clients.

Xtansia:
### Description
Add's models for the `/_remotestore/_restore` API endpoint

Screen Shot 2022-12-07 at 10 45 26 AM

As seen [here](https://github.com/opensearch-project/OpenSearch/blob/5500114cc55d94bb8a7699f3db6870d83dc92445/server/src/main/java/org/opensearch/action/admin/cluster/remotestore/restore/RestoreRemoteStoreResponse.java#L75), `accepted` may be set in the response instead of `remote_store`, unsure of exact causal path however.

### Issues Resolved
Resolves #69 

### DIFF
--- a/model/remotestore/restore/operations.smithy
+++ b/model/remotestore/restore/operations.smithy
@@ -7,6 +7,7 @@
 $version: "2"
 namespace OpenSearch
 
+@unstable
 @externalDocumentation(
     "OpenSearch Documentation": "https://opensearch.org/docs/latest/opensearch/remote/#restoring-from-a-backup"
 )
--- a/model/remotestore/restore/operations.smithy
+++ b/model/remotestore/restore/operations.smithy
@@ -1,0 +1,20 @@
+// SPDX-License-Identifier: Apache-2.0
+//
+//  The OpenSearch Contributors require contributions made to
+//  this file be licensed under the Apache-2.0 license or a
+//  compatible open source license.
+
+$version: "2"
+namespace OpenSearch
+
+@externalDocumentation(
+    "OpenSearch Documentation": "https://opensearch.org/docs/latest/opensearch/remote/#restoring-from-a-backup"
+)
+
+@suppress(["HttpUriConflict"])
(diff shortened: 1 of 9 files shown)


harshavamsi:
@Xtansia can we also add tests here?

Xtansia:
@harshavamsi Have added a test case

nhtruong (approved):

model/remotestore/restore/operations.smithy
```diff
@@ -0,0 +1,21 @@
...
+
+@unstable
+@externalDocumentation(
+    "OpenSearch Documentation": "https://opensearch.org/docs/latest/opensearch/remote/#restoring-from-a-backup"
+)
+
+@suppress(["HttpUriConflict"])
+@http(method: "POST", uri: "/_remotestore/_restore")
+@documentation("Restore one or more indices from a remote backup.")
+operation PostRemoteStoreRestore{
```
VachaShah, line 18: Are the names of the operation and structures inline with what we would want in the generated code? Or do they need to be changed?

Xtansia: They're named to match the naming scheme of the rest of the specs in the repo. Naming (or encoding friendly names as metadata similar to #73) will likely be a bigger issue, but hard to solve in advance until we get to the point of actually generating parts of a client from these.

VachaShah (approved):

Issue #82, opened 2023-03-16
Generate missing models from legacy spec


nhtruong:
This spawns from https://github.com/opensearch-project/opensearch-api-specification/issues/80

Basically use [legacy spec](https://github.com/opensearch-project/OpenSearch/tree/42058aa76a18f531ceaac7deb4e94944cf02f463/rest-api-spec/src/main/resources/rest-api-spec/api) to generate smithy models so we don't have to fill out 308 operations by hand.

The translator code can be found here: https://github.com/nhtruong/opensearch-smithy-spec-translator

nhtruong:
## Goals
We chose maintaining OpenSearch API Spec via Smithy, even though the end-result is standard OpenAPI spec, mainly for its maintainability. So, the generated Smithy Spec should take advantage of Smithy unique features while assure that the models can still be properly translated to OpenAPI:
- Each OpenSearch action, each comprised of one or many API operations, are grouped into folders. For example, the smithy files for `indicies.create` action will be stored in `./model/indices/create` folder. (This grouping is impossible with OpenAPI 3.1, one of many reasons we chose Smithy)
- Each action has 2 smithy files:
  - `operations.smithy`: Describes the operations of said action
  - `structures.smithy`: Describes the structures of the Inputs and Outputs of the operations.
- Most parameters are not only shared within the operations of the same action, but they are also reused in other actions. The parameter definitions, therefore, are stored in common locations for resuse. For example, `common_booleans.smithy` and `common_string_lists.smithy`

reta:
@nhtruong the choice of Smithy is still not obvious, specifically taking into account plugins and extensions, I believe there are compelling arguments to use OpenAPI 3.0+ directly [1], without Smithy shim.

[1] https://github.com/opensearch-project/opensearch-clients/issues/55

nhtruong:
@reta I'm still drafting this comment and accidentally hit `Comment`. Still editing it now. It will be a bit clearer. As discussed in https://github.com/opensearch-project/opensearch-api-specification/issues/80, the end result will be OpenAPI 3.1. We're only using smithy, for now at least, as a device to backfill missing info (like Output schemas, and Body schemas), because it's easier to do so.

VachaShah:
Reopening this issue, since there is a part 2 for the PR.

PR #83, opened 2023-03-16
Generated api

nhtruong:
### Description
Generated smithy models from [legacy spec](https://github.com/opensearch-project/OpenSearch/tree/42058aa76a18f531ceaac7deb4e94944cf02f463/rest-api-spec/src/main/resources/rest-api-spec/api) using my [translation script](https://github.com/nhtruong/opensearch-smithy-spec-translator)

### Issues Resolved
closes https://github.com/opensearch-project/opensearch-api-specification/issues/82


nhtruong:
Created this draft to collect feedback as I'm almost done with the [translation script](https://github.com/nhtruong/opensearch-smithy-spec-translator).

nhtruong:
Files of interest:
- [Structures](https://github.com/opensearch-project/opensearch-api-specification/pull/83/files#diff-2fa53d6c5ecff002eecadb3c1075d3dba56d8a635d4c4fa0b430c2f44896c274) and [operations](https://github.com/opensearch-project/opensearch-api-specification/pull/83/files#diff-dadd5d90ddef675d0e60edcf193aea5a5bc1565e603cf849993af22a1e058649) of `index` action where multiple operations are grouped into one action
- [Common booleans](https://github.com/opensearch-project/opensearch-api-specification/pull/83/files#diff-b475b325271b2f2e0c45577cba47b4b5d9f8a253811d586173dd246f99a34866)
- [Common enums](https://github.com/opensearch-project/opensearch-api-specification/pull/83/files#diff-791cc46830d2c07f89eaba9b92b55a030288e4d56f0bc589f50d261b949166f0)
- [Common integers](https://github.com/opensearch-project/opensearch-api-specification/pull/83/files#diff-477adfd1358f2c470f807925ad0b99b4a5327ed1271d85d0cc11febb987e6a76)
- [Common lists](https://github.com/opensearch-project/opensearch-api-specification/pull/83/files#diff-71d67a3e6a35585697ad0b77d3782acb70487d27295422c1f2e85caf2bf4c1ff)
- [Common strings](https://github.com/opensearch-project/opensearch-api-specification/pull/83/files#diff-161fa3f651009f3b87e0fa56e9d38476db795a9c77bde5873acadef166feabdc)

model/_global/bulk/operations.smithy
```diff
@@ -0,0 +1,58 @@
...
+  output: Bulk_Output
+}
+
+@vendorExtensions(
+  "x-operation-group": "bulk",
+  "x-version-added": "1.0"
+)
+@http(method: "POST", uri: "/{index}/_bulk")
+@documentation("Allows to perform multiple index/update/delete operations in a single request.")
+operation Bulk_2 {
```
Xtansia, line 43: I think you can achieve uniqueness while avoiding the numbering by following the naming template of some of the handwritten operations: `{method}{operation}With{pathParams}`
e.g:
```
PostBulk
PutBulk
PostBulkWithIndex
PutBulkWithIndex
```

nhtruong: I saw that but there are actions with multiple endpoints using the same action like The [index](https://github.com/opensearch-project/opensearch-api-specification/pull/83/files#diff-dadd5d90ddef675d0e60edcf193aea5a5bc1565e603cf849993af22a1e058649) action where there are 2 POSTs.

nhtruong: There are other actions that use the same verb twice

Xtansia: I think it should just be something sort of like this pseudo code:
```
with = pathParams.len > 0 ? "With${pathParams.map(CamelCase).join("And")}" : ""
name = "${CamelCase(method)}${CamelCase(operationName)}${with}"
```

nhtruong:
[Added a comment in the issue](https://github.com/opensearch-project/opensearch-api-specification/issues/82#issuecomment-1486073933) to explain what I'm trying to achieve with this PR.

model/_global/bulk/operations.smithy
nhtruong: The `WITH` suffix was a [bit more complicated](https://github.com/nhtruong/opensearch-smithy-spec-translator/commit/38867328699d7b845cdb4a9cb171a4b8827c76de#diff-4d29948f6a7d0e22ddb3eb9e4547957e17d90a43d5352bacd6775120cdcc6ff1R54-R60) but it's been solved. Thanks for the nudge! :D (Also that method would have been 3-4 times as long in JS or Python. That's why I love Ruby so much!)

VachaShah (commented):
* Can we commit the translation script as well? It would help if we need to do this again for some APIs.
* Can the translation script convert a set of APIs to Smithy models?
* Currently, it looks like this script is replacing the already existing Smithy models. Is that going to be the case?

nhtruong:
@VachaShah I've given this topic some thought since we last discussed it.

The biggest hurdles I faced when writing the translator are: 
- Grouping identical parameters into the same Smithy shapes:
  - Params with slightly different descriptions but are actually the same because they were written by different people (i.e. same description but phrased differently)
  - Identical params but with different names
- Renaming vague parameter names to somethings more specific to avoid collisions. For example, `name` will be turned into `template_name`, `alias_name`, or `pipeline_name`... and then find out which `name` param in an operation is a `template_name` or a `pipeline_name` and so on.
- Params that are very similar to one another but with small differences. 
  - They usually have identical schema but have subtle nuances in certain operations. They will refer to the same Smithy shape that has the most common description, and have this description overridden in certain operations where they differ.
  - Sometimes they only differ in the default value, and deprecation flag. They will also reference the same Smithy shape but with unique default value / deprecation flag.
  - There are also enum params with identical names but slightly different enum options. I ended up creating a different shapes for different sets of options.

These hurdles had to be handled on case-to-case basis. That's why there are [a ton of if-else guards](https://github.com/nhtruong/opensearch-smithy-spec-translator/blob/main/lib/parameter.rb) in the translator, and even more manual updates to the original spec like this [commit](https://github.com/nhtruong/opensearch-smithy-spec-translator/commit/02315da8e62d7064fd9e40798221b1f64e7eedc9). There's no one-size-fit-all solution that we can program into the translator to handle this. That means, when a new API action is added to the legacy spec, and we're to incrementally convert it to Smithy using the translator, the translator will have to be able to:
- Map the some of the new action's params to the existing params. This is not programmatically viable for the reasons stated above.
- Assuming that we can somehow do the mapping above, the translator will still have to create new Smithy shapes for brand new params that don't exist in any current operations. That means the translator will have to be aware of the existing shape to avoid shape name collisions. And if there are collisions, how would it know what to rename the param to? 

So in the end, each new API action added will most likely become a new edge-case that you will have to modify the translator to handle. It would be a lot simpler and faster for us to convert it to Smithy manually. We only see new features that add new endpoints every 3 or 4 months. Also we really should encourage Core team to update this Smithy spec rather than the legacy spec, which is currently not being used anywhere else.

nhtruong:
Regarding existing models being replace: The newest version of the translator (which is not reflected in the current draft) is aware of the existing files and it will create file of similar name. I'll then use it to update the existing files manually.

This will be turned in 2 PRs:
- A PR with the definition of all parameters, existing files after being merged with the generated version. A couple of newly generated API Actions to demonstrate how the new Smithy models will look like. Total about 20 files to make it easier to review
- A PR with all other generated models. If the first PR's been approved and merged, there's not much to review in this one even though it will include about 300 files.

VachaShah:
> @VachaShah I've given this topic some thought since we last discussed it.
> 
> The biggest hurdles I faced when writing the translator are:
> 
> * Grouping identical parameters into the same Smithy shapes:
>   
>   * Params with slightly different descriptions but are actually the same because they were written by different people (i.e. same description but phrased differently)
>   * Identical params but with different names
> * Renaming vague parameter names to somethings more specific to avoid collisions. For example, `name` will be turned into `template_name`, `alias_name`, or `pipeline_name`... and then find out which `name` param in an operation is a `template_name` or a `pipeline_name` and so on.
> * Params that are very similar to one another but with small differences.
>   
>   * They usually have identical schema but have subtle nuances in certain operations. They will refer to the same Smithy shape that has the most common description, and have this description overridden in certain operations where they differ.
>   * Sometimes they only differ in the default value, and deprecation flag. They will also reference the same Smithy shape but with unique default value / deprecation flag.
>   * There are also enum params with identical names but slightly different enum options. I ended up creating a different shapes for different sets of options.
> 
> These hurdles had to be handled on case-to-case basis. That's why there are [a ton of if-else guards](https://github.com/nhtruong/opensearch-smithy-spec-translator/blob/main/lib/parameter.rb) in the translator, and even more manual updates to the original spec like this [commit](https://github.com/nhtruong/opensearch-smithy-spec-translator/commit/02315da8e62d7064fd9e40798221b1f64e7eedc9). There's no one-size-fit-all solution that we can program into the translator to handle this. That means, when a new API action is added to the legacy spec, and we're to incrementally convert it to Smithy using the translator, the translator will have to be able to:
> 
> * Map the some of the new action's params to the existing params. This is not programmatically viable for the reasons stated above.
> * Assuming that we can somehow do the mapping above, the translator will still have to create new Smithy shapes for brand new params that don't exist in any current operations. That means the translator will have to be aware of the existing shape to avoid shape name collisions. And if there are collisions, how would it know what to rename the param to?
> 
> So in the end, each new API action added will most likely become a new edge-case that you will have to modify the translator to handle. It would be a lot simpler and faster for us to convert it to Smithy manually. We only see new features that add new endpoints every 3 or 4 months. Also we really should encourage Core team to update this Smithy spec rather than the legacy spec, which is currently not being used anywhere else.

This makes sense that the translator will anyway have to adapted to the new APIs. We can make this part of the process when we deprecate the rest-api-spec on how to write Smithy models/OpenAPI specs for future.

VachaShah:
> Regarding existing models being replace: The newest version of the translator (which is not reflected in the current draft) is aware of the existing files and it will create files of similar names. I'll then use them to update the existing files manually.
> 
> This will be turned in 2 PRs:
> 
> * A PR with the definition of all parameters, existing files after being merged with the generated version. A couple of newly generated API Actions to demonstrate how the new Smithy models will look like. Total about 20 files to make it easier to review
> * A PR with all other generated models. If the first PR's been approved and merged, there's not much to review in this one even though it will include about 300 files.

Sounds good!

nhtruong:
There are a couple of roadblocks I'm facing right now with Smithy. When running  `./gradlew build`. I get several instances of these errors:
### URI Conflict:
```
Operation URI, `/{index}/_source/{id}`, conflicts with other operation URIs in the same service: 
[`OpenSearch#CatIndices_WithIndex` (/_cat/indices/{index}), `OpenSearch#CatIndices` (/_cat/indices), 
`OpenSearch#CatNodes` (/_cat/nodes), `OpenSearch#ClusterGetSettings` (/_cluster/settings), 
`OpenSearch#IndicesGetSettings_WithName` (/_settings/{name}), `OpenSearch#IndicesGetSettings` (/_settings), 
`OpenSearch#Search_Get` (/_search)]
```
Tagging every operation with `@suppress(["HttpUriConflict"])` trait doesn't resolve it

### Path Parameters as Lists:
```
Trait `httpLabel` cannot be applied to `OpenSearch#IndicesGetSettings_WithIndexName_Input$name`. 
This trait may only be applied to shapes that match the following selector: 
structure > member[trait|required] :test(> :test(string, number, boolean, timestamp))
```
Meaning smithy doesn't allow a list as a path parameter. However, it's very common in Opensearch to have a comma-deliminated  list of indices as path parameters. Sure we can convert them to strings and somehow tell the client generators that these strings are actually lists of strings. This will be even more complicated when we want to tell the generators that it's a list of enums, which there are a few.

VachaShah:
For URI Conflict:

Is it considering `(/_cat/indices/{index})` and `(/_cat/indices)` as same URI?

Path Parameters as List:

Does [Greedy Labels](https://smithy.io/2.0/spec/http-bindings.html?highlight=greedy#greedy-labels) help?

nhtruong:
- The URI conflicts comes from `{index}/<anything>` conflicting with `_cat/indices/...` and such because smithy believes `_cat` can also be `{index}` param too.
- Greedy label is not the same. In this case we're looking at comma-diliminated path params like  books and movies in `GET /_cat/indices/books,movies`
- There's also another smithy error that I forgot to capture: There are endpoints where some query params share the same names as the body params. Smithy doesn't allow this because all parameters (Query/Path/Body) are flatten into the same Smithy input structure that requires unique names for all parameters.
- We are also not allowed to have the same `@output` structure for different operations. There's a couple of workarounds but Smithy doesn't condone them.

All of these stem from unconventional/outdated API design of Elastic/OpenSearch:
- Having multiple operations serving the exact same functionality
- Path param at the root of the URI: `PUT {index}/_doc` instead of `PUT indices/{index}/docs`
- REST actions in URI: `POST /{index}/_create/{id}`
- Unconventional use of HTTP Verbs. Most notably `PUT` being used to create new resources, and or not being idempotent.
- Duplicate parameters in Query and Body.

I believe these are the reasons Elastic decided to go with a custom JSON schema instead of OpenAPI (eps the first bullet-point), even though OpenAPI can be extended to handle this API's oddities. Smithy is, unfortunately, a lot less forgiving when it comes to bad API design practice, which makes it very difficult to retrofit this API into Smithy models.

nhtruong:
Workarounds for these issues after discussing with others:
- Apply `@pattern` to path params to resolve URI Conflict errors
- Convert list path params to String but mark them as list  using vendorExtensions
- Rename query string params to avoid collisions with body params but keep the original names when using query trait.
- Do not mark output structures with `@output` trait

nhtruong:
Note that this is the first of 2 PRs that will translate the legacy rest-api to smithy:
- Part 1 has the definitions of all parameters in the API, and definitions of operations that are already in in the repo but updated to match their translated versions. This PR only has about 60 files to review.
- Part 2 has the rest of the operations that are not yet defined in the repo. This PR is going to have about 300 files added.
This is to help reviewers pinpoint on the files that matter: The common param files, and a few operation definitions. Once part 1 has been approved and merged, part 2 doesn’t need much reviewing since the operations it provides will have identical patterns seen in Part 1.

VachaShah (commented):
I took a first pass at it and this looks great @nhtruong! I will take a second look for the parameters.

Once this is finalized, can we also include the changes to [DEVELOPER_GUIDE](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md) since the file names, paths etc. have changed?

VachaShah (approved):
@nhtruong The changes overall look good to me, can we also add the changes to the developer guide?

Xtansia:
Can we mark those of the `time` type with an extension property signifying it? I can currently assume it based on the pattern, but may be useful to have something explicit?

nhtruong:
Going to make some changes based on feedback here and on slack:
- Update Dev guide
- Denote `time` with `x-data-type: time` (will use this extension for list and enum-list path params too)
- Redefine BodyParams to include `content: Document` with `@httpPayload` trait

Xtansia:
This probably more affects when we bring in the rest of the APIs, but we'll need to think about how we represent the "bulk" APIs request bodies in Smithy as they're not strictly JSON, they're new-line delimited JSON objects:
e.g: https://opensearch.org/docs/2.6/api-reference/document-apis/bulk/
```
POST _bulk
{ "delete": { "_index": "movies", "_id": "tt2229499" } }
{ "index": { "_index": "movies", "_id": "tt1979320" } }
{ "title": "Rush", "year": 2013 }
{ "create": { "_index": "movies", "_id": "tt1392214" } }
{ "title": "Prisoners", "year": 2013 }
{ "update": { "_index": "movies", "_id": "tt0816711" } }
{ "doc" : { "title": "World War Z" } }
```

nhtruong:
@Xtansia  Wouldn't the be considered an array of objects and is still a valid JSON?

Xtansia:
> @Xtansia Wouldn't the be considered an array of objects and is still a valid JSON?

No, it's not an array, it does not have opening and closing brackets or commas. It's specifically one serialized object per line

Xtansia:
In code it'd be an array, but on the wire it is not valid JSON as a whole

nhtruong:
Gotcha. Nice catch!

model/_global/info/operations.smithy
```diff
@@ -0,0 +1,26 @@
...
+//
+//  The OpenSearch Contributors require contributions made to
+//  this file be licensed under the Apache-2.0 license or a
+//  compatible open source license.
+
+$version: "2"
+namespace OpenSearch
+use opensearch.openapi#vendorExtensions
+
+// TODO: Fill in API Reference URL
```
VachaShah, line 11: Is this still a TODO?

VachaShah (approved):
Just a small comment, otherwise LGTM!

Issue #100, opened 2023-05-17
[FEATURE] Add Request and Response Bodies for 4 main Point-in-time endpoints


nhtruong:
There are [6 PIT actions](https://opensearch.org/docs/latest/point-in-time-api/) that was added to OS 2.4. The following 4 currently do not have Request and Response bodies:
- **create_pit**: `POST {index}/_search/point_in_time`
- **get_all_pits**: `GET _search/point_in_time/_all`
- **delete_all_pits**: `DELETE _search/point_in_time/_all`
- **delete_pit**: `DELETE _search/point_in_time`

Follow the [Developer's Guide](./DEVELOPER_GUIDE.md) doc to fill in the missing info.
You might have to manually make calls to an OS cluster to figure out the shape of the response bodies.

NOTE: Don't work on this issue until https://github.com/opensearch-project/opensearch-api-specification/pull/91 has been merged.

Tokesh:
Hello everyone!
We discussed at the meeting with @nhtruong , that @aabeshov and I will do this issue!

Tokesh:
Hello everyone! 
We have fortunately made this issue and are waiting for your comments about our work!

P.S. We divide it to 2 PR, because we have some problems with counting of job in our CCI program
@aabeshov provided Request and Response Bodies for [create_pit and delete_all_pits](https://github.com/opensearch-project/opensearch-api-specification/pull/104)
I provided Request and Response Bodies for [get_all_pits and delete_pit](https://github.com/opensearch-project/opensearch-api-specification/pull/105)
If its not comfortable for checking or merging, we will definitely discuss this in our conference of CCI and will do it in the future in one pull request

VachaShah:
> Hello everyone! We have fortunately made this issue and are waiting for your comments about our work!
> 
> P.S. We divide it to 2 PR, because we have some problems with counting of job in our CCI program @aabeshov provided Request and Response Bodies for [create_pit and delete_all_pits](https://github.com/opensearch-project/opensearch-api-specification/pull/104) I provided Request and Response Bodies for [get_all_pits and delete_pit](https://github.com/opensearch-project/opensearch-api-specification/pull/105) If its not comfortable for checking or merging, we will definitely discuss this in our conference of CCI and will do it in the future in one pull request. We had no experience in production or open source development before and we are just learning!

Thank you @Tokesh and @aabeshov for contributing! The division into PRs is totally fine. Will review the PRs.

VachaShah:
[...]

PR #105, opened 2023-05-28
Added Request and Response Bodies for 2 main Point-in-time endpoints

Tokesh:
### Issues Resolved
[#100]


model/_global/delete_pit/structures.smithy
```diff
@@ -7,18 +7,22 @@
...
-// TODO: Fill in Body Parameters
-structure DeletePit_BodyParams {}
+list PitInfoList {
+    member: PitInfo
+}
 
 @input
-structure DeletePit_Input with [DeletePit_QueryParams] {
-    @httpPayload
-    content: DeletePit_BodyParams,
```
nhtruong, line 20: We must use @httpPayload trait to mark the body structure.

nhtruong:
Don't forget to resolve the conflicts

nhtruong (dismissed):

model/_global/delete_pit/structures.smithy
```diff
@@ -7,20 +7,32 @@
 $version: "2"
 namespace OpenSearch
 
-@mixin
-structure DeletePit_QueryParams {
+list PitInfoList {
+    member: PitInfo
```
Xtansia, line 11: We need to be careful of giving things too generic names as everything is within one namespace here. I think it'd make sense for this list and the below structure to be `DeletedPit` rather than `PitInfo`

model/_global/get_all_pits/structures.smithy
```diff
@@ -7,14 +7,21 @@
 $version: "2"
 namespace OpenSearch
 
-@mixin
-structure GetAllPits_QueryParams {
+structure PitDetail{
+    pitId: String,
```
Xtansia, line 11: This should be `pit_id`

Xtansia (changes requested):

model/_global/delete_pit/structures.smithy
Tokesh: > We need to be careful of giving things too generic names as everything is within one namespace here. I think it'd make sense for this list and the below structure to be `DeletedPit` rather than `PitInfo`

Thank you, i got it. You can recheck now

Xtansia (approved):

VachaShah (approved):

Issue #106, opened 2023-05-29
[FEATURE] Add Specs for Security plugin API


Tokesh:
This issue covers adding models for [Security plugins](https://opensearch.org/docs/2.7/security/access-control/api/) listed below:

- [x] [Account](https://opensearch.org/docs/2.7/security/access-control/api/#account)
- [x] [Action groups](https://opensearch.org/docs/2.7/security/access-control/api/#action-groups)
- [x] [Users](https://opensearch.org/docs/2.7/security/access-control/api/#users)
- [x] [Roles](https://opensearch.org/docs/2.7/security/access-control/api/#roles)
- [x] [Role mappings](https://opensearch.org/docs/2.7/security/access-control/api/#role-mappings)
- [x] [Tenants](https://opensearch.org/docs/2.7/security/access-control/api/#tenants)
- [x] [Configuration](https://opensearch.org/docs/2.7/security/access-control/api/#configuration)
- [x] [Distinguished names](https://opensearch.org/docs/2.7/security/access-control/api/#distinguished-names)
- [x] [Certificates](https://opensearch.org/docs/2.7/security/access-control/api/#certificates)
- [x] [Cache](https://opensearch.org/docs/2.7/security/access-control/api/#cache)
- [x] [Health](https://opensearch.org/docs/2.7/security/access-control/api/#health)
- [x] [Audit logs](https://opensearch.org/docs/2.7/security/access-control/api/#audit-logs)

Follow the [Developer's Guide](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md) doc to fill in these actions.
You might have to manually make calls to an OS cluster to figure out the shape of the response bodies.
(We discussed about this API with @nhtruong on CCI meetup)

Tokesh:
Hello everyone!
This issue is similar to the previous ones. Because of this, I am ready to continue my work! :) 
Today I will try to write a PR

Tokesh:
Models have been added for all the above APIs.

PR #107, opened 2023-05-29
adding specs for Account API

Tokesh:
Specs for [Account API](https://opensearch.org/docs/2.7/security/access-control/api/#account)
### Issues Resolved
[#106]


model/common_structures.smithy
```diff
@@ -59,3 +59,8 @@ structure Hits{
     _source: Document,
     fields: Document
 }
+
+structure Response{
+    status: String
+    message: String
+}
```
nhtruong, line 66: A structure named response is very vague. Any Response body can be Response

Tokesh: > A structure named response is very vague. Any Response body can be Response

Can we name it "UpdateResponse"?

Tokesh: because we use it in many API's in update/delete parts

nhtruong (approved):
🚀

VachaShah (approved):

PR #127, opened 2023-06-30
Correct security plugin operations that take lists or return maps.

Xtansia:
### Description
The aws.protocols#restJson1 protocol explicitly disallows lists and maps for @httpPayload.

Requires workaround due to AbstractRestProtocol being package-private, and a bug in HttpBindingIndex waiting on PR https://github.com/awslabs/smithy/pull/1840

### DIFF
--- a/buildSrc/src/main/kotlin/versions.kt
+++ b/buildSrc/src/main/kotlin/versions.kt
@@ -1,1 +1,1 @@
-const val smithyVersion = "1.31.0"
+const val smithyVersion = "1.33.0"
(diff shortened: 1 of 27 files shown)


VachaShah (commented):
@Xtansia Do you think this is worth describing somewhere in the guides?

Xtansia:
> @Xtansia Do you think this is worth describing somewhere in the guides?

@VachaShah If you're referring to the protocol/hacks, I'm not sure where it'd fit into the guides as it should mostly be invisible to someone working on the models. Ideally 90% of this can be removed again soon™️ if the fix is merged upstream into Smithy.

VachaShah (approved):

Issue #129, opened 2023-07-10
Add Api definitions for create, get, and delete data_stream


nhtruong:
This is part of the `index_management` plugin that have now been added to OS as core features in the `indices` namespace, so the operation groups of these actions would be
`indices.create_data_stream`
`indices.get_data_stream`
`indices.delete_data_stream`

https://opensearch.org/docs/latest/im-plugin/data-streams/

Tokesh:
I'd like to work on this issue :)
I'm already read this article about data streams. I think i have only one question about create_data_stream API.
In article they gave us such examples:
PUT _data_stream/logs-redis
PUT _data_stream/logs-nginx
Are "logs-redis" and "logs-nginx" just indexes, not index_templates?

nhtruong:
`A data stream is internally composed of multiple backing indexes. Search requests are routed to all the backing indexes, while indexing requests are routed to the latest write index.`

So it makes sense that you need an index_template (it's like grepping for indices with similar names) to create a data-stream, since a data-stream is a collection of indices.

nhtruong:
@Tokesh Looks like even though these endpoints were part of the legacy `index_management` plug-in. They have been added to OS as core features of the `indices` namespace. I've update the Issue's description to reflect that.

Tokesh:
> So it makes sense that you need an index_template (it's like grepping for indices with similar names) to create a data-stream, since a data-stream is a collection of indices.

> @Tokesh Looks like even though these endpoints were part of the legacy `index_management` plug-in. They have been added to OS as core features of the `indices` namespace. I've update the Issue's description to reflect that.

Thank you so much for answers. I got it!

Xtansia:
Closing as #131 was merged

PR #131, opened 2023-07-14
Added Request and Response Bodies for Data Stream APIs

Tokesh:
### Description
I generally did everything in the latest version of the OS. I tried my best to check everything. 
But not sure about `IndicesCreateDataStream `request body. Do we need to fill out this ```structure IndicesCreateDataStream_BodyParams```. When I tested this, I didn't add anything and it worked.
I tried to understand all the mistakes of my past PRs. I hope this is the correct request/response bodies 😳
P.S. I moved the AttributesMap to commonStructure which located in root folder, because it's not only used in the Security folder.

### Issues Resolved
[#129]

### DIFF
--- a/model/common_structures.smithy
+++ b/model/common_structures.smithy
@@ -59,3 +59,8 @@ structure Hits{
     _source: Document,
     fields: Document
 }
+
+map AttributesMap {
+    key: String,
+    value: String
+}
--- a/model/common_structures.smithy
+++ b/model/common_structures.smithy
@@ -59,8 +59,3 @@ structure Hits{
     _source: Document,
     fields: Document
 }
-
-map AttributesMap {
-    key: String,
-    value: String
-}
--- a/model/indices/data_stream.smithy
+++ b/model/indices/data_stream.smithy
@@ -1,0 +1,30 @@
+// SPDX-License-Identifier: Apache-2.0
+//
+//  The OpenSearch Contributors require contributions made to
+//  this file be licensed under the Apache-2.0 license or a
+//  compatible open source license.
+
+$version: "2"
+namespace OpenSearch
+
+structure DataStream{
+    name:String,
+    timestamp_field:AttributesMap,
+    indices: IndicesList
+    generation: Integer,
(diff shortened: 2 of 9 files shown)


nhtruong:
If Possible can you also add description for the input bodies as well? Thank you.

model/indices/data_stream.smithy
```diff
@@ -0,0 +1,30 @@
...
+    name:String,
+    timestamp_field:AttributesMap,
+    indices: IndicesList
+    generation: Integer,
+    status: String,
+    template: String
+}
+
+list IndicesList{
+    member: AttributesMap
```
nhtruong, line 20: Even IF `timestamp_field` and `indices` have the exact same schema, we should still define two different models instead of using a generic one called `AttributesMap`. More over, from what I read in https://opensearch.org/docs/latest/im-plugin/data-streams/, `indices` have a more defined structure than Attributes map.

This should be
```
structure DataStream{
    name: String,  # Remember to put space after the colon as well.
    timestamp_field: DataStreamTimestampField,
    indices: DataStreamIndices
    generation: Integer,
    status: String,
    template: String
}

list DataStreamIndices {
    member: DataStreamIndex
}

structure DataStreamIndex {
  index_name: String,
  index_uuid: String
}

map DataStreamTimestampField {
  key: String
  value: String
}
```

Tokesh: Oh, i got it. Thank you! You can review it again

model/common_structures.smithy
```diff
@@ -59,3 +59,8 @@ structure Hits{
     _source: Document,
     fields: Document
 }
+
+map AttributesMap {
+    key: String,
+    value: String
+}
```
nhtruong, line 66: This should be removed right?

Tokesh: > This should be removed right?

Yes, you're right ! Thank you

nhtruong (approved):

VachaShah (approved):

PR #153, opened 2023-10-01
Fix missing pluralization on Role

jgillich:
### Description
Both `cluster_permission` and `index_permission` should be plural as per the [docs](https://opensearch.org/docs/1.2/security-plugin/access-control/api#create-role).

### DIFF
--- a/model/security/role.smithy
+++ b/model/security/role.smithy
@@ -16,8 +16,8 @@ structure Role {
     reserved: Boolean
     hidden: Boolean
     description: String
-    cluster_permission: ClusterPermission
-    index_permission: IndexPermission
+    cluster_permissions: ClusterPermission
+    index_permissions: IndexPermission
     tenant_permissions: TenantPermission
     static: Boolean
 }
--- a/model/security/role.smithy
+++ b/model/security/role.smithy
@@ -17,7 +17,7 @@ structure Role {
     hidden: Boolean
     description: String
     cluster_permissions: ClusterPermission
-    index_permissions: IndexPermission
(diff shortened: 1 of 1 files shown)


Xtansia (dismissed):

model/security/role.smithy
```diff
@@ -16,8 +16,8 @@ structure Role {
     reserved: Boolean
     hidden: Boolean
     description: String
-    cluster_permission: ClusterPermission
-    index_permission: IndexPermission
+    cluster_permissions: ClusterPermission
+    index_permissions: IndexPermissionList
```
Xtansia, line 20: It looks like `IndexPermission` is also missing a `dls: String` field

model/security/role.smithy
```diff
@@ -16,8 +16,8 @@ structure Role {
...
     tenant_permissions: TenantPermission
```
Xtansia, line 21: `tenant_permissions` should also actually be a list of objects, rather than a list of strings. With each `TenantPermission` being of shape (in pseudo-code):
```
{
    tenant_patterns: List<String>,
    allowed_actions: List<String>
}
```

Xtansia (changes requested):
Thanks for catching this @jgillich, with your latest fix for being a list, I looked a bit closer, and noticed a couple other discrepancies that'd be awesome if you could fix while you're in this area?

jgillich:
Sure thing @Xtansia, I hope I did everything correctly :smile:

Xtansia (approved):
Looks good, thanks!

VachaShah (approved):

Issue #168, opened 2023-12-20
[FEATURE] Add missing API specs


dblock:
### Is your feature request related to a problem?

The spec repo should be 100% of specs that ship with OpenSearch and kept up-to-date.

### What solution would you like?

1. Specs up-to-date to the latest distribution of OpenSearch.
2. A way to know which specs aren't up-to-date (a tool)?

### Do you have any additional context?

This is the current status of API specs.

- [ ] alerting
- [ ] anomaly-detection
- [ ] asynchronous-search
- [ ] cross-cluster-replication
- [ ] geospatial
- [ ] index-management
- [ ] job-scheduler
- [ ] k-nn
- [ ] ml-commons
- [ ] neural-search
- [ ] notifications
- [ ] observability
- [ ] opensearch-reports
- [ ] performance-analyzer
- [ ] security
- [ ] security-analytics
- [ ] sql

dblock:
> @dblock are they all plugins/extensions or do some of them going to be a core feature of OS?

Lots in core. I updated the list above.

dblock:
> @dblock are we doing both `_plugins` and `_opendistro`? We can totally DRY this. I'll need to update the validator to account for this. I'm also going to do this for `knn` namespace as an example.

I think `_opendistro` is there for backwards compat, we should not include it.

dblock:
I should probably take that back, if we want to claim coverage we need some way to know about these and skip them. Maybe the right way would be to mark them deprecated? What's the correct way to do that @nhtruong?

nhtruong:
I'm thinking of having a file named `_opendistro.yaml` that looks like this:
```yaml
_reports/on_demand/{reportDefinitionId}: POST
_rollup/jobs: GET,PUT
```
It will be loaded up with all `_plugins/...` endpoints that have the `_opendistro/...` equivalents. The merger tool will take this file into account and generate the opendistro endpoints that are identical to the plugins counterparts but with `deprecated` and `x-ignored` set to `true`. This has a few benefits:
- Save the contributors from having to fill-out these endpoints either manually or through any DRY mechanism we come up with (Since the opendistro version has additional flags, DRY'ing wont be simple)
- Keep the `spec` folder tidy and free of hundreds of deprecated endpoints
- Once these endpoints are removed from OS, we simply just remove said file and the logic in the merger.

dblock:
@nhtruong This is a good plan.

nhtruong:
[...]

PR #190, opened 2024-02-15
Added spec for k-NN stats api

saimedhi:
### Description

- Added spec for k-NN stats api 
- Tested the api with different path and query params [here](https://github.com/saimedhi/opensearch_api_testing/blob/main/knn.stats.md).
- k-NN stats doc https://opensearch.org/docs/latest/search-plugins/knn/api/#stats
- plugin code ref:  https://github.com/opensearch-project/k-NN/blob/main/src/main/java/org/opensearch/knn/plugin/rest/RestKNNStatsHandler.java

### Issues Resolved
Related to https://github.com/opensearch-project/opensearch-api-specification/issues/168,
https://github.com/opensearch-project/opensearch-py/issues/300


model/common_strings.smithy
```diff
@@ -247,6 +247,11 @@ string PathTaskId
 @documentation("The name of the template.")
 string PathTemplateName
 
+@xDataType("array")
```
saimedhi, line 250: @nhtruong,  
Do you think this is required here? `@xDataType("array") `

nhtruong: Yes. All comma separated values in `path parameters` are actually array

nhtruong: Btw, since it's an array, the name should be `PathStats`

nhtruong: Looks like the values of PathStats are limited to certain enum values. So this will need `xEnumOptions` as well.

model/knn/stats/structures.smithy
```diff
@@ -0,0 +1,46 @@
...
+//  this file be licensed under the Apache-2.0 license or a
+//  compatible open source license.
+
+$version: "2"
+namespace OpenSearch
+
+@mixin
+structure KNNStats_QueryParams {
+    @httpQuery("timeout")
+    timeout: Timeout,
```
nhtruong, line 13: I see `pretty` querystring param in the examples as well.

saimedhi: Theo, pretty is a [global parameter](https://github.com/opensearch-project/opensearch-py/blob/main/utils/generate_api.py#L61).

Those parameters can be used with all APIs. So, we won't write them in spec. Please confirm?

saimedhi:
@nhtruong made the requested changes. And also PR #191 , #192 ready for review.

saimedhi:
@Xtansia, @nhtruong  Please take a look and let me know if any changes needed.

Xtansia (dismissed):

dblock (approved):

Xtansia (approved):

PR #218, opened 2024-04-08
Removed line-wrap for all yaml files

nhtruong:
When splitting and converting the spec from single JSON to multiple YAML file, I wasn't aware that some line-wrap was added by the YAML encoder by default(e.g. a 120-char-per-line limit was accidentally added). This PR is removing that limit


Xtansia:
@nhtruong Is there a reason to remove the line-wrapping? Does it cause issues with tooling?

nhtruong:
No it doesn't. It just creates extra empty lines here and there. And I'd rather let the user use their IDE's line-wrap features rather than enforcing a fixed one. We all different screen-sizes.

Issue #205, opened 2024-03-21
[FEATURE] Lint API specs


dblock:
### Is your feature request related to a problem?

In https://github.com/opensearch-project/opensearch-api-specification/pull/203 we renamed a path parameter from `NodeId` to `node_id` because it caused a downstream generator in the opensearch-py client to produce a camelCase parameter which was caught in a linter there and errored.

### What solution would you like?

A linter for consistency in the API spec for parameters, variables, etc.

### Do you have any additional context?

https://github.com/opensearch-project/opensearch-py/pull/700#issuecomment-2010365233

nhtruong:
I will add this to the validator im working on as one of the follow up PRs.

dblock:
Coming from https://github.com/opensearch-project/opensearch-api-specification/pull/218, linter should check description format (capitalized, full sentences with periods), and no line wrapping.

PR #219, opened 2024-04-09
Added `tools/linter` along with jest tests.

nhtruong:
### Description
Added a Linter to validate all `yaml` files in the `/spec` folder

### Issues Resolved
https://github.com/opensearch-project/opensearch-api-specification/issues/205
(Not closing as we still need to turn this into a workflow with a follow-up PR)


dblock (dismissed):
Looks good. 

1. Add the linter build/test to CI part of this commit (even if you want to add linting itself as part of another), so future changes to the linter are tested.
2. How much of the linter is OpenSearch-specific? Longer term, I wonder if it should be built as its own project in its own repo.

nhtruong:
@dblock 
**2. How much of the linter is OpenSearch-specific? Longer term, I wonder if it should be built as its own project in its own repo.**

Half of the validator rules are OpenSearch specific:
 - Rules regarding `x-operation-group`
 - Rules regarding parameter naming conventions

The other rules enforce the multi-file structure we created for OpenSearch (diving paths into namespaces in schemas into categories). The merger tool assumes that there are no infractions in these rules to ensure that the result single-file is a valid OpenAPI file. Projects that want to reuse this structure will benefit from these (Though the namespace is actually a part of `x-operation-group` so we will have to untangle that first.)

dblock (approved):

PR #246, opened 2024-04-16
Fixed few param descriptions

### LINKED ISSUE #243: [BUG] indices.put_settings::query.ignore_unavailable param description is wrong.
### What is the bug?
indices.put_settings::query.ignore_unavailable param description is wrong.

From spec: https://github.com/opensearch-project/opensearch-api-specification/releases/download/main/opensearch-openapi.yaml

- `indices.put_settings::query.ignore_unavailable` param description is wrong. It is given description of other param `indices.put_settings::query.flat_settings`

  ```
  indices.put_settings::query.flat_settings:
      in: query
      name: flat_settings
      description: If `true`, returns settings in flat format.
      schema:
        type: boolean
      style: form


```
 indices.put_settings::query.ignore_unavailable:
      in: query
      name: ignore_unavailable
      description: If `true`, returns settings in flat format.
      schema:
        type: boolean
      style: form
```

saimedhi:
### Description
Fixed few param descriptions

### Issues Resolved
#243 


saimedhi:
@nhtruong Please take a look

spec/namespaces/indices.yaml
```diff
@@ -2972,7 +2972,7 @@ components:
       description: |-
         Name of the data stream, which must meet the following criteria:
         Lowercase only;
-        Cannot include `\`, `/`, `*`, `?`, `"`, `<`, `>`, `|`, `,`, `#`, `:`, or a space character;
```
nhtruong, line 2975: why is this needed?

saimedhi: getting lint error with python generator: invalid escape sequence ' \ `'

saimedhi: And also an error while generating python client docs.
expected doc:
`Cannot include ‘\’, ‘/’`
obtained:
`Cannot include ‘’, ‘/’`

nhtruong: Try replacing `\` with `\\` in the descriptions

spec/namespaces/indices.yaml
```diff
@@ -2972,7 +2972,7 @@ components:
...
+        Cannot include `/`, `*`, `?`, `"`, `<`, `>`, `|`, `,`, `#`, `:`, backslash, or a space character;
```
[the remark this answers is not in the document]
nhtruong: Would `\\` work?

spec/namespaces/indices.yaml
saimedhi: @nhtruong, just to give you some background: Here are the two things
- [Generating APIs](https://github.com/opensearch-project/opensearch-py/blob/main/DEVELOPER_GUIDE.md#client-code-generator) from the spec
       Here, param descriptions will be generated as it is. 

- [Building documentation](https://github.com/opensearch-project/opensearch-py/blob/main/DEVELOPER_GUIDE.md#documentation) using Sphinx.

- Currently, 
 Generated API:`Cannot include ‘\’, ‘/’` 
 Sphinx documentation:`Cannot include ‘’, ‘/’` 

- When replaced \ with \\,
 Generated API:`Cannot include ‘\\’, ‘/’`          **not as expected**
 Sphinx documentation:`Cannot include ‘\’, ‘/’` 


Do you think I should create a workaround in Python? I'm ready to do it if necessary, but I'm also worried that other clients might face the same problem.

dblock: That seems right @saimedhi. Having a `\` is valid in a spec, but may need to be converted in other formats (python). This is similar to HTML which encodes characters like `<` to `&lt;`.

dblock (approved):

PR #251, opened 2024-04-18
Replace pull_request_target workflow with two workflows that upload/download an artifact.

dblock:
### Description

When we use `pull_request_target` we check out code that has potentially been modified by the requester, which isn't safe. This PR replaces the workflow with something inspired by https://securitylab.github.com/research/github-actions-preventing-pwn-requests/ - a gather step and a comment step that use an artifact to pass information between them.

Here's a a gather workflow that produces the artifact: https://github.com/dblock/opensearch-api-specification/actions/runs/8740901986
Here's a comment workflow that uses it: https://github.com/dblock/opensearch-api-specification/actions/runs/8740884453

### DIFF
--- a/.github/workflows/coverage-comment.yml
+++ b/.github/workflows/coverage-comment.yml
@@ -14,28 +14,11 @@ jobs:
       github.event.workflow_run.conclusion == 'success'
     steps:
       - name: Download Coverage Report
-        uses: actions/github-script@v3.1.0
+        uses: actions/download-artifact@v4
         with:
-          script: |
-            var artifacts = await github.actions.listWorkflowRunArtifacts({
-               owner: context.repo.owner,
-               repo: context.repo.repo,
-               run_id: ${{github.event.workflow_run.id }},
-            });
-            var matchArtifact = artifacts.data.artifacts.filter((artifact) => {
-              return artifact.name == "coverage"
-            })[0];
-            var download = await github.actions.downloadArtifact({
-               owner: context.repo.owner,
-               repo: context.repo.repo,
-               artifact_id: matchArtifact.id,
-               archive_format: 'zip',
-            });
-            var fs = require('fs');
-            fs.writeFileSync('${{github.workspace}}/coverage.zip', Buffer.from(download.data));
-      - run: |
-            unzip coverage.zip
-            cat coverage.json
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          name: coverage
+          run-id: ${{ github.event.workflow_run.id }}
 
       - name: 'Comment on PR'
         uses: actions/github-script@v3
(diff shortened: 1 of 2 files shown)


.github/workflows/coverage-gather.yml
```diff
@@ -50,14 +48,15 @@ jobs:
...
+          {
+            "pull_request":${{ github.event.number }},
+            "current":$current,
+            "total":$total,
+            "percent":$percent
+          }
           EOL
-      - uses: peter-evans/create-or-update-comment@v4
-        if: github.event_name == 'pull_request_target'
+      - uses: actions/upload-artifact@v2
```
Xtansia, line 59: `actions/upload-artifact@v2` is set to be deprecated, `v4` is the latest: https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/

.github/workflows/coverage-comment.yml
```diff
@@ -0,0 +1,53 @@
...
+               owner: context.repo.owner,
+               repo: context.repo.repo,
+               artifact_id: matchArtifact.id,
+               archive_format: 'zip',
+            });
+            var fs = require('fs');
+            fs.writeFileSync('${{github.workspace}}/coverage.zip', Buffer.from(download.data));
+      - run: |
+            unzip coverage.zip
+            cat coverage.json
```
Xtansia, line 38: `actions/download-artifact@v4` now supports downloading artifacts from other workflow runs: https://github.com/actions/download-artifact?tab=readme-ov-file#download-artifacts-from-other-workflow-runs-or-repositories

Xtansia (changes requested):

.github/workflows/coverage-comment.yml
dblock: Updated. This is much nicer.

Successful comment run: https://github.com/dblock/opensearch-api-specification/actions/runs/8752961841
Comment on: https://github.com/dblock/opensearch-api-specification/pull/5

dblock:
Also https://code.dblock.org/2024/04/19/commenting-on-pull-requests-in-github-actions.html.

.github/workflows/coverage-comment.yml
```diff
@@ -0,0 +1,36 @@
...
+          name: coverage
+          run-id: ${{ github.event.workflow_run.id }}
+
+      - name: 'Comment on PR'
+        uses: actions/github-script@v3
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            const fs = require('fs');
+            var data = JSON.parse(fs.readFileSync('./coverage.json'));
```
nhtruong, line 29: nit: avoid declaring a variable with `var`, as you MIGHT overwrite `var data` declared else where use `const` in this case instead.

Xtansia (approved):

PR #253, opened 2024-04-18
Fixed search_pipeline spec

### LINKED ISSUE #248: [BUG] Differences in search_pipeline API between OpenSearch Server and Specification.
### What is the bug?
The difference lies between the search_pipeline spec and server code. I'm updating only the ones I've observed. I'll take another careful look and update this issue soon.

| | Spec | Server |
|-|-|-|
| API Name | search_pipeline.create | search_pipeline.put |
| Path Param Name |  pipeline | id |
|search_pipeline.get Path Param pipeline | required = True | required = False |
.....

The delete method still needs to be added to the spec.

saimedhi:
### Description

- Corrected spec for search_pipeline.get, search_pipeline.create
- Added  search_pipeline.delete

### Issues Resolved
Closes #248 

### DIFF
--- a/spec/namespaces/search_pipeline.yaml
+++ b/spec/namespaces/search_pipeline.yaml
@@ -4,14 +4,15 @@ info:
   description: OpenSearch Search_pipeline API
   version: 1.0.0
 paths:
-  /_search/pipeline/{pipeline}:
+  /_search/pipeline/{id}:
     get:
       operationId: search_pipeline.get.0
       x-operation-group: search_pipeline.get
       x-version-added: '2.9'
       description: Retrieves information about a specified search pipeline.
       parameters:
-        - $ref: '#/components/parameters/search_pipeline.get::path.pipeline'
+        - $ref: '#/components/parameters/search_pipeline.get::query.cluster_manager_timeout'
+        - $ref: '#/components/parameters/search_pipeline.get::path.id'
       responses:
         '200':
           $ref: '#/components/responses/search_pipeline.get@200'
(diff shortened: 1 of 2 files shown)


spec/namespaces/search_pipeline.yaml
```diff
@@ -4,14 +4,15 @@ info:
   description: OpenSearch Search_pipeline API
   version: 1.0.0
 paths:
-  /_search/pipeline/{pipeline}:
+  /_search/pipeline/{id}:
```
saimedhi, line 7: @nhtruong  Do you think we should replace path param name from `pipeline` to `id`?

spec/namespaces/search_pipeline.yaml
```diff
@@ -23,12 +24,26 @@ paths:
       externalDocs:
```
saimedhi, line 24: @nhtruong @Xtansia `PUT /_search/pipeline/my_pipeline` endpoint [creates](https://opensearch.org/docs/latest/search-plugins/search-pipelines/creating-search-pipeline/) and [updates](https://opensearch.org/docs/latest/search-plugins/search-pipelines/creating-search-pipeline/#updating-a-search-pipeline) search_pipeline. 
So, should we name it `search_pipeline.put` instead of `search_pipeline.create`?

Xtansia: Yes I think so, it makes more sense semantically. Especially when things such as the _doc APIs set the precedent that "create" means fail if the item already exists.

spec/namespaces/search_pipeline.yaml
nhtruong: Both make sense to me. If the endpoints using this paths haven't made it to any clients then we're free to rename without introducing breaking changes. I know that Ruby and JS clients haven't implemented these yet. @Xtansia what about Java, Net, and Rust?

spec/namespaces/search_pipeline.yaml
nhtruong: agree with @Xtansia

spec/namespaces/search_pipeline.yaml
Xtansia: Hasn't been added into any of those clients

nhtruong (approved):

spec/namespaces/search_pipeline.yaml
nhtruong: @saimedhi Then this name change is fine!

PR #264, opened 2024-04-22
Removed default values from parameter descriptions

### LINKED ISSUE #263: [BUG]Remove default values present in parameter descriptions.
### What is the bug?
Remove default values present in parameter descriptions as they are present in schema[default]. 

example:

   ```
 cluster.put_settings::query.flat_settings:
      in: query
      name: flat_settings
      description: 'Return settings in flat format (default: false)'
      schema:
        type: boolean
        default: false
      style: form

```

### PR DESCRIPTION
**Pull request by saimedhi**
### Description
Removed default values from parameter descriptions

### Issues Resolved
Closes #263 

### DIFF
--- a/spec/namespaces/indices.yaml
+++ b/spec/namespaces/indices.yaml
@@ -3359,7 +3359,7 @@ components:
     indices.exists_template::query.flat_settings:
       in: query
       name: flat_settings
-      description: 'Return settings in flat format (default: false)'
+      description: Return settings in flat format.
       schema:
         type: boolean
         default: false
@@ -3367,7 +3367,7 @@ components:
     indices.exists_template::query.local:
       in: query
       name: local
-      description: 'Return local information, do not retrieve the state from cluster-manager node (default: false)'
+      description: Return local information, do not retrieve the state from cluster-manager node.
       schema:
         type: boolean
         default: false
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by saimedhi** [comment]
@nhtruong, please take a look

File: spec/namespaces/indices.yaml [hunk 1]
```diff
@@ -3477,9 +3477,10 @@ components:
     indices.forcemerge::query.max_num_segments:
       in: query
       name: max_num_segments
-      description: 'The number of segments the index should be merged into (default: dynamic)'
+      description: The number of segments the index should be merged into.
       schema:
         type: number
+        default: dynamic
```
Xtansia, line 3483: This one should just be reworded rather than specifying `default:` as it's not that the value of the field will be literally the string `dynamic` (as the schema type is number), it's just that the cluster will dynamically pick the number of segments it determines to be required/ideal. The docs https://opensearch.org/docs/latest/api-reference/index-apis/force-merge/#query-parameters describe the parameter as:
> The number of larger segments into which smaller segments are merged. Set this parameter to 1 to merge all segments into one segment. The default behavior is to perform the merge as necessary.

Xtansia (changes requested):

Xtansia (approved):

PR #271, opened 2024-04-24
Adds API spec for security plugin

### LINKED ISSUE #233: [FEATURE] Add specs for `security` namespace
Part of [Adding missing API specs issue](https://github.com/opensearch-project/opensearch-api-specification/issues/168)

- [ ] /_plugins/_security/api/actiongroup/: ["GET"]
- [ ] /_plugins/_security/api/actiongroup/{name}: ["GET","PUT","DELETE"]
- [ ] /_plugins/_security/api/actiongroups/: ["PATCH"]
- [ ] /_plugins/_security/api/allowlist: ["GET","PUT","PATCH"]
- [ ] /_plugins/_security/api/audit/: ["GET","PATCH"]
- [ ] /_plugins/_security/api/authtoken: ["POST"]
- [ ] /_plugins/_security/api/cache: ["GET","PUT","POST"]
- [ ] /_plugins/_security/api/internalusers/: ["GET","PATCH"]
- [ ] /_plugins/_security/api/internalusers/{name}/authtoken: ["POST"]
- [ ] /_plugins/_security/api/migrate: ["POST"]
- [ ] /_plugins/_security/api/permissionsinfo: ["GET"]
- [ ] /_plugins/_security/api/roles/: ["PATCH"]
- [ ] /_plugins/_security/api/rolesmapping/: ["GET","PATCH"]
- [ ] /_plugins/_security/api/ssl/{certType}/reloadcerts/: ["PUT"]
- [ ] /_plugins/_security/api/tenancy/config: ["GET","PUT"]
- [ ] /_plugins/_security/api/user/: ["GET"]
- [ ] /_plugins/_security/api/user/{name}: ["GET","PUT","DELETE"]
- [ ] /_plugins/_security/api/user/{name}/authtoken: ["POST"]
- [ ] /_plugins/_security/api/validate: ["GET"]
- [ ] /_plugins/_security/api/whitelist: ["GET","PUT","PATCH"]
- [ ] /_plugins/_security/authinfo: ["GET","POST"]
- [ ] /_plugins/_security/configupdate: ["PUT"]
- [ ] /_plugins/_security/dashboardsinfo: ["GET","POST"]
- [ ] /_plugins/_security/health: ["POST"]
- [ ] /_plugins/_security/tenantinfo: ["GET","POST"]
- [ ] /_plugins/_security/whoami: ["GET","POST"]
- [ ] /_plugins/_security/whoamiprotected: ["GET"]

DarshitChanpura:
### Description
Corrects invalid entries and adds missing entries for the security plugin.

### Issues Resolved
- Resolves #233 


spec/schemas/security._common.yaml
```diff
@@ -358,3 +362,88 @@ components:
...
+      properties:
+        user:
+          type: string
+        user_name:
+          type: string
+        has_api_access:
+          type: boolean
+        disabled_endpoints:
+          type: object
+          description: An object with disabled APIs as key and array of http methods as values
```
dblock, line 449: Add a period?

DarshitChanpura:
@dblock ty for your review. This isn't ready yet. I'm doing a final pass-through the spec file to validate grammar and previously added specs. I'll mark it as ready once all changes are complete.

DarshitChanpura:
@dblock @nhtruong This PR is ready for review.

QQ, how are we planning to test the validity of all the added specs?

DarshitChanpura:
A new endpoint is in the process of being added to security plugin. I will update the spec once the PR is merged in security repo: https://github.com/opensearch-project/security/pull/4299

spec/schemas/security._common.yaml
```diff
@@ -183,60 +244,232 @@ components:
...
+            type: string
+        masked_fields:
+          type: array
+          items:
+            type: string
+        allowed_actions:
+          type: array
+          items:
+            type: string
+
```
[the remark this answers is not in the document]
dblock: Why not!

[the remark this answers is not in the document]
dblock: Actually, does this mess with editor folding of nodes? I mean in VSCode for example does it help or distract?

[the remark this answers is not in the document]
nhtruong: Actually with IDE this doesn't matter much. This is a very low priority though. @DarshitChanpura What IDE do you use and why do you prefer adding an empty line?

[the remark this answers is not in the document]
DarshitChanpura: I use VSCode and IntelliJ interchangeably. I added this on my own (accidentally ?!) when I was trying to sort the entries lexicographically. I think it would be beneficial to add a new-line between entries.

Talking about folding in VSCode and IDEA, I think it works just fine:

| VScode | IDEA | 
|---------| ----- |
Screenshot 2024-05-06 at 11 50 53 AM | Screenshot 2024-05-06 at 11 52 03 AM

spec/_superseded_operations.yaml
```diff
@@ -532,14 +562,10 @@ $schema: ../json_schemas/_superseded_operations.yaml
...
 /_opendistro/_security/kibanainfo:
-  superseded_by: /_plugins/_security/kibanainfo
+  superseded_by: /_plugins/_security/dashboardsinfo
   operations:
     - GET
     - POST
-/_opendistro/_security/sslinfo:
-  superseded_by: /_plugins/_security/sslinfo
-  operations:
-    - GET
```
nhtruong, line 542: Is this entry being removed because it's now out of sync with `/_plugins/_security/sslinfo`?

spec/namespaces/security.yaml
```diff
@@ -139,17 +339,65 @@ paths:
...
+        '501':
+          $ref: '#/components/responses/security.cache@501'
+    put:
+      operationId: security.cache.3
+      x-operation-group: security.cache
+      x-version-added: '1.0'
+      description: Not supported for cache API.
+      responses:
+        '501':
+          $ref: '#/components/responses/security.cache@501'
```
nhtruong, line 375: Should the clients ignore `security.cache` group and not generate the `client.security.cache()` method then? If so, these operations should have `x-ignorable: true`

spec/namespaces/security.yaml
```diff
@@ -629,303 +1057,371 @@ components:
...
+    security.generate_obo_token:
+      content:
+        application/json:
+          schema:
+            $ref: '../schemas/security._common.yaml#/components/schemas/OBOToken'
+      required: true
     security.patch_action_group:
       content:
-        application/x-ndjson:
+        application/json:
```
nhtruong, line 1088: Isn't the body in the form of NDJSON if the schema type is `array` instead of `object`?

spec/schemas/security._common.yaml
```diff
@@ -292,69 +602,97 @@ components:
...
+          type: string
+        is_node_certificate_request:
+          type: string
+
+    TenantInfo:
+      type: object
+      properties:
+        "<indexOrAlias>":
           type: string
+          value: "<tenant-name>"
```
nhtruong, line 698: I think think you should use [additionalProperties](https://json-schema.org/understanding-json-schema/reference/object#additionalproperties) here instead of `properties`. There are many of objects with this property in this repo you can use as examples.

spec/namespaces/security.yaml
Xtansia: [Security APIs accept an actual array not NDJSON](https://opensearch.org/docs/latest/security/access-control/api/#patch-action-groups):
```json
[
  {
    "op": "add", "path": "/CREATE_INDEX", "value": { "allowed_actions": ["indices:admin/create", "indices:admin/mapping/put"] }
  },
  {
    "op": "remove", "path": "/CRUD"
  }
]
```

nhtruong: So it's going to be comma-separated JSON?

Xtansia: @nhtruong It is valid standards compliant JSON, the top-level type can be an array it doesn't have to be an object. https://stackoverflow.com/a/3833312

nhtruong: Gotcha. I was confused by the discrepancy between this and the bulk endpoints in core namespace where NDJSON is used.

Xtansia: Yeah those ones are NDJSON which isn't valid JSON because it's not a true JSON array it's just a new object on each line to make it more streaming/memory friendly as you can parse each line independently. That's not the case here though these are true JSON arrays.

spec/_superseded_operations.yaml
DarshitChanpura: we didn't support `_plugins/_security/sslinfo`. only `_opendistro/..` is supported currently. See [here](https://github.com/opensearch-project/security/blob/905c97d4fed185362a12c4b46aa56ad8febabb4d/src/main/java/org/opensearch/security/ssl/rest/SecuritySSLInfoAction.java#L50).

spec/namespaces/security.yaml
DarshitChanpura: not sure what x-ignorable does exactly, but currently these API even though exposed do not have appropriate handlers. I've opened an issue to capture these. https://github.com/opensearch-project/security/issues/4310

Should I mark them as x-ingorable?

spec/namespaces/security.yaml
DarshitChanpura: I'll change it to json

DarshitChanpura: These should be JSON.

spec/namespaces/security.yaml
DarshitChanpura: I've marked the operations as ignorable. Check [this](https://github.com/opensearch-project/opensearch-api-specification/pull/271/files#diff-7b266d1960855855ef31c24d1c2e983de7d60d31afd7927a04dfab0384a0362dR355) out.

nhtruong: It's one of the [extensions](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md#openapi-extensions) that we incorporate into the spec to add extra data to the spec that OpenAPI does not natively support. `x-ignorable` tells the client generators to skip that operation and not generate a method for it.

spec/schemas/security._common.yaml
nhtruong: I know what you mean. In JSON schema however, `properties` describes fix/predefined keys (similar to a structure in Smithy) while `additionalProperties` describes free-form keys (similar to a map in Smithy). You can actually use both in an object to describe map with some reserved keys.

The schema
```yaml
TenantInfo:
      type: object
      properties:
        "<indexOrAlias>":
          type: string
          value: "<tenant-name>"
```
describes an object with exactly one key called `<indexOrAlias>` with value of type string (`value: "<tenant-name>"` will be ignored because `value` is not part of JSON schema)
You can either do:
```
TenantInfoMap:
      type: object
      additionalProperties:
            type: string        
```
which is basically a map with keys of type string (the default schema for keys) and value of type string as well.
Or if you really want to describe this entity with the highest resolution possible (though might not be necessary):
```yaml       
TenantInfoMap:
     additionalProperties:
          $ref: #/components/schemas/TenantName
     oneOf:
          - type: object
             propertyNames:
                  $ref: #/components/schemas/Index
          - type: object
             propertyNames:
                  $ref: #/components/schemas/Alias
TenantName:
     type: string
Index:
     type: string
Alias:
     type: string
```
This schema is identical to the one above as far as schema validation is concerned (basically a map with both keys and values as string) but it provides much more information about the meanings of the keys and the values. Plus, you can expand the `TenantName`, `Index` and `Alias` to give you info via `description` or `title` or add restrictions to them via `pattern`, `enum` and so on.

DarshitChanpura: > describes an object with exactly one key called <indexOrAlias> with value of type string 

in this case it could be more than 1 entries so, 
```
TenantInfo:
      type: object
      additionalProperties:
            type: string        
```
would suffice. Let me know if this needs more resolution.

nhtruong: We can always up the resolution in the future or simply add a `description` for TenantInfo. We also have Generated Docs is our roadmap so any extra info added will help with that future project.

nhtruong (approved):

nhtruong:
@Xtansia This looks good to me but would love a 2nd pair of eyes to review this.

Xtansia (approved):
LGTM

DarshitChanpura:
> @DarshitChanpura We have a new test framework in added in https://github.com/opensearch-project/opensearch-api-specification/pull/299, maybe you'll be interested in writing tests for the security plugin using it?

Yes, I'll take a stab at it.

PR #280, opened 2024-05-03
Fixed missing global params

nhtruong:
This bug happens when the original name of the params in `_global_parameters.yaml` are identical to the games they will be generated in the result file which has the form of `_global::query.{name}`.  During the generation the original name was removed and replaced by the global name. Swapping the order of key deleting and assignment solved this. Also updated the names in the `_global_parameters.yaml` to reflect that the merger will handle the naming convention to keep this file simple.


dblock (approved):
This should have a test, but won't hold the PR for the fix.

nhtruong:
@dblock There are [tests](https://github.com/opensearch-project/opensearch-api-specification/blob/main/tools/test/merger/fixtures/expected.yaml) that assure that the global params are add. This was just a very edge case

dblock:
@nhtruong You'll agree that no tests covered this case, this or another edge case can reintroduce the problem.

PR #283, opened 2024-05-06
Rollups API spec

djmadeira:
### Description
Add specs for rollup APIs


package-lock.json
```diff
@@ -1,6 +0,0 @@
-{
-  "name": "opensearch-api-specification",
-  "lockfileVersion": 3,
-  "requires": true,
-  "packages": {}
-}
```
dblock: Should we be running an `npm install` to lock dependencies? Or exclude this file in .gitignore. Either way probably a separate PR.

dblock (changes requested):
Thanks! Can you please sign DCO, `git commit -s --amend`.

djmadeira:
@dblock done

djmadeira:
@dblock looks like it's passing now

dblock:
@nhtruong or @Xtansia take a look at the spec?

nhtruong:
This is great, Thanks @djmadeira
Regarding the `x-operation-group`, instead of `rollups.get_rollup`, `rollups.delete_rollup`... we should use `rollups.get`, `rollups.delete` instead to avoid repeating the world rollup as all of these actions are in the `rollups` namespace. We have a similar naming convention in the `indices` namespace with `indices.create`, `indices.delete` and so on.

djmadeira:
@nhtruong addressed feedback

nhtruong:
@djmadeira looks like some `x-operation-group` haven't not been updated. And there are 2 `transforms.get@200` entries and they caused some workflows to fail.

djmadeira:
@nhtruong looks like I got some wires crossed between the two PRs. fixed now.

nhtruong (approved):

PR #286, opened 2024-05-07
Added testing to the generated OpenAPI.

### LINKED ISSUE #277: [PROPOSAL] Add integration tests
### What are you proposing?

Coming from https://github.com/opensearch-project/opensearch-api-specification/pull/271#issuecomment-2088669972, we don't actually know whether the specs authored here are correct. Add integration tests that exercise these specs. 

### What users have asked for this feature?

See https://github.com/opensearch-project/opensearch-api-specification/pull/271#issuecomment-2088669972.

### What problems are you trying to solve?

- Test that the specs authored match what OpenSearch code is doing.
- Early signal if OpenSearch is breaking specs.

### What is the developer experience going to be?

Every PR will start various versions of OpenSearch, load the specs, run tests against the local version of OpenSearch.

### What will it take to execute?

Unsure. Maybe there's an existing tool? Otherwise I would start with something simple that executes all `GET` requests and compares outputs against a previous recording. Then author requests/responses and compare. Finally, we need coverage to know how many of the APIs are invoked, how many request/response bodies are matched.

dblock:
### Description

This adds tests using [dredd](https://dredd.org).

1. Builds the spec that only includes APIs with `x-tested`. This is a new feature of the merger than can be extended in the future to exclude deprecated APIs, or ones for a specified hosted target (AOS, AOSS, etc.), or version of OpenSearch.
2. Starts an OpenSearch docker container.
3. Runs dredd against it for the APIs present in the specs.

Passing spec.

```
pass: GET (200) / duration: 90ms
complete: 1 passing, 0 failing, 0 errors, 0 skipped, 1 total
complete: Tests took 15136ms
```

Failing spec:

```
fail: GET (200) / duration: 75ms
info: Displaying failed tests...
fail: GET (200) / duration: 75ms
fail: body: At '/version/build_flavor' Missing required property: build_flavor

request: 
method: GET
uri: /
headers: 
    Accept: application/json; charset=UTF-8
    User-Agent: Dredd/14.1.0 (Darwin 23.4.0; arm64)
    Authorization:  Basic YWRtaW46YWRtaW4=

body: 



expected: 
headers: 
    Content-Type: application/json; charset=UTF-8

body: 
{
  "cluster_name": "",
  "cluster_uuid": "",
  "name": "",
  "tagline": "",
  "version": {
    "build_date": "",
    "build_flavor": "",
    "build_hash": "",
    "build_snapshot": false,
    "build_type": "",
    "lucene_version": "",
    "minimum_index_compatibility_version": "",
    "minimum_wire_compatibility_version": "",
    "number": ""
  }
}
statusCode: 200


actual: 
statusCode: 200
headers: 
    content-type: application/json; charset=UTF-8
    content-length: 566

bodyEncoding: utf-8
body: 
{
  "name": "2c8264d2ffb3",
  "cluster_name": "docker-cluster",
  "cluster_uuid": "grah_e-CRTii2hv1unQ1bg",
  "version": {
    "distribution": "opensearch",
    "number": "2.9.0",
    "build_type": "tar",
    "build_hash": "1164221ee2b8ba3560f0ff492309867beea28433",
    "build_date": "2023-07-18T21:22:48.164885046Z",
    "build_snapshot": false,
    "lucene_version": "9.7.0",
    "minimum_wire_compatibility_version": "7.10.0",
    "minimum_index_compatibility_version": "7.0.0"
  },
  "tagline": "The OpenSearch Project: https://opensearch.org/"
}

complete: 0 passing, 1 failing, 0 errors, 0 skipped, 1 total
complete: Tests took 83ms
```

In order to get `GET /` to pass we needed to supply an example of a response, otherwise dredd thinks that `build_flavor` is a required parameter (see https://github.com/apiaryio/dredd/issues/1430).

### Issues Resolved

Resolves #277 


dblock:
The spec is pretty far from passing dredd, but I'll keep chipping at it. @Xtansia @nhtruong LMK what you think about this approach?

dblock:
This is ready to review @Xtansia @nhtruong. Leaving as draft so I can add at least some tests for HEAD/PUT/POST before adopting dredd.

tools/merger/OpenApiMerger.ts
```diff
@@ -119,4 +122,23 @@ export default class OpenApiMerger {
...
+  #filter_methods_with_key (paths: any, key: string): void {
+    Object.entries(this.spec.paths as Document).forEach(([path, path_item]) => {
+      Object.entries(path_item as Document).forEach(([method, method_spec]) => {
+        if (method_spec[key] === undefined) {
+          delete (this.spec.paths[path][method])
+          if (Object.keys(this.spec.paths[path] as Document).length === 0) {
+            delete (this.spec.paths[path])
+          }
+        }
+      })
```
dblock, line 141: @nhtruong if you want to help me rewrite this without `delete` I would gladly re-enable `no-dynamic-delete` - my typescript-fu did not cut it with `.filter` and friends.

tools/merger/OpenApiMerger.ts
```diff
@@ -46,6 +48,7 @@ export default class OpenApiMerger {
       const spec = read_yaml(`${folder}/${file}`)
       this.redirect_refs_in_namespace(spec)
       this.spec.paths = { ...this.spec.paths, ...spec.paths }
+      this.filter_methods(spec)
```
Xtansia, line 51:
```suggestion
      this.spec.paths = { ...this.spec.paths, ...this.filter_paths(spec.paths as OpenAPIV3.PathsObject) }
```

tools/merger/OpenApiMerger.ts
```diff
@@ -119,4 +122,23 @@ export default class OpenApiMerger {
...
+      Object.entries(path_item as Document).forEach(([method, method_spec]) => {
+        if (method_spec[key] === undefined) {
+          delete (this.spec.paths[path][method])
+          if (Object.keys(this.spec.paths[path] as Document).length === 0) {
+            delete (this.spec.paths[path])
+          }
+        }
+      })
+    })
+  }
```
Xtansia, line 143:
```suggestion
  filter_paths (paths: OpenAPIV3.PathsObject): OpenAPIV3.PathsObject {
    for (const x_include_value of this.options.x_include) {
      paths = this.#filter_operations_with_key(paths, x_include_value)
    }
    return paths
  }

  #filter_operations_with_key (paths: OpenAPIV3.PathsObject, key: string): OpenAPIV3.PathsObject {
    const entries = Object.entries(paths)
      .map(([path, path_item]) => {
        const entries = Object.entries(path_item as Record<string, any>)
          .filter(([op_key, value]) => !is_http_method(op_key) || value[key] !== undefined)
        return [path, Object.fromEntries(entries)]
      })
      .filter(([_, path_item]) => Object.keys(path_item).filter(is_http_method).length > 0)
    return Object.fromEntries(entries)
  }
```

Xtansia (commented):
@dblock I've added some suggestions on how to do the filtering without in-place deletes, also avoids the assumption that all properties of a path object are http methods, there can be a path level description or parameters as well even if we don't currently use them.

Requires the following added to `helpers.ts`:
```typescript
import { OpenAPIV3 } from 'openapi-types'

export const OPENAPI_HTTP_METHODS: OpenAPIV3.HttpMethods[] = Object.values(OpenAPIV3.HttpMethods)

export function is_http_method (str: string): str is OpenAPIV3.HttpMethods {
  return OPENAPI_HTTP_METHODS.includes(str as OpenAPIV3.HttpMethods)
}
```

tools/merger/OpenApiMerger.ts
nhtruong: It's a good lint to have to prevent accidental deletions. But I think it's more prevalent to cases where the feature is heavily influenced  by the user input, as it's rather unpredictable. If you know that this has a very low risk of accidental deletions, we can disable the lint for the specific line like [this](https://github.com/opensearch-project/opensearch-api-specification/blob/52463cedb11012b33478c7ae01efb90adb18dd9c/tools/helpers.ts#L40) and leave the lint rule on.

spec/namespaces/cat.yaml
```diff
@@ -842,11 +843,46 @@ components:
     cat.help@200:
       description: ''
       content:
-        application/json:
+        text/plain:
```
nhtruong, line 846: Created an issue for this: https://github.com/opensearch-project/opensearch-api-specification/issues/293

spec/namespaces/_core.yaml
```diff
@@ -9,6 +9,7 @@ paths:
       operationId: info.0
       x-operation-group: info
       x-version-added: '1.0'
+      x-tested: true
```
nhtruong, line 12: We should add `x-tested` to this [list](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md#openapi-extensions) too :)

tools/eslint.config.mjs
```diff
@@ -48,6 +48,7 @@ export default [
           allowRuleToRunWithoutStrictNullChecksIKnowWhatIAmDoing: false
         }
       ],
+      '@typescript-eslint/no-var-requires': 'off',
```
nhtruong, line 51: Was this triggered by `const hooks = require('hooks')` in `tests/all.js` file?

tools/merger/OpenApiMerger.ts
dblock: @Xtansia helped me with the code, newer revisions put this back

spec/namespaces/cat.yaml
dblock: Thanks! Tests caught the bug. Which is good news.

tools/eslint.config.mjs
nhtruong: This lint should only be applied to typescript files. We should disable it for `js` files only: https://github.com/Shopify/eslint-plugin-shopify/issues/159#issuecomment-485813838

dblock: That works! 

https://eslint.org/blog/2022/08/new-config-system-part-2/

dblock:
@nhtruong I added POST and DELETE specs, which required correcting the response code from POST to be 201, but now the linter complains that the responses for all `index` should be the same.

```
$ npm run lint:spec

> opensearch_api_tools@1.0.0 lint:spec
> ts-node tools/src/linter/lint.ts

Validating /Users/dblock/source/opensearch-project/opensearch-api-specification/dblock-opensearch-api-specification/spec ...
Errors found:

{
  file: 'namespaces/_core.yaml',
  location: 'Operation Group: index',
  message: "3 'index' operations must have an identical set of responses."
}

Total errors: 1
```

Help me correct this the right way?

nhtruong:
Regarding: https://github.com/opensearch-project/opensearch-api-specification/pull/286#issuecomment-2108698322https://github.com/opensearch-project/opensearch-api-specification/pull/286#issuecomment-2108698322

We'll need to update the linter to accept a list of exceptions OR update this lint rule to consider 200 and 201 to be equal. We can't get rid of this lint because the client generators expect operations of the same group to share certain properties, and `responses` is an important one.

Xtansia:
You can use `2XX` as the key to just say it'll be some successful response code but not specific (https://swagger.io/specification/#responses-object), though that loses some fidelity

dblock:
Closing in favor of #299

PR #287, opened 2024-05-08
Specify charset returned from OpenSearch.

dblock:
### Description

Coming from https://github.com/opensearch-project/opensearch-api-specification/pull/286 where I started running dredd against the spec. It fails with the following.

```
fail: headers: At '/content-type' No enum match for: "application/json; charset=utf-8"

expected: 
headers: 
    Content-Type: application/json

actual: 
statusCode: 200
headers: 
    content-type: application/json; charset=UTF-8
```

In this PR we match the charset correctly in the spec.

I *think* this is right, but I am not sure.

### Issues Resolved

### DIFF
--- a/spec/namespaces/_core.yaml
+++ b/spec/namespaces/_core.yaml
@@ -2110,7 +2110,7 @@ components:
   requestBodies:
     bulk:
       content:
-        application/x-ndjson:
+        application/x-ndjson; charset=utf-8:
           schema:
             type: array
             items:
@@ -2122,7 +2122,7 @@ components:
       required: true
     clear_scroll:
       content:
-        application/json:
+        application/json; charset=utf-8:
           schema:
             type: object
             properties:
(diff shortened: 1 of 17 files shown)


Xtansia:
Was wondering why this wasn't an issue with the previous dredd setup, turns out we just explicitly changed the expectation in the hooks files https://github.com/opensearch-project/opensearch-api-specification/blob/c256d99275990778e8686540d9b1371e450a1791/test/models/cat/indices/hooks.js#L24

dblock:
> Was wondering why this wasn't an issue with the previous dredd setup, turns out we just explicitly changed the expectation in the hooks files
> 
> https://github.com/opensearch-project/opensearch-api-specification/blob/c256d99275990778e8686540d9b1371e450a1791/test/models/cat/indices/hooks.js#L24

Interesting! I couldn't find any explicit needs to specify the charset in the spec, so I am going to do the above hack.

PR #292, opened 2024-05-13
Use the standard src/tests paths, move package.json to the root.

dblock:
### Description

In #286 I am adding tests and things are getting messy having to deal with relative paths for tooling (for example, I'd like to lint .js files in tests that are dredd hooks). This PR adopts a more standard layout of `tools` and `tests` and moves `package.json` to the root, making it easier for developers to work in the repo. 

A nice side-effect is that we can stop using relative paths in imports in tests by adding `baseUrl` into `tsconfig.json` and `moduleDirectories: ['node_modules', 'tools']` into `jest.config.js`.


nhtruong:
Is this specific to `dredd`? As in can't it be placed in `tools` folder?

dblock:
> Is this specific to `dredd`? As in can't it be placed in `tools` folder?

It's not. I think working in this repo is unnecessary cumbersome because you have to use relative paths for includes, and having to always remember to `cd tools` to run them. Tools then have to default to paths above `tools` (e.g. default output).

When adding spec tests (dredd) we run into an additional problem. Dredd has files such as hooks.js and we will place them into `tests/spec` (since these are tests for the spec). So now we want to run a linter on those files, but the package.json for the linter lives in ... tools, which is weird.

dblock:
@nhtruong if your issue is the name `src`, I could rename it back to `tools`, but still move `package.json` to the root. LMK what you think.

nhtruong:
> > Is this specific to `dredd`? As in can't it be placed in `tools` folder?
> 
> It's not. I think working in this repo is unnecessary cumbersome because you have to use relative paths for includes, and having to always remember to `cd tools` to run them. Tools then have to default to paths above `tools` (e.g. default output).
> 
> When adding spec tests (dredd) we run into an additional problem. Dredd has files such as hooks.js and we will place them into `tests/spec` (since these are tests for the spec that lives at the root). So now we want to run a linter on those files, but the package.json for the linter lives in ... tools, which is weird.
> 
> I could move tools into src/tools/ instead of just src, and still move package.json to the root, but that seems unnecessary.

I see. That makes sense. I'm in!

nhtruong:
This is not a `nodes.js` repo so I prefer `tools` over `src` as it tells us that applications in the folder are tools for the repo, not the focal point of the repo.

dblock:
> This is not a `nodes.js` repo so I prefer `tools` over `src` as it tells us that applications in the folder are tools for the repo, not the focal point of the repo.

Ok, and where do you want the tests for tools? B or C?

nhtruong:
Tests for the tools should be in `tools`. Test for the spec itself should be at the root of the repo.

nhtruong:
Btw I hope your IDE gives you a hand with renaming/moving files because all of these changes would be very painful without!

dblock:
> Tests for the tools should be in `tools`. Test for the spec itself should be at the root of the repo. (So C-ish if that `tests` folder only refers to the tests for tools)

Ok. Here's what I did:

1. Moved package.json to root.
2. Moved `tools/linter` and `tools/merger` to `tools/src`, otherwise `tools/test` looks like a tool.
3. Moved `tools/test` to `tools/tests`.
4. No more relative includes.

@nhtruong

nhtruong (approved):

PR #294, opened 2024-05-13
Validate all schemas defined in the spec as valid JSON schemas

nhtruong:
- Updated default arg values for merge.ts
- Added silencing feature for merger tool
- Added SchemasValidator


tools/src/linter/SchemasValidator.ts
```diff
@@ -0,0 +1,100 @@
...
+  ajv: AJV
+
+  constructor (root_folder: string) {
+    this.root_folder = root_folder
+    this.ajv = new AJV()
+    addFormats(this.ajv)
+  }
+
+  validate (): ValidationError[] {
+    this.spec = new OpenApiMerger(this.root_folder, true).merge().components as Record<string, any>
```
Xtansia, line 18: What's the reasoning for needing to run against the merged output? 

I naively would think something like below would work:
```typescript
import AJV from 'ajv'
import addFormats from 'ajv-formats'
import fs from 'fs'
import { type OpenAPIV3 } from 'openapi-types'
import { type ValidationError } from '../types'
import { SchemaVisitor } from './utils/SpecificationVisitor'
import { is_ref, type MaybeRef, SpecificationContext } from './utils'
import { read_yaml } from '../../helpers'

export default class SchemasValidator {
  root_folder: string
  ajv: AJV

  constructor (root_folder: string) {
    this.root_folder = root_folder
    this.ajv = new AJV()
    addFormats(this.ajv)
  }

    const errors: ValidationError[] = []

    const visitor = new SchemaVisitor((ctx, schema) => {
      this.#validate_schema(ctx, schema, errors)
    })

    const list_folder = (subdir: string): string[] =>
      fs.readdirSync(`${this.root_folder}/${subdir}`)
        .filter((f) => f.endsWith('.yaml'))
        .map((f) => `${subdir}/${f}`)

    const files: string[] = [
      '_global_parameters.yaml',
      ...list_folder('namespaces'),
      ...list_folder('schemas')
    ].sort()

    for (const file of files) {
      visitor.visit_specification(new SpecificationContext(file), read_yaml(`${this.root_folder}/${file}`) as OpenAPIV3.Document)
    }

    return errors
  }

  #validate_schema (ctx: SpecificationContext, schema: MaybeRef<OpenAPIV3.SchemaObject>, errors: ValidationError[]): void {
    if (is_ref(schema)) return

    try {
      // eslint-disable-next-line @typescript-eslint/no-floating-promises
      this.ajv.validateSchema(schema, true)
    } catch (err) {
      errors.push(ctx.error((err as Error).message))
    }
  }
}
```

tools/src/merger/OpenApiMerger.ts
```diff
@@ -9,11 +9,13 @@ import GlobalParamsGenerator from './GlobalParamsGenerator'
...
   root_folder: string
   spec: Record<string, any>
+  silent: boolean
 
   paths: Record<string, Record<string, OpenAPIV3.PathItemObject>> = {} // namespace -> path -> path_item_object
   schemas: Record<string, Record<string, OpenAPIV3.SchemaObject>> = {} // category -> schema -> schema_object
 
-  constructor (root_folder: string) {
+  constructor (root_folder: string, silent: boolean = false) {
+    this.silent = silent
```
dblock, line 18: Passing `silent` around is going to get messy. Follow the logger pattern in most applications - make a logger class and pass that around (ideally assign it in the constructor). The logger should have an `info`, `warn` and `error` method that checks log level. Then you can implement things like `--verbose` or `--silent` and the way output is rendered without having to change everything in the code.

tools/src/merger/SupersededOpsGenerator.ts
```diff
@@ -11,14 +11,14 @@ export default class SupersededOpsGenerator {
...
+  generate (spec: Record<string, any>, silent: boolean): void {
     for (const [path, { superseded_by, operations }] of _.entries(this.superseded_ops)) {
       const regex = this.path_to_regex(superseded_by)
       const operation_keys = operations.map(op => op.toLowerCase())
       const superseded_path = this.copy_params(superseded_by, path)
       const path_entry = _.entries(spec.paths as Document).find(([path, _]) => regex.test(path))
-      if (!path_entry) console.log(`Path not found: ${superseded_by}`)
-      else spec.paths[superseded_path] = this.path_object(path_entry[1], operation_keys)
+      if (path_entry != null) spec.paths[superseded_path] = this.path_object(path_entry[1], operation_keys)
+      else if (!silent) console.log(`Path not found: ${superseded_by}`)
```
dblock, line 21: This becomes 

`if (path_entry != null) ... else logger.warn(...)`

tools/src/linter/SchemasValidator.ts
nhtruong: This was my original strat at first. But say if you want to validate:
```yaml
# file: schemas/_common.yaml

type: array
items:
  $ref: #/components/schemas/Index
```

Then ajv must be made aware of they key `#/components/schemas/Index` beforehand with
```ts
ajv.addSchema(index_schema, '#/components/schemas/Index')
```

This gets complicated when the `Index` schema is referenced in different files:
- The key is `#/components/schemas/Index` in the `schemas/_common.yaml` file itself
- The key is `_common.yaml#/components/schemas/Index` in a namespace file
- The key is `../schemas/_common.yaml#/components/schemas/Index` in a namespace file

Meaning that whenever you want to validate an individual file, you will have to load up all schemas files, and add the referenced schemas using `ajv.addSchema` accordingly. It's too complicated and not as efficient as loading up the single-file spec.

nhtruong:
Converting this to draft because I've just discovered that I can strict mode on and it's catching a lot more infractions in our current spec. Look into that now

spec/schemas/_common.yaml
```diff
@@ -20,7 +20,6 @@ components:
       description: |-
         A duration. Units can be `nanos`, `micros`, `ms` (milliseconds), `s` (seconds), `m` (minutes), `h` (hours) and
         `d` (days). Also accepts "0" without a unit and "-1" to indicate an unspecified value.
-      x-data-type: time
```
nhtruong, line 23: @Xtansia removing `x-data-type`. Do you still need this for .Net Generator? There's only 1 time schema called `Duration` and all parameters using this datatype reference this schema.

nhtruong:
> API specs implemented for 307/649 (47%) APIs.

@dblock These comments are getting out of hand 😅 There should be a cool down period, or maybe we should only triggered when the PR is first created and/or run it constantly like this when a certain tag is attached to the PR?

dblock:
> > API specs implemented for 307/649 (47%) APIs.
> 
> @dblock These comments are getting out of hand 😅 There should be a cool down period, or maybe it should only be triggered when the PR is first created and/or ONLY run it constantly like this when a certain tag is attached to the PR?

For sure. It could delete the previous comment/update the existing one, etc. Feel free to open an issue.

spec/schemas/_common.yaml
Xtansia: It's currently used, but the .NET generator doesn't automatically download or re-run against the new spec so won't break anything immediately and can be easily fixed in my next generation pass. So no worries from my side about dropping it 👍

nhtruong:
@dblock @Xtansia This is ready for review again

spec/schemas/cat.nodes.yaml
```diff
@@ -101,11 +101,10 @@ components:
           type: string
         master:
           description: |-
+            (Deprecated in 2.0.0. To promote inclusive language, use 'cluster_manager' instead.)
             Indicates whether the node is the elected master node.
             Returned values include `*`(elected master) and `-`(not elected master).
           type: string
-          x-version-deprecated: 2.0.0
-          x-deprecation-message: To promote inclusive language, use 'cluster_manager' instead.
```
Xtansia, line 108: What's the reason for this change? It seems like a backwards step and degradation for our codegen use case (ie. this should eventually get annotated as `@Deprecated` in Java & `[Obsolete]` in C#)

tools/src/logger.ts
```diff
@@ -0,0 +1,25 @@
...
+  log (level: LogLevel, message: string): void {
+    console.log(`[${level}] ${message}`)
+  }
+
+  info (message: string): void {
+    if (this.level === 'info') this.log('info', message)
+  }
+
+  warn (message: string): void {
+    if (this.level === 'info' || this.level === 'warn') this.log('warn', message)
```
dblock, line 19: console has `warn` if I am not mistaken, so should this be `console.warn`?

tools/src/logger.ts
```diff
@@ -0,0 +1,25 @@
+export type LogLevel = 'info' | 'warn' | 'error'
+
+export class Logger {
+  level: string
```
dblock, line 4: Turn level into an integer/enum where info > warn > debug > trace and compare numbers. So if logger level is info, you also want to see warnings.

dblock (commented):
It's not a lot of changes, but you might want to split up logger changes.

spec/schemas/cat.nodes.yaml
nhtruong: This is the only case where these extensions are used in a schema (They only appear in parameters otherwise). So, I thought it was a misuse. Can add these to the validator's vocabulary. Are there other extensions do you think should be added to the vocabulary as well?

nhtruong: Added [these](https://github.com/opensearch-project/opensearch-api-specification/pull/294/files#diff-e154da95e179db1faa593dbfd567bccc4f91cf6620672188f6e62c564c88f716R12-R17)

tools/src/Logger.ts
```diff
@@ -0,0 +1,33 @@
+export enum LogLevel {
+  info = 1,
+  warn = 2,
+  error = 3
```
dblock, line 4: In all the loggers the order is usually the other way around, you should flip these.

https://stackoverflow.com/questions/2031163/when-to-use-the-different-log-levels

dblock (approved):
Write tests for logger, flip order. Sorry to be annoying ;) Otherwise this PR LGTM.

tools/src/Logger.ts
nhtruong: Ahhh Today I learned. Thanks

nhtruong: It's been updated now. I've squashed all commits too. Feel free to merge it after approving.

dblock (approved):

PR #299, opened 2024-05-21
Test Framework

nhtruong:
This PR adds a custom test framework for OpenSearch Open API Spec. Each YAML file in the `./tests` folder (and its sub-folders) is a [test story](https://github.com/opensearch-project/opensearch-api-specification/pull/299/files#diff-174d29e4dc9217a0bb386de2e0005fa82849de0a518ce9f0bff54db8d5a529ba) with 3 main components:
- prologues: An array of requests made to the OpenSearch cluster to prepare it for the test (e.g. create and index and a few docs to test the search operations)
- chapters: An array of requests, run sequentially, that actually test the spec:
  - Test if the operation exists in the spec
  - Test the provided parameters against their schemas in the spec
  - Test the provided request body against its schema in the spec
  - Test if the response code matches the expected response code
  - Test the response payload against its schema in the spec
- epilogues: An array of requests co clean up the test environment (e.g. delete all indices and documents created during the test)

A [JSON Schema](https://github.com/opensearch-project/opensearch-api-specification/pull/299/files#diff-e510f040270ce54f07291232cd654ef659812b19b04654275887846ed5a0841f) for the test stories is also provided so that the IDE can help developers with autocomplete and validation when drafting the stories.

Make sure you have an OpenSearch cluster running at `http://localhost:9200` before running the following command to initiate the test suite
```bash
npm run test:spec
```

Bellow is the result of running `npm run test:spec -- --ignore_passed N`
Check [start.ts](https://github.com/opensearch-project/opensearch-api-specification/pull/299/files#diff-2c08c535b96bdd6d179c3ba64bbe42eb5457dcfabb9d4c0b9407dcd19e9fe19b) to see the options that can be passed to `npm run test:spec` command.

image

Remaining work:
- [x] Grab server URL from environment variable and add authentication.
- [x] Add a github workflow
- [ ] Update repo docs
- [ ] Add tests fro the framework (heavy lift)
- [ ] Test should fail when a field is found in the response but not defined in the spec


dblock (commented):
This looks good overall! Needs READMEs, tests, and all the TODOs. 

Left small usage/naming comments.

Try to get something that runs under CI first in.

tools/src/tester/ChapterEvaluator.ts
```diff
@@ -0,0 +1,67 @@
...
+export default class ChapterEvaluator {
+  chapter: Chapter
+  result: Result = Result.PASSED
+  skip_payload: boolean = false
+
+  constructor (chapter: Chapter) {
+    this.chapter = chapter
+  }
+
+  async evaluate (skipped: boolean): Promise<ChapterEvaluation> {
```
Xtansia, line 15: I'd rename `skipped` to `skip`

tools/src/tester/StoryEvaluator.ts
```diff
@@ -0,0 +1,78 @@
...
+      description: this.story.description,
+      prologues: await this.#evaluate_supplemental_chapters(this.story.prologues),
+      chapters: await this.#evaluate_chapters(this.story.chapters),
+      epilogues: await this.#evaluate_supplemental_chapters(this.story.epilogues),
+      result: this.result
+    }
+  }
+
+  async #evaluate_chapters (chapters: Chapter[]): Promise<ChapterEvaluation[]> {
+    let skipped: boolean = this.skipped
```
Xtansia, line 46: I think this should be `has_errors` rather than `skipped` to be clearer about purpose

tools/src/tester/ResultsDisplayer.ts
```diff
@@ -0,0 +1,110 @@
...
+
+  constructor (evaluation: StoryEvaluation, gap: number = 4) {
+    this.evaluation = evaluation
+    this.skip_components = [Result.PASSED, Result.SKIPPED].includes(evaluation.result)
+    this.gap = gap
+  }
+
+  display (ignore_skipped = false, ignore_passed = true): void {
+    this.ignore_passed = ignore_passed
+    this.ignore_skipped = ignore_skipped
```
Xtansia, line 34: Given the constructor and display are called next to each other, these should be passed to the constructor if they're going to be assigned to member fields.

tools/src/tester/TestsRunner.ts
```diff
@@ -0,0 +1,65 @@
...
+import ResultsDisplayer from './ResultsDisplayer'
+
+declare global {
+  // eslint-disable-next-line no-var
+  var chapter_reader: ChapterReader
+  // eslint-disable-next-line no-var
+  var schema_validator: SchemaValidator
+  // eslint-disable-next-line no-var
+  var spec_parser: SpecParser
+}
```
Xtansia, line 19: Using globals is a code smell to me and would prefer passing instances via constructors, especially given these are being built and assigned from the constructor below.

tools/src/tester/TestsRunner.ts
```diff
@@ -0,0 +1,65 @@
...
+    return (await Promise.all(promises)).sort((a, b) => a.display_path.localeCompare(b.display_path))
+  }
+
+  #collect_story_files (folder: string, file: string, prefix: string): StoryFile[] {
+    const path = file === '' ? folder : `${folder}/${file}`
+    const next_prefix = prefix === '' ? file : `${prefix}/${file}`
+    if (fs.statSync(path).isFile()) {
+      const story = read_yaml(path) as Story
+      return [{
+        display_path: next_prefix === '' ? (path.split('/').reverse()[0]) : next_prefix,
```
Xtansia, line 55: This could just use [path.basename(path)](https://nodejs.org/api/path.html#pathbasenamepath-suffix)

tools/src/tester/TestsRunner.ts
```diff
@@ -0,0 +1,65 @@
...
+  async run (ignore_skipped = false, ignore_passed = true): Promise<void> {
+    const evaluations = await this.evaluate()
+    for (const evaluation of evaluations) {
+      const displayer = new ResultsDisplayer(evaluation)
+      displayer.display(ignore_skipped, ignore_passed)
+    }
+  }
+
+  async evaluate (): Promise<StoryEvaluation[]> {
+    const story_files = this.#collect_story_files(`${process.cwd()}/${this.path}`, '', '')
```
Xtansia, line 41: This would be safer to use [`path.resolve(this.path)`](https://nodejs.org/api/path.html#pathresolvepaths) so that it handles the case where `this.path` might already be an absolute path

tests/index_lifecycle.yaml
```diff
@@ -0,0 +1,85 @@
...
+      status: 200
+
+  - synopsis: Check if the index `movies` exists. It should not.
+    path: /{index}
+    method: HEAD
+    parameters:
+      index: movies
+    response:
+      status: 404
+
```
dblock, line 53: We also should test a `GET` that returns a 404 with details (incomplete schema below).

```
@@ -2314,6 +2322,18 @@ components:
             type: object
             additionalProperties:
               $ref: '../schemas/indices._common.yaml#/components/schemas/IndexState'
+    indices.get@404:
+      description: ''
+      content:
+        application/json:
+          schema:
+            type: object
+            properties:
+              status:
+                description: HTTP status code returned for the operation.
+                type: number
+              error:
+                $ref: '../schemas/_common.yaml#/components/schemas/ErrorCause'
     indices.get_alias@200:
       description: ''
       content:
diff --git a/tests/index_lifecycle.yaml b/tests/index_lifecycle.yaml
index 7ea4296..8e078ff 100644
--- a/tests/index_lifecycle.yaml
+++ b/tests/index_lifecycle.yaml
@@ -51,6 +51,14 @@ chapters:
     response:
       status: 404
 
+  - synopsis: Check if the index `movies` exists. It should not and return details.
+    path: /{index}
+    method: GET
+    parameters:
+      index: movies
+    response:
+      status: 404
+
```

We should have an error/warning for additional fields missing in spec, in the above example the entire error has things like `index_uuid`:

```json
{
  "error": {
    "root_cause": [
      {
        "type": "index_not_found_exception",
        "reason": "no such index [movies]",
        "index": "movies",
        "resource.id": "movies",
        "resource.type": "index_or_alias",
        "index_uuid": "_na_"
      }
    ],
    "type": "index_not_found_exception",
    "reason": "no such index [movies]",
    "index": "movies",
    "resource.id": "movies",
    "resource.type": "index_or_alias",
    "index_uuid": "_na_"
  },
  "status": 404
}
```

tests/index_lifecycle.yaml
```diff
@@ -0,0 +1,85 @@
+$schema: ../../json_schemas/story.schema.yaml
+
+description: This story tests all endpoints relevant the lifecycle of an index, from creation to deletion.
+prologues: []
+epilogues:
```
dblock, line 5: This story's epilogue will fail since movies doesn't exist and the games and books indices are being deleted as part of the chapters. Generally I think cleaning up after a test is the right thing because it avoids unnecessarily polluting the next test, and I suppose we can ignore the errors, but maybe there's a generic cleanup we can do between stories like delete all indices, snapshots, settings, etc.... just thinking out loud.

dblock (commented):
Tested this out.

tests/index_lifecycle.yaml
```diff
@@ -0,0 +1,85 @@
...
+  - path: /books,movies,games
+    method: DELETE
+    ignore_errors: false
```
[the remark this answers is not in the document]
nhtruong: Yes it's `true` by default. I just have it explicitly defined so I can toggle it on and off for testing :)

tests/index_lifecycle.yaml
nhtruong: Having stored procedures (borrowing a word from SQL land) that delete all indices and such is a brilliant idea. We should save it for a follow up PR though.

I think `ignore_errors` option is still necessaries for the cleanup to account for when not all chapters are run and we don't have a stored procedure that can clean up after said test.

tools/src/tester/start.ts
```diff
@@ -0,0 +1,21 @@
+import OpenApiMerger from '../merger/OpenApiMerger'
+import { LogLevel } from '../Logger'
+import TestsRunner from './TestsRunner'
+import { Command, Option } from '@commander-js/extra-typings'
+
+const command = new Command()
+  .description('Run test stories against the OpenSearch spec.')
+  .addOption(new Option('--spec_path <path>', 'path to the root folder of the multi-file spec').default('./spec'))
```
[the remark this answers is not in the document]
nhtruong: `--ignore-passed` meaning ignore printing PASSED chapters . You can turn it on so you can see all chapters being printed out when one of the chapters failed instead of just the failed one.

tools/src/tester/helpers.ts
```diff
@@ -0,0 +1,9 @@
+import { type Evaluation, Result } from './types/eval.types'
+
+export function overall_result (evaluations: Evaluation[]): Result {
+  if (evaluations.length === 0) return Result.PASSED
+  if (evaluations.some(e => e.result === Result.ERROR)) return Result.ERROR
+  if (evaluations.some(e => e.result === Result.FAILED)) return Result.FAILED
+  if (evaluations.every(e => e.result === Result.SKIPPED)) return Result.SKIPPED
+  return Result.PASSED
```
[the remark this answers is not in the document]
nhtruong: Those 2 code blocks are not logically equivalent though. In the former, a `Result.ERROR` anywhere in the array will result in `Result.ERROR` being returned. In the latter, if a FAILED appears before an ERROR, FAILED will be returned instead. We can definite incorporate this logic into that switch statement with a flag but it will make it even more complicated. I think trading a bit of performance for a better readability is fair here.

tools/src/tester/TestsRunner.ts
dblock: You could use a singleton.

tests/index_lifecycle.yaml
dblock: It's pretty unusual for a test framework. Generally you don't expect setup/teardown to fail with an unexpected error. You'll want to find a way where deleting an index succeeds on a 404 (index doesn't exist).

tools/src/tester/helpers.ts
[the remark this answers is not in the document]
Xtansia: That's a fair point regarding the difference in behaviour regarding errors vs failed. You could split the difference between lean more towards your preferred style but still reduce the amount of looping by counting upfront:
```
  const counts = evaluations.map(e => e.result)
    .reduce<Record<string, number>>((c, r) => ({ ...c, [r]: (c[r] ?? 0) + 1 }), { })

  if (counts[Result.ERROR] > 0) return Result.ERROR
  if (counts[Result.FAILED] > 0) return Result.FAILED
  if (counts[Result.SKIPPED] === evaluations.length) return Result.SKIPPED
  return Result.PASSED
```

tests/index_lifecycle.yaml
dblock: I like this but don't like the name. I think rather than ignore, 404 means success in this case. So this is really something like 

```yaml
response:
  status:
    - 200
    - 404
```

nhtruong: Just implemented `additional_success_statuses`
```
epilogues:
  - path: /books
    method: DELETE
    additional_success_statuses: [404]
```

json_schemas/test_story.schema.yaml
```diff
@@ -55,7 +55,7 @@ definitions:
       - $ref: '#/definitions/ChapterRequest'
       - type: object
         properties:
-          additional_success_status:
+          additional_success_statuses:
```
dblock, line 58: I would rename this to `status` and override the default value of `200` or `201` so that there's no hidden functionality.

tools/src/tester/start.ts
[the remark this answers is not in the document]
nhtruong: Went Redid the hide/show logic and apply 1 single rule across the board: If a component is either FAILED or ERROR, all of its sub-components will be displayed. The `--ignore` option has been removed.

tests/index_lifecycle.yaml
dblock: This hides that 200 status and doesn't allow you to remove 200. This would make it clearer:

```
epilogues:
  - path: /books
    method: DELETE
    status:
      - 200
      - 404
```

tools/src/tester/start.ts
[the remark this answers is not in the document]
nhtruong: image

tests/index_lifecycle.yaml
nhtruong: Will handle the missing field definition in a followup PR

nhtruong:
This is ready for the view and merged so people can start adding tests. I'll handle the remaining work in followup PRs:
- [ ] Update repo docs
- [ ] Add tests fro the framework (heavy lift)
- [ ] Test should fail when a field is found in the response but not defined in the spec

dblock (approved):
This LGTM! @Xtansia any objections to merging as is?

Xtansia (commented):
There's still some parts I think could be improved, but not against merging as is to iterate on it

tools/src/tester/start.ts
```diff
@@ -0,0 +1,23 @@
...
+import { LogLevel } from '../Logger'
+import TestsRunner from './TestsRunner'
+import { Command, Option } from '@commander-js/extra-typings'
+import _ from 'lodash'
+
+const command = new Command()
+  .description('Run test stories against the OpenSearch spec.')
+  .addOption(new Option('--spec, --spec_path <path>', 'path to the root folder of the multi-file spec').default('./spec'))
+  .addOption(new Option('--tests, --tests_path <path>', 'path to the root folder of the tests').default('./tests'))
+  .addOption(new Option('--tab_size <size>', 'tab size for displayed results').default('4'))
```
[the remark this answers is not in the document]
nhtruong: THanks for the callout. Will update them in the next PR

tools/src/tester/helpers.ts
```diff
@@ -0,0 +1,8 @@
+import { type Evaluation, Result } from './types/eval.types'
+
+export function overall_result (evaluations: Evaluation[]): Result {
+  if (evaluations.some(e => e.result === Result.ERROR)) return Result.ERROR
+  if (evaluations.some(e => e.result === Result.FAILED)) return Result.FAILED
+  if (evaluations.some(e => e.result === Result.SKIPPED)) return Result.SKIPPED
```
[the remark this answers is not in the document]
nhtruong: This is for evaluating components results. A component is skipped when a previous component results in an error (say if a chapter could not be executed for validation, all subsequent chapters will be skipped). However that would result in `ERROR` for overall. So for now line 6 is actually redundant. 

Hypothetically when we allow the user to skip some component evaluation (say skip evaluating the response body of a new endpoint still being developed), I think we should still mark the parent component as skipped. The print out will show all subcomponents so you can tell which subcomponent causes the `SKIPPED` status for the parent component. So:
- If you see ERROR as the result for a component, at least one of its subcomponent has ERROR
- otherwise, If you see FAILED as  the result for a component, at least one of its subcomponent has FAILED
- otherwise, If you see SKIPPED as the result for a component, at least one of its subcomponent has SKIPPED
- else, all components have been vested and passed the test.

tools/src/tester/start.ts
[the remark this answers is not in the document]
dblock: https://github.com/opensearch-project/opensearch-api-specification/pull/301 with tests

PR #301, opened 2024-05-30
Renaming parameters and adding command line tests.

dblock:
### Description

1. Renames `parameters_like_this` to `parameters-like-this` (from @Xtansia's comment in #299) .
2. Renamed tab size to tab width.
3. Adds tests for some of the command line parameters (the others are harder).
 
### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -48,6 +48,7 @@ export default [
           allowRuleToRunWithoutStrictNullChecksIKnowWhatIAmDoing: false
         }
       ],
+      '@typescript-eslint/no-extraneous-class': 'off',
       'array-callback-return': 'off',
       'new-cap': 'off',
       'no-return-assign': 'error',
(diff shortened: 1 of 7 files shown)


eslint.config.mjs
```diff
@@ -48,6 +48,7 @@ export default [
           allowRuleToRunWithoutStrictNullChecksIKnowWhatIAmDoing: false
         }
       ],
+      '@typescript-eslint/no-extraneous-class': 'off',
```
dblock, line 51: Not sure there's a better way than a `static Ansi` class?

nhtruong: You can also move those functions into the `tools/testers/helpers.ts` file or make an `ansi.ts` file in have them imported into ResultDisplays.ts

dblock: Extracted to `ansi`.

nhtruong (approved):

Xtansia (approved):

PR #304, opened 2024-05-31
Add text embedding processor

miguel-vila:
Signed-off-by: miguel-vila <miguelvilag@gmail.com>

### Description
Adds a definition for the text embedding processor: https://opensearch.org/docs/latest/ingest-pipelines/processors/text-embedding/

Please confirm whether the `field_map` definition makes sense

### DIFF
--- a/tests/text_embedding_processor.yaml
+++ b/tests/text_embedding_processor.yaml
@@ -1,0 +1,32 @@
+$schema: ../json_schemas/test_story.schema.yaml
+
+skip: false
+description: This story tests that you can create 
+epilogues:
+  - path: /_ingest/pipeline/books_pipeline
+    method: DELETE
+    status: [200, 404]
+chapters:
+  - synopsis: Create ingest pipeline for text embedding
+    path: /_ingest/pipeline/{id}
+    method: PUT
+    parameters:
+      id: books_pipeline
+    request_body:
+      payload:
+        description: "Extracts text from field and embeds it"
+        processors:
+          - text_embedding:
+              model_id: "text-embedding-model"
+              field_map:
+                text: "passage_embedding"
+    response:
+      status: 200
+  - synopsis: Query created pipeline
+    path: /_ingest/pipeline/{id}
+    method: GET
+    parameters:
+      id: books_pipeline
+    response:
+      status: 200
+      
--- a/tests/text_embedding_processor.yaml
+++ b/tests/text_embedding_processor.yaml
@@ -1,7 +1,9 @@
 $schema: ../json_schemas/test_story.schema.yaml
 
 skip: false
-description: This story tests that you can create 
+description: |
+  This test story checks that we can create an ingest pipeline with a text 
+  embedding processor
 epilogues:
   - path: /_ingest/pipeline/books_pipeline
     method: DELETE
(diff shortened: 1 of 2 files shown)


dblock:
Thank you! 

We just merged a test framework in #299. Want to try to add a test for this API? Check out https://github.com/opensearch-project/opensearch-api-specification/blob/main/tests/index_lifecycle.yaml for an example. You can run it with `npm run test:spec` against a live local instance of OpenSearch (documentation incoming).

dblock (dismissed):

dblock:
@nhtruong @Xtansia spec looks good to you?

miguel-vila:
@dblock  I have added a test at [ee776b4](https://github.com/opensearch-project/opensearch-api-specification/pull/304/commits/ee776b41f25508bc9157c70e4777cb7a9482c6cd) but it fails locally because the response doesn't seem to include the `_meta` field. Might look into it later but not sure if it's because there's some real misalignment between the spec and the implementation.

dblock:
> _meta

So likely a bug! That's why we need tests ;)

nhtruong:
@dblock @miguel-vila  Small discrepancies between the spec and the actual implementation of OS are to be expected. Much of spec of the core features that we have right now are inherited from ElasticSearch. We will need help from OS core team to review the spec of the core features. For now, @miguel-vila, you can remove the `_meta` as a required property.

tests/text_embedding_processor.yaml
```diff
@@ -0,0 +1,28 @@
+$schema: ../json_schemas/test_story.schema.yaml
+
+skip: false
+description: |
+  This test story checks that we can create an ingest pipeline with a text 
+  embedding processor
```
dblock, line 6: Let's sweat some small stuff since this is new. Add a period. Maybe shorten, "Create and use an ingest pipeline with a text embedding processor."

tests/text_embedding_processor.yaml
```diff
@@ -0,0 +1,28 @@
...
+epilogues:
+  - path: /_ingest/pipeline/books_pipeline
+    method: DELETE
+    status: [200, 404]
+chapters:
+  - synopsis: Create ingest pipeline for text embedding
```
dblock, line 12: Create an ingest pipeline ... + add a period.

dblock (commented):
I think we're going to have many ingest pipeline tests. How about we organize things in folders that match the schema?

So this test should probably go into something like tests/ingest/ingest_with_text_embedding_processor.yaml?

It would be great if the story included a GET of the pipeline, maybe even used the processor.

nhtruong:
I think this can be merge now. We address those minor wording issues later in the test.

nhtruong (approved):

miguel-vila:
I could address some of the bigger changes in a separate PR.

In particular, I wanted to test a whole flow of creating a model and using search against it (I think the search was failing due to the model not existing, which makes sense because I set some random `"text-embedding-model"` as model id) but then I noticed your test framework would need some extensions:

- allow remembering some ids (e.g. the model id) and using them in later requests
- allow retrying some requests (we need to wait for the model to be deployed)

I have some work in progress in that direction but these might be big/controversial changes so would be good to know your thoughts, might push my WIP PR later.

PR #310, opened 2024-06-04
Added Tests for the test framework

nhtruong:

### DIFF
--- a/DEVELOPER_GUIDE.md
+++ b/DEVELOPER_GUIDE.md
@@ -8,9 +8,11 @@
     * [Superseded Operations](#superseded-operations)
     * [Global Parameters](#global-parameters)
     * [OpenAPI Extensions](#openapi-extensions)
+  * [Writing Spec Tests](#writing-spec-tests)
+    * [Running Spec Tests Locally](#running-spec-tests-locally)
   * [Tools](#tools)
     * [Setup](#setup)
-    * [Merger](#merger)
+    * [Spec Merger](#spec-merger)
       * [Arguments](#arguments)
       * [Example](#example)
     * [Spec Linter](#spec-linter)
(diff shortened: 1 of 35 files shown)


DEVELOPER_GUIDE.md
```diff
@@ -145,6 +145,64 @@ This repository includes several OpenAPI Specification Extensions to fill in any
 - `x-global`: Denotes that the parameter is a global parameter that is included in every operation. These parameters are listed in the [spec/_global_parameters.yaml](spec/_global_parameters.yaml).
 - `x-default`: Contains the default value of a parameter. This is often used to override the default value specified in the schema, or to avoid accidentally changing the default value when updating a shared schema.
 
+## Adding tests for the spec
```
dblock, line 148: Capitalize for consistency, maybe just "Writing Spec Tests"? Add to TOC.

DEVELOPER_GUIDE.md
```diff
@@ -145,6 +145,64 @@ This repository includes several OpenAPI Specification Extensions to fill in any
...
+
+To assure the correctness of the spec, you must add tests for the spec in the [tests/](tests) directory. Each yaml file in the tests directory represents a test story that tests a collection of related operations. A test story has 3 main components:
+- prologues: These are the operations that are executed before the test story is run. They are used to set up the environment for the test story.
+- chapters: These are the operations that are being tested.
+- epilogues: These are the operations that are executed after the test story is run. They are used to clean up the environment after the test story.
+
+Below is an example test story that tests the index operations:
```
dblock, line 155: Maybe link to the index lifecycle story for the latest version?

dblock (approved):
Looks good. Fix CI, + some minor things.

I do recommend putting back the empty.yaml as a a minimal test of passing something that has no actual tests in it.

tools/tests/tester/helpers.ts
```diff
@@ -0,0 +1,46 @@
...
+import { type OpenAPIV3 } from 'openapi-types'
+import YAML from 'yaml'
+import type { StoryEvaluation } from '../../src/tester/types/eval.types'
+import type { Story } from '../../src/tester/types/story.types'
+import { read_yaml } from '../../helpers'
+import StoryEvaluator from '../../src/tester/StoryEvaluator'
+
+export function create_shared_resources (spec: any): void {
+  // The password must match the one specified in .github/workflows/test-spec.yml
+  process.env.OPENSEARCH_PASSWORD = 'myStrongPassword123!'
```
Xtansia, line 14: Why are we hard coding a password here, this should just be set in the workflow

tools/tests/tester/TestsRunner.test.ts
```diff
@@ -0,0 +1,27 @@
+import { read_yaml } from '../../helpers'
+import TestsRunner from '../../src/tester/TestsRunner'
+import { type OpenAPIV3 } from 'openapi-types'
+import { load_expected_evaluation, scrub_errors } from './helpers'
+
+test('stories folder', async () => {
+  // The password must match the one specified in .github/workflows/test-spec.yml
+  process.env.OPENSEARCH_PASSWORD = 'myStrongPassword123!'
```
Xtansia, line 8: Why are we hard coding a password here, this should just be set in the workflow

Xtansia (changes requested):

tools/tests/tester/TestsRunner.test.ts
nhtruong: I should change it to `process.env.OPENSEARCH_PASSWORD =  process.env.OPENSEARCH_PASSWORD ?? 'myStrongPassword123!'`. The idea is streamlining local testing for the user. When they run `docker compose up`, this is the same password the cluster will be setup with. So, they can quickly run the tests right after the contain is set up and ready.

nhtruong: It's not a security concern since this is just for testing and the password is already shown in the workflow.

DEVELOPER_GUIDE.md
nhtruong: I'll add it as a reference but this one has the comments explaining what each property means too.

nhtruong:
Figuring out why the tests fail here but not locally. Changing this to draft. Please ignore the commit pushes till I change it back to reviewable.

nhtruong:
@dblock We currently does not allow an empty story because the `chapters` array [must have at least one element](https://github.com/opensearch-project/opensearch-api-specification/blob/main/json_schemas/test_story.schema.yaml#L25) in it. The deleted `empty.yaml` story is not a valid story.

Xtansia (approved):

Issue #266, opened 2024-04-23
[FEATURE] Add text quote linter.


dblock:
### Is your feature request related to a problem?

Coming from https://github.com/opensearch-project/opensearch-api-specification/pull/264 we removed quotes around YAML text values which are unnecessary.

### What solution would you like?

A linter that fails CI when unnecessary/incorrect (single/double) quotes are used.

PR #312, opened 2024-06-04
Adds a YAML eslint linter.

dblock:
### Description

Lints YAML using a formatting linter. Fixes things like spaces, preferring single/double quotes, etc.

We have a [spec linter](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md#spec-linter) that is about the spec itself in which we could implement some of these rules too, but feels easier. That one probably could have been (still can be) a generic eslint linter, too. 

### Issues Resolved

Closes https://github.com/opensearch-project/opensearch-api-specification/issues/266.

### DIFF
--- a/.github/opensearch-cluster/docker-compose.yml
+++ b/.github/opensearch-cluster/docker-compose.yml
@@ -7,5 +7,5 @@ services:
       - "9200:9200"
       - "9600:9600"
     environment:
-      - "discovery.type=single-node"
+      - discovery.type=single-node
       - "OPENSEARCH_INITIAL_ADMIN_PASSWORD=${OPENSEARCH_PASSWORD:-myStrongPassword123!}"
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -1,23 +1,32 @@
-import path from 'path'
-import { fileURLToPath } from 'url'
-import { FlatCompat } from '@eslint/eslintrc'
 import pluginJs from '@eslint/js'
+import pluginTs from '@typescript-eslint/eslint-plugin'
+import parserTs from '@typescript-eslint/parser'
+import eslintPluginYml from 'eslint-plugin-yml'
+import parserYml from "yaml-eslint-parser"
+import globals from 'globals'
 import licenseHeader from 'eslint-plugin-license-header'
-
-// mimic CommonJS variables -- not needed if using CommonJS
-const _filename = fileURLToPath(import.meta.url)
-const _dirname = path.dirname(_filename)
-const compat = new FlatCompat({ baseDirectory: _dirname, recommendedConfig: pluginJs.configs.recommended })
 
 export default [
   pluginJs.configs.recommended,
-  ...compat.extends('standard-with-typescript'),
   {
     files: ['**/*.{js,ts}'],
+    languageOptions: {
+      parser: parserTs,
+      parserOptions: {
+        project: './tsconfig.json'
+      },
+      globals: {
+        ...globals.jest,
+        ...globals.node,
+      },
+    },
     plugins: {
+      '@typescript-eslint': pluginTs,
       'license-header': licenseHeader
     },
     rules: {
+      ...pluginJs.configs.recommended.rules,
+      ...pluginTs.configs["recommended-type-checked"].rules,
       '@typescript-eslint/consistent-indexed-object-style': 'error',
       '@typescript-eslint/consistent-type-assertions': 'error',
       '@typescript-eslint/dot-notation': 'error',
@@ -71,6 +80,20 @@ export default [
           '*/'
         ]
       ]
+    }
+  },
+  ...eslintPluginYml.configs['flat/standard'],
+  {
+    files: ["**/*.yaml", "**/*.yml"],
+    languageOptions: {
+      parser: parserYml
+    },
+    plugins: {
+      yml: eslintPluginYml
+    },
+    rules: {
+      'yml/no-empty-document': 'off',
+      'yml/quotes': ['error', { prefer: 'single' }]
     }
   }
 ]
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -27,6 +27,11 @@ export default [
     rules: {
       ...pluginJs.configs.recommended.rules,
       ...pluginTs.configs["recommended-type-checked"].rules,
+      '@typescript-eslint/no-unsafe-member-access': 'off',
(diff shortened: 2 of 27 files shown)


tools/src/linter/components/base/FileValidator.ts
```diff
@@ -22,10 +22,10 @@ export default class FileValidator extends ValidatorBase {
   }
 
   validate (): ValidationError[] {
-    const extension_error = this.validate_extension()
-    if (extension_error) return [extension_error]
     const yaml_error = this.validate_yaml()
     if (yaml_error) return [yaml_error]
+    const extension_error = this.validate_extension()
+    if (extension_error) return [extension_error]
```
nhtruong, line 28: Is there a reason for this reordering? a `txt` file should result in a failure before we try to parse its content as YAML right?

.github/opensearch-cluster/docker-compose.yml
```diff
@@ -4,8 +4,8 @@ services:
...
-      - "9200:9200"
-      - "9600:9600"
+      - '9200:9200'
+      - '9600:9600'
     environment:
-      - "discovery.type=single-node"
-      - "OPENSEARCH_INITIAL_ADMIN_PASSWORD=${OPENSEARCH_PASSWORD}"
\ No newline at end of file
+      - discovery.type=single-node
+      - 'OPENSEARCH_INITIAL_ADMIN_PASSWORD=${OPENSEARCH_PASSWORD}'
```
nhtruong, line 11: Personally I prefer being able to add the quotes on line even though it's not necessary for that specific case. It just makes it easier to edit and you don't have to keep removing and adding the quotes as the value changes. It's the same reason some languages, like JS itself, allows trailing comma like the comma after "world" in:
```javascript
{ 
   foo: "bar",
   hello: "world",
}
``` 
so that when you add or a remove a property after `hello: "world"` you dont have to keep removing or adding the comma. There's no harm in keeping the quotes either. So I think this lint is not necessary.

tools/tests/linter/fixtures/folder_validators/namespaces/invalid_folder/dup_path_b.yaml
```diff
@@ -1,3 +1,3 @@
 paths:
-  '/_rollover': {}
+  /_rollover: {}
   '/{index}/_rollover': {}
```
nhtruong, line 3: This lint actually makes things look inconsistent and doesn't add any value.

tools/tests/merger/fixtures/spec/namespaces/shelter.yaml
```diff
@@ -24,22 +24,22 @@ paths:
...
+      in: path
+      schema:
+        type: number
+  responses:
+    adopt@200:
+      description: ''
+      application/json:
         schema:
-          $ref: '../schemas/animals.yaml#/components/schemas/Animal'
-      adopt::path.docket:
```
nhtruong, line 35: We discussed in this [PR](https://github.com/opensearch-project/opensearch-api-specification/pull/271/files) about allowing (or even enforcing) an empty line between properties to help discern large and complex properties. Is this still the case?

eslint.config.mjs
```diff
@@ -51,7 +68,22 @@ export default [
       'array-callback-return': 'off',
       'new-cap': 'off',
       'no-return-assign': 'error',
-      'object-shorthand': 'error'
+      'object-shorthand': 'error',
+      'no-constant-condition': 'off'
```
nhtruong, line 72: We're turning off a lot of TS lints in this change.

tools/src/linter/components/base/FileValidator.ts
dblock: Yes, the yaml file will fail to parse with eslint, and when you ignore it in the configuration it will still try to load it and fail, so I needed to rename it. 

But after renaming it the linter test cannot test failing yaml parsing anymore because it trips on the extension. This seems like the easiest solution.

.github/opensearch-cluster/docker-compose.yml
dblock: The lint makes it possible to not have to think about it, and not get inconsistent format for VARIABLES, where some are one way and others are the other way.

I’ll remove though, but lmk if this changes your mind.

eslint.config.mjs
dblock: Actually we're not, these are added via `recommended-type-checked` and I believe we didn't have them enabled before. 

```
$ git diff
diff --git a/eslint.config.mjs b/eslint.config.mjs
index 53e6a18..acb7a4b 100644
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -52,6 +52,13 @@ export default [
         }
       ],
+      '@typescript-eslint/no-explicit-any': 'error',
+      '@typescript-eslint/no-unsafe-assignment': 'error',
+      '@typescript-eslint/no-unsafe-return': 'error',
+      '@typescript-eslint/no-unsafe-call': 'error',
+      '@typescript-eslint/no-unsafe-member-access': 'error',
+      '@typescript-eslint/no-unused-vars': ['error', { "argsIgnorePattern": "^_" }],
+      '@typescript-eslint/require-await': 'error',
       'array-callback-return': 'off',
       'new-cap': 'off',
       'no-return-assign': 'error',
```

```
$ npm run lint

✖ 193 problems (193 errors, 0 warnings)
  3 errors and 0 warnings potentially fixable with the `--fix` option.
```

dblock: Two of these are easy to enable. The others give hundreds of errors.

- [opensearch-api-specification#318](https://github.com/opensearch-project/opensearch-api-specification/pull/318): Turn on @typescript-eslint/no-unused-var. 
- [opensearch-api-specification#319](https://github.com/opensearch-project/opensearch-api-specification/pull/319): Turn on @typescript-eslint/require-await.

tools/tests/merger/fixtures/spec/namespaces/shelter.yaml
dblock: Undid.

dblock:
This is ready. I only auto-fixed `yml/indent` and turned off single/double quote rules that remove quotes from keys.

nhtruong (dismissed):
Looks good to me. Need a rebase.

dblock:
> Looks good to me. Need a rebase.

Rebased.

nhtruong (approved):

PR #315, opened 2024-06-05
Full flow to test neural search

miguel-vila:
Signed-off-by: miguel-vila <miguelvilag@gmail.com>

The intention would be to test a more complex flow.
This would require some changes to the testing framework:

- support keeping info from responses and using it for later requests (e.g. the model id, which is then needed to create the ingestion
- support some sort of retry for the task API (we need to wait for the model deployment before using it)

To test some of this I also had to add a new `ml` API spec, it's still a bit rough and might use some better descriptions.


tests/ingest/ingest_with_text_embedding_processor.yaml
```diff
@@ -0,0 +1,108 @@
...
+    status: [200, 404]
+  # - path: /_plugins/_ml/model_groups/{model_group_id}
+  #   method: DELETE
+  #   status: [200, 404]
+  #   parameters_from_environment:
+  #     model_group_id: test_model_group_id
+  - path: /_plugins/_ml/models/{model_id}
+    method: DELETE
+    status: [200, 404]
+    parameters_from_environment:
```
dblock, line 22: We should support a syntax that resolves from multiple sources:

```yaml
parameters:
   model_id: "${env.test_model_id}"
```

Note though that if the model could not be created this cleanup doesn't work. Is there a way to delete all models or specify the model ID in the request when it gets created?

tests/ingest/ingest_with_text_embedding_processor.yaml
```diff
@@ -0,0 +1,108 @@
...
+            ml_commons:
+              only_run_on_ml_node: false
+  # - synopsis: Create model group
+  #   path: /_plugins/_ml/model_groups/_register
+  #   method: POST
+  #   request_body:
+  #     payload:
+  #       name: "NLP_Group"
+  #       description: "Model group for NLP models"
+  #   remember:
```
dblock, line 41: I like the way GitHub actions does this. How about an `output` that can be referenced?

```yaml
id: create_model_chapter
output:
   test_model_group_id: response_body.model_group_id
```

Then it could be referenced as `create_model_chapter.test_model_group_id`?

tests/ingest/ingest_with_text_embedding_processor.yaml
```diff
@@ -0,0 +1,108 @@
...
+        model_format: TORCH_SCRIPT
+        # @TODO: Add model group, need to support passing things from the environment
+        # to the request body
+    remember:
+      payload:
+        model_registration_task_id: "task_id"
+  # - synopsis: Wait for model registration task to complete and keep the model id
+  #   path: /_plugins/_ml/tasks/{task_id}
+  #   method: GET
+  #   retry_on_response:
```
dblock, line 60: I think we want to retry on any failure a number of times.

```
retry: 2
```

But I understand that in this case we are waiting for a field. Today we check for response schema. In addition I think we can add a `response_body` that expects some values, then `retry: 2` makes sense.

```
retry: 2
response_body:
    status: "CREATED"
```

dblock (commented):
This is awesome. I think we should sweat the DSL a bit. See my comments below. I would PR the features like `retry`, `response_body` and collecting output separately.

tests/ingest/ingest_with_text_embedding_processor.yaml
dblock: @miguel-vila I opened https://github.com/opensearch-project/opensearch-api-specification/issues/328 so we don't lose this. Comment if you're working on it!

dblock:
I added support for retry in https://github.com/opensearch-project/opensearch-api-specification/pull/453 with the waiting for tasks example from here. 

@miguel-vila any interest in finishing this on top of my PR?

dblock:
Finished in https://github.com/opensearch-project/opensearch-api-specification/pull/504.

PR #302, opened 2024-05-31
Refactor spec tester to improve reusability of classes

Xtansia:
### Description
Refactors sections of the spec tester to allow better reusability of the classes and remove the `SharedResources` pattern. Additionally pulls out a common set of parameters for connecting to OpenSearch.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #316, opened 2024-06-05
Remove shared resources.

dblock:
### Description

The use of shared resource is an anti-pattern that ties together unrelated concepts.  A simple example is of `ChapterReader` that is only needed by `ChapterEvaluator`. By introducing a shared resource we start treating the spec, reader and evaluator as singletons in a way that combines them in 1 place, creating unneeded interdependencies. The one thing we do need to pass around is the actual spec, which makes sense as all these classes use the spec in some way. I think the resulting code is a lot clearer (and much shorter, especially in tests where we don't need to pre-create anything).

I also cleaned up some test befores/after's.


nhtruong:
I believe @Xtansia's https://github.com/opensearch-project/opensearch-api-specification/pull/302 is already handling this.

dblock:
> I believe @Xtansia's #302 is already handling this.

Yes, but mine is much simpler because it removes unnecessary `this._stuff`. I can do it on top of @Xtansia's PR too.

nhtruong:
Got it. Although, I'm afraid you guys are probably stepping on each other's toe haha.

dblock:
> Got it. Although, I'm afraid you guys are probably stepping on each other's toe haha.

Yes, I should have looked. It bugged me so I went ahead and made the change. I don't mind doing the work twice or throwing it out :)

dblock:
https://github.com/opensearch-project/opensearch-api-specification/pull/302 addressed this, closing

Issue #313, opened 2024-06-04
[BUG] npm run test requires a docker instance of OpenSearch


dblock:
### What is the bug?

Tests require a local OpenSearch running on 9200. This is unexpected for unit tests.

### How can one reproduce the bug?

```
$ npm run test -- tools/tests/tester/TestsRunner.test.ts 

> opensearch_api_tools@1.0.0 test
> jest tools/tests/tester/TestsRunner.test.ts

 FAIL  tools/tests/tester/TestsRunner.test.ts
  ✕ stories folder (134 ms)

  ● stories folder

    connect ECONNREFUSED ::1:9200

      26 |     const response: Record<string, any> = {}
      27 |     const [url, params] = this.#parse_url(chapter.path, chapter.parameters ?? {})
    > 28 |     await axios.request({
         |     ^
      29 |       url,
      30 |       auth: {
      31 |         username: 'admin',

      at Function.Object.<anonymous>.AxiosError.from (node_modules/axios/lib/core/AxiosError.js:89:14)
      at RedirectableRequest.handleRequestError (node_modules/axios/lib/adapters/http.js:610:25)
      at ClientRequest.eventHandlers.<computed> (node_modules/follow-redirects/index.js:38:24)
      at Axios.request (node_modules/axios/lib/core/Axios.js:45:41)
      at ChapterReader.read (tools/src/tester/ChapterReader.ts:28:5)
      at ChapterEvaluator.evaluate (tools/src/tester/ChapterEvaluator.ts:35:22)
      at StoryEvaluator.#evaluate_chapters (tools/src/tester/StoryEvaluator.ts:71:28)
      at StoryEvaluator.evaluate (tools/src/tester/StoryEvaluator.ts:50:22)
      at TestsRunner.run (tools/src/tester/TestsRunner.ts:45:26)
      at Object.<anonymous> (tools/tests/tester/TestsRunner.test.ts:20:30)

    Cause:
    connect ECONNREFUSED ::1:9200



Test Suites: 1 failed, 1 total
Tests:       1 failed, 1 total
Snapshots:   0 total
Time:        1.42 s, estimated 2 s
```

### What is the expected behavior?

Tests run by default should not require a server.

### Do you have any additional context?

We can move some tests to integration tests, use mocks, or the new `--dry-run` option added in #303. 

Running all tests fails with unhelpful errors.

```
node:internal/child_process/serialization:159
    const string = JSONStringify(message) + '\n';
                   ^

[...]

PR #317, opened 2024-06-05
Make TestsRunner.test.ts and TestsRunner.test.ts independent.

dblock:
### Description

If you run `npm run test` the two test suites will conflict as they both create/delete the same indices (books). Split the fixtures, called one books and the other shoes.

Use `beforeAll` to avoid global setup.

Renamed `tools/tests/tester/overall_result.test.ts` to `helpers.test.ts` to match the file being tested.

Removes the need for `--runInBand`, runs faster and with defaults.

The only downside is duplication of the YAML tests.

### Issues Resolved

Part of https://github.com/opensearch-project/opensearch-api-specification/issues/313.


nhtruong:
Did you create `books` and `shoes` folders to circumvent the need for `--runInBand` where each set creates and deletes a different index?  While that will solve this particular set of integ tests, as we add more integ tests in the future, the tests can interfere with each other unpredictably and result in flaky tests. This add extra complexity for test writers as we will have to be assure that all resources created in all tests are mutually exclusive.

dblock:
> Did you create `books` and `shoes` folders to circumvent the need for `--runInBand` where each set creates and deletes a different index? While that will solve this particular set of integ tests, as we add more integ tests in the future, the tests can interfere with each other unpredictably and result in flaky tests. This adds extra complexity for test writers as we will have to be sure that all resources created in all tests are mutually exclusive.

Yes. I come from the point of view that tests should be able to run in parallel and not be interfering with each-other. But I understand why integration tests may be different. Would you prefer a solution where we make these tests "integration" and run them with `--runInBand`?

nhtruong:
Yes. I think running integration tests with `--runInBand` is the way to go. We can divide the `tools/tests` folder into:
- `tests/integ/tester`
- `tests/unit/tester`
- `tests/unit/merger`
- `tests/unit/linter`
and have 2 `npm run commands`
- test:tools:unit --> jest tools/tests/unit
- test:tools:integ --> jest tools/tests/integ --runInBand

dblock:
Closing in favor of https://github.com/opensearch-project/opensearch-api-specification/pull/320.

PR #324, opened 2024-06-06
Support outputs on tests

miguel-vila:
Adds a way to describe outputs in chapters, that can be used for later requests in the params or in the request bodies.

Added some unit tests and an integration test that uses this feature.


tools/src/tester/helpers.ts
```diff
@@ -1,8 +1,152 @@
...
 
-export function overall_result (evaluations: Evaluation[]): Result {
+export function overall_result(evaluations: Evaluation[]): Result {
   if (evaluations.some(e => e.result === Result.ERROR)) return Result.ERROR
   if (evaluations.some(e => e.result === Result.FAILED)) return Result.FAILED
   if (evaluations.some(e => e.result === Result.SKIPPED)) return Result.SKIPPED
   return Result.PASSED
 }
+
+export function extract_output_values(response: ActualResponse, chapterOutput?: Output): EvaluationWithOutput {
```
dblock, line 12: Refactor into `ChapterOutput`, I think it should be constructed from `ActualResponse`, be a method on `ActualResponse`, or something like that.

dblock (commented):
This looks good! Rebase, add some to the documentation (DEVELOPER_GUIDE on writing tests probably). 

I suggest wrapping up the chapter output into a class and refactoring all the helper methods as class methods, it will make things a lot cleaner/better organized.

tools/src/tester/types/story.types.ts
```diff
@@ -1,5 +1,7 @@
 /* eslint-disable */
 
+import { ChapterOutput } from "./eval.types";
+
```
nhtruong, line 4: This file is not supposed to be edited by hand. It's generated from running `npx ts-node tools/src/tester/_generate_story_types.ts`.

The goal is maintaining parity between the JSON Schemas defined in `test_story.schema.yaml` and the TS types defined here.

tools/src/tester/StoryEvaluator.ts
```diff
@@ -8,11 +8,11 @@
 */
 
 import { type Chapter, type Story, type SupplementalChapter } from './types/story.types'
-import { type ChapterEvaluation, Result, type StoryEvaluation } from './types/eval.types'
+import { type ChapterEvaluation, Result, type StoryEvaluation, ChaptersEvaluations, StoryOutputs, Evaluation } from './types/eval.types'
 import ChapterEvaluator from './ChapterEvaluator'
 import type ChapterReader from './ChapterReader'
 import SharedResources from './SharedResources'
-import { overall_result } from './helpers'
+import { check_story_variables, extract_output_values, overall_result } from './helpers'
```
nhtruong, line 15: If `check_story_variables` is only used here, it should be defined here instead.

tools/src/tester/helpers.ts
```diff
@@ -7,11 +7,156 @@
...
+      return `Chapter makes reference to non existent output "${output_name}" in chapter "${chapter_id}"`
+    }
+  }
+}
+
+export function check_story_variables(story: Story): string | undefined {
+  const story_outputs: StoryOutputs = {}
+  const all_episodes = (story.prologues || []).concat(story.chapters || []).concat(story.epilogues || [])
+  for(const episode of all_episodes) {
+    const error = check_used_variables(episode, story_outputs)
```
nhtruong, line 151: Isn't `story_ouputs` always `{}` in this case?

nhtruong (changes requested):

tools/src/tester/StoryEvaluator.ts
```diff
@@ -46,9 +46,21 @@ export default class StoryEvaluator {
...
+    if (variables_error) {
+      return {
+        result: Result.ERROR,
+        display_path: this.display_path,
+        full_path: this.full_path,
+        description: this.story.description,
+        chapters: [],
+        message: variables_error
+      }
+    }
```
nhtruong, line 59: These errors should be assigned to the chapters that throw them so the view can pinpoint where to fix.

tools/src/tester/ResultsDisplayer.ts
```diff
@@ -41,7 +41,7 @@ export default class ResultsDisplayer {
 
   #display_story (): void {
     const result = this.evaluation.result
-    const message = this.evaluation.full_path
+    const message = this.evaluation.message || this.evaluation.full_path
```
nhtruong, line 44: With [this](https://github.com/opensearch-project/opensearch-api-specification/pull/324/files/b0340fe5bdfac2013e63fa77cbf854b91ad16e31#r1633900477) we wont need this change anymore.

tools/src/tester/ChapterOutput.ts
```diff
@@ -0,0 +1,36 @@
...
+  }
+
+  /**
+   * Creates a dummy ChapterOutput from an Output object
+   * where the values will be the response paths.
+   * Used for a validation check.
+   * @param output
+   * @returns
+   */
+  static create_dummy_from_output (output: Output): ChapterOutput {
```
dblock, line 33: Nit: this seems unnecessary, do `new ChapterOutput` where it's called?

tools/src/tester/SupplementalChapterEvaluator.ts
```diff
@@ -0,0 +1,58 @@
...
+
+    constructor(chapter_reader: ChapterReader) {
+        this._chapter_reader = chapter_reader;
+    }
+
+    async evaluate(chapter: SupplementalChapter, story_outputs: StoryOutputs): Promise<{ evaluation: ChapterEvaluation, evaluation_error: boolean }> {
+        const title = `${chapter.method} ${chapter.path}`
+        const response = await this._chapter_reader.read(chapter, story_outputs)
+        const status = chapter.status ?? [200, 201]
+        const output_values = extract_output_values(response)
```
dblock, line 27: This should read `response.output_values`, it's very business-logic specific to the reponse.

tools/src/tester/ChapterEvaluator.ts
```diff
@@ -10,40 +10,43 @@
...
     if (skip) return { title: chapter.synopsis, overall: { result: Result.SKIPPED } }
-    const response = await this._chapter_reader.read(chapter)
+    const response = await this._chapter_reader.read(chapter, story_outputs)
     const operation = this._operation_locator.locate_operation(chapter)
     if (operation == null) return { title: chapter.synopsis, overall: { result: Result.FAILED, message: `Operation "${chapter.method.toUpperCase()} ${chapter.path}" not found in the spec.` } }
     const params = this.#evaluate_parameters(chapter, operation)
     const request_body = this.#evaluate_request_body(chapter, operation)
     const status = this.#evaluate_status(chapter, response)
     const payload = status.result === Result.PASSED ? this.#evaluate_payload(response, operation) : { result: Result.SKIPPED }
+    const output_values = extract_output_values(response, chapter.output)
```
dblock, line 39: `extract_output_values` should either be a private method on `ChapterEvaluator` or maybe this works better as `chapter.output.extract_output_values_from_response(response)`?

dblock (changes requested):
@miguel-vila Let's get this merged? Iterate to green and say when you think you're done with the changes?

I don't love it that we're adding a ton of things to `helpers`, that's an OO anti-pattern. Consider refactoring those. Can also be refactired later.

tools/tests/tester/start.test.ts
```diff
@@ -78,15 +77,15 @@ test('extract_output_values', () => {
     d: 'payload.a.arr[0].d',
     e: 'payload.a.arr[1].e'
   }
-  expect(extract_output_values(response, output1)).toEqual(passed_output({
+  expect(ChapterOutput.extract_output_values(response, output1)).toEqual(passed_output({
```
dblock, line 80: A `ChapterOutput` is something that is the result of an `Output` (a schema) and a response (values). I think the OO way should be closer to this:

```
new ChapterOutput(output, values)
```

dblock (approved):
@nhtruong take a look, I am good without all the refactors, can be done later

nhtruong (approved):

PR #331, opened 2024-06-10
Reorganized tests and added tests for _doc.

dblock:
### Description

1. Reorganized tests to match spec.
2. Added tests for POST/PUT/DELETE `_doc`, adding the missing 201 response.


DEVELOPER_GUIDE.md
```diff
@@ -155,7 +155,7 @@ To assure the correctness of the spec, you must add tests for the spec in the [t
 - chapters: These are the operations that are being tested.
 - epilogues: These are the operations that are executed after the test story is run. They are used to clean up the environment after the test story.
 
-Below is the simplified version of the test story that tests the [index operations](tests/index.yaml):
```
nhtruong, line 158: the operations tested in the index lifecycle story is actually part of the `indices` namespace. I think we should put it in `tests/indices/lifecycle.yaml`

dblock: Makes sense to organize tests in namespaces, updated and added to DEVELOPER_GUIDE.

nhtruong (approved):

Issue #338, opened 2024-06-12
[FEATURE] Warn on missing fields


dblock:
### Is your feature request related to a problem?

Coming from https://github.com/opensearch-project/opensearch-api-specification/pull/336 we are adding a missing field to a response body.

### What solution would you like?

1. Spec testing should warn that the response contains a field that is not present in the spec.
2. It should be possible to run the tester with a flag that turns the warning in (1) into an error.

I believe this is not possible with AJV today, opened https://github.com/ajv-validator/ajv/issues/2453.

PR #336, opened 2024-06-12
Add `distribution` field to `OpenSearchVersionInfo`

Xtansia:
### Description
Adds missing `distribution` field to `OpenSearchVersionInfo`

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #342, opened 2024-06-17
Fail on missing properties.

dblock:
### Description

A large category of bugs in the schema is that some fields are missing. For example, the distribution filed in the response from GET / was added in https://github.com/opensearch-project/opensearch-api-specification/pull/336.

We should have detected the missing field in tests. Had the info schema had `additionalProperties: false` the schema validation would have failed with "data/version must NOT have additional properties". However adding that option means that our published schema cannot be future-proof for when new fields are added.

This PR dynamically injects `additionalProperties: ...` with `ajv-errors` that allows to custom render an error message that contains the property name that is missing. 

For now it only applies to any set of properties with a `required` field. Otherwise it creates some false positives on index responses with dynamic mappings.

With this change:

```
npm run test:spec -- --verbose --tests tests/_core/info.yaml --opensearch-insecure

FAILED  Test root endpoint. (/Users/dblock/source/opensearch-project/opensearch-api-specification/dblock-opensearch-api-specification/tests/_core/info.yaml)
    FAILED  CHAPTERS 
        FAILED  Get server info. 
            PASSED  REQUEST BODY 
            PASSED  RESPONSE STATUS 
            FAILED  RESPONSE PAYLOAD (data/version/distribution property is not defined in the spec)
```

You get all the errors:

```
FAILED  Test the creation of an ingest pipeline with a text embedding processor.
 (/Users/dblock/source/opensearch-project/opensearch-api-specification/dblock-opensearch-api-specification/tests/ingest/pipeline.yaml)
    FAILED  CHAPTERS 
        FAILED  Create ingest pipeline for text embedding. 
            PASSED  PARAMETERS 
                PASSED  id 
            FAILED  REQUEST BODY (data/processors/0/text_embedding/model_id property is not defined in the spec, data/processors/0/text_embedding/field_map property is not defined in the spec)
            PASSED  RESPONSE STATUS 
            PASSED  RESPONSE PAYLOAD 
        FAILED  Query created pipeline. 
            PASSED  PARAMETERS 
                PASSED  id 
            PASSED  REQUEST BODY 
            PASSED  RESPONSE STATUS 
            FAILED  RESPONSE PAYLOAD (data/books_pipeline/processors/0/text_embedding/model_id property is not defined in the spec, data/books_pipeline/processors/0/text_embedding/field_map property is not defined in the spec)
```

### Issues Resolved

Closes https://github.com/opensearch-project/opensearch-api-specification/issues/338.


dblock:
@nhtruong I'll write tests and document this, but would love your thoughts in the meantime please?

tools/src/tester/StricterMergedOpenApiSpec.ts
```diff
@@ -0,0 +1,49 @@
...
+  private inject_additional_properties(ctx: SpecificationContext, spec: OpenAPIV3.Document) {
+    const visitor = new SchemaVisitor((_ctx, schema) => {
+      if ('properties' in schema) {
+        // causes any undeclared field in the response to produce an error
+        (schema as any).additionalProperties = {
+          not: true,
+          errorMessage: "property is not defined in the spec"
+        }
+      }
+    });
```
nhtruong, line 45: This should check if the additionalProperties is null first.

nhtruong: I also wonder if making `addiltionalProperties: false` (when it's null) will have the same effect anyway without `ajv-errors`?

dblock: By default, with or without `ajv-errors`, `addiionalProperties: false` produces `(data/version must NOT have additional properties)` without specifying which property is missing. The missing properties are available in the `params` value returned from `validate`, but we want a useful error message.

```
[
  {
    instancePath: '/version',
    schemaPath: '#/additionalProperties',
    keyword: 'additionalProperties',
    params: { additionalProperty: 'distribution' },
    message: 'must NOT have additional properties'
  }
]
```

dblock:
@nhtruong @Xtansia this is ready for review

tools/src/merger/OpenApiMerger.ts
```diff
@@ -85,7 +85,9 @@ export default class OpenApiMerger {
       const spec = read_yaml(`${folder}/${file}`)
       const category = file.split('.yaml')[0]
       this.redirect_refs_in_schema(category, spec)
-      this.schemas[category] = spec.components.schemas as Record<string, OpenAPIV3.SchemaObject>
+      if (spec.components?.schemas !== undefined) {
+        this.schemas[category] = spec.components?.schemas
+      }
```
nhtruong, line 90: Is this change to make it easier to test? Because `spec.components.schemas` always exists in the actual spec. The Merger runs on the assumption that the Spec has past the Linter check.

dblock: Yes, the spec I added in https://github.com/dblock/opensearch-api-specification/tree/fail-on-additional-properties/tools/tests/tester/fixtures/specs/complete doesn't have this. It feels like it should be possible to have a valid spec without schemas, no?

tools/src/tester/MergedOpenApiSpec.ts
```diff
@@ -0,0 +1,49 @@
...
+    const spec = (new OpenApiMerger(this.file_path, this.logger)).merge()    
+    const ctx = new SpecificationContext(this.file_path)
+    this.inject_additional_properties(ctx, spec)
+    this._spec = spec
+    return this._spec
+  }
+
+  private inject_additional_properties(ctx: SpecificationContext, spec: OpenAPIV3.Document): void {
+    const visitor = new SchemaVisitor((_ctx, schema: any) => {
+      if (('required' in schema) && ('properties' in schema) && !('additionalProperties' in schema)) {
```
nhtruong, line 38: This could cause unintended effects if `required` is explicitly set to `false` and `properties` is explicitly set to `{}`

tools/src/tester/ResultLogger.ts
```diff
@@ -84,11 +84,6 @@ export class ConsoleResultLogger implements ResultLogger {
     const result = ansi.padding(this.#result(evaluation.result), 0, prefix)
     const message = evaluation.message != null ? `${ansi.gray('(' + evaluation.message + ')')}` : ''
     console.log(`${result} ${title} ${message}`)
-    if (evaluation.error != null && this._verbose) {
-      console.log('-'.repeat(100))
-      console.error(evaluation.error)
-      console.log('-'.repeat(100))
-    }
```
nhtruong, line 91: I probably missed it but is this being replaced by something else?

tools/src/tester/MergedOpenApiSpec.ts
dblock: It doesn't look like we have anything like that, required is usually a list of properties. What do you recommend this check be?

tools/src/tester/SchemaValidator.ts
```diff
@@ -25,6 +31,10 @@ export default class SchemaValidator {
   validate (schema: OpenAPIV3.SchemaObject, data: any): Evaluation {
     const validate = this.ajv.compile(schema)
     const valid = validate(data)
+    if (! valid) {
+      this.logger.info(`# ${JSON.stringify(schema, null, 2)}`)
+      this.logger.info(`* ${JSON.stringify(data, null, 2)}`)
```
nhtruong, line 36: Nit: this is used often enough to DRY it with `to_json` function in `helpers.ts` and to keep the JSON format consistent throughout the framework.

tools/src/merger/OpenApiMerger.ts
nhtruong: An OpenAPI spec, yes. But not an OpenSearch OpenAPI Spec. I'm just nit picking. This bit of code is harmless to the main use of the merger.

tools/src/tester/MergedOpenApiSpec.ts
nhtruong: We also do not have  lint to prevent those either. You can do:
`if (schema.required && schema.properties && !schema.additionalProperties)`
OR if you prefer explicit null checks:
`if (schema.required != null && schema.properties !=null && schema.additionalProperties == null)`

tools/src/tester/SchemaValidator.ts
dblock: Good idea, https://github.com/opensearch-project/opensearch-api-specification/pull/342/commits/b8484ac83bcb7ccdc3ccf0987e158893917adae2.

tools/src/tester/MergedOpenApiSpec.ts
dblock: Updated.

nhtruong (approved):

Issue #350, opened 2024-06-21
[BUG] BulkRequest with scripted noop upsert UpdateOperation fails with "missing required property".


BrendonFaleiro:
### What is the bug?
The `source` in InlineGet should be optional. The when a scripted update is a noop, there is no source response. 

### How can one reproduce the bug?
See https://github.com/opensearch-project/opensearch-java/issues/1042

PR #351, opened 2024-06-21
Made the InlineGet `source` field optional.

BrendonFaleiro:
Address https://github.com/opensearch-project/opensearch-api-specification/issues/350

Signed-off-by: Brendon Faleiro <faleiro@amazon.com>


dblock (changes requested):
Thanks. Do you have time to write a test for this change? We're trying to use regressions to improve coverage.

Something minimal that calls this API, it should fail on the schema check as long as it doesn't return `_source`, https://github.com/opensearch-project/opensearch-api-specification/tree/main/tests

dblock:
I added a test on top of this change in https://github.com/opensearch-project/opensearch-api-specification/pull/375.

PR #358, opened 2024-06-21
Remove ability to skip stories in tests.

dblock:
### Description

Potentially controversial @nhtruong :)

I find the `skip: true` option harmful. I fail to see when I'd want to use this. Either I run individual tests (supported with `--tests`) or the entire suite. Disabling a solo test story seems to be an invitation for skipping something that's flaky, for example. If I really want to disable one locally, I can just comment it out. 


nhtruong (approved):
It's just a feature that's present in many test frameworks. It's for WIP tests or turning off flaky tests while working on solutions to resolve the flaky tests. Since we're not building an app with interwoven components, the need for this feature is very small. I'm okay with removing it to avoid it being abused.

PR #359, opened 2024-06-24
Test framework/guide

nhtruong:
- Reduced noise when running spec tests
- `--verbose` now provides better context for non-2XX responses.
- Moved `linter/utils` to `_utils` because it's also being used by the tester tool (To avoid misleading stacktrace when an error appears in utils)
- Added SPEC_TESTING.md


SPEC_TESTING.md
```diff
@@ -0,0 +1,136 @@
...
+* [SPEC TESTING](#spec-testing)
+  * [Running Spec Tests Locally](#running-spec-tests-locally)
+  * [Writing Spec Tests](#writing-spec-tests)
+    * [Simple Test Story](#simple-test-story)
+    * [Using Output from Previous Chapters](#using-output-from-previous-chapters)
+<!-- TOC -->
+
+# SPEC TESTING
+
+We devise our own test framework to test the spec against an OpenSearch cluster. We're still adding more features to the framework as the needs arise, and this document will be updated accordingly. This test framework has also been integrated into the repo's CI/CD pipeline. Checkout the [test-spec](.github/workflows/test-spec.yml) workflow for more details.
```
Xtansia, line 11: I think tense wise this should be `We have devised...`

tools/src/linter/InlineObjectSchemaValidator.ts
```diff
@@ -10,8 +10,8 @@
 import type NamespacesFolder from './components/NamespacesFolder'
 import type SchemasFolder from './components/SchemasFolder'
 import { type ValidationError } from 'types'
-import { SchemaVisitor } from './utils/SpecificationVisitor'
-import { is_ref, type MaybeRef, SpecificationContext } from './utils'
+import { SchemaVisitor } from '../_utils/SpecificationVisitor'
+import { is_ref, type MaybeRef, SpecificationContext } from '../_utils'
```
Xtansia, line 14: What's the reason for the leading underscore on `_utils`?

tools/src/tester/test.ts
```diff
@@ -7,22 +7,23 @@
...
+  OPENSEARCH_USERNAME_OPTION,
+  OpenSearchHttpClient
 } from '../OpenSearchHttpClient'
 import ChapterReader from './ChapterReader'
 import ChapterEvaluator from './ChapterEvaluator'
 import OperationLocator from './OperationLocator'
 import SchemaValidator from './SchemaValidator'
 import StoryEvaluator from './StoryEvaluator'
-import { ConsoleResultLogger } from './ResultLogger'
+import {ConsoleResultLogger} from './ResultLogger'
```
Xtansia, line 26: I'm surprised the linter isn't complaining about this as it should be enforcing spacing around the braces (or at least it was)

Xtansia (changes requested):

tools/src/tester/test.ts
nhtruong: I setup my IDE to run `--fix` on save. I thought the linter wanted space too.

tools/src/linter/InlineObjectSchemaValidator.ts
nhtruong: To differentiate it from other folders like `tester` `linter` `merger` where they are tools while `_utils` is not a tool.

SPECIFICATION_TESTING.md
```diff
@@ -0,0 +1,136 @@
+<!-- TOC -->
```
dblock, line 1: We don't have these markers in other docs, maybe remove? The [Markdown all in one](https://marketplace.visualstudio.com/items?itemName=yzhang.markdown-all-in-one) vscode plugin doesn't require them.

tools/src/tester/test.ts
dblock: This looks suspicious. Other files have a space. Something is off with the eslint rule file being used?

https://github.com/opensearch-project/opensearch-api-specification/blob/main/tools/src/coverage/coverage.ts#L12

dblock (changes requested):
Nits + let's resolve this linter problem.

Feel free to break up the PR, the doc looks good.

SPECIFICATION_TESTING.md
nhtruong: It appears in [DEVELOPER_GUIDE](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md?plain=1#L1), too. Shall I remove it there too?

dblock: Yes, but I am nitpicking.

tools/src/tester/test.ts
nhtruong: Fixed it. Didn't dig too deep but `object-curly-spacing` rule was deprecated in ESLint v8.53.0 and moved to a different plugin. Installing the plugin and enabled that rule solved it.

SPECIFICATION_TESTING.md
dblock: Note that it's not valid markdown ;)

dblock (approved):

SPECIFICATION_TESTING.md
Xtansia: It's just an HTML comment, and inline raw HTML is valid according to:
- [GitHub Flavored Markdown](https://github.github.com/gfm/#raw-html)
- [CommonMark](https://spec.commonmark.org/0.31.2/#raw-html)
- [John Gruber's Original Definition](https://daringfireball.net/projects/markdown/syntax#html)

Xtansia (approved):

SPECIFICATION_TESTING.md
dblock: I stay corrected.

PR #362, opened 2024-06-26
Added StoryValidator

### LINKED ISSUE #354: [FEATURE] Validate test story schema.

Currently if you accidentally nest a field, such as `payload` or `request_body` in the wrong place in a test story, it will get ignored, which may lead to false positives in tests.

Validate schema of test stories being run to strictly conform.

nhtruong:
Validating stories before running them.

### Issues Resolved
closes #354 

### DIFF
--- a/tests/_core/search.yaml
+++ b/tests/_core/search.yaml
@@ -1,6 +1,5 @@
 $schema: ../../json_schemas/test_story.schema.yaml
 
-skip: false
 description: Test search endpoint.
 prologues:
   - path: /movies/_doc
@@ -12,8 +11,7 @@ prologues:
         director: Bennett Miller
         title: Moneyball
         year: 2011
-    response:
-      status: 201
+    status: [201]
 epilogues:
   - path: /movies
     method: DELETE
--- a/tools/src/tester/ResultLogger.ts
+++ b/tools/src/tester/ResultLogger.ts
@@ -29,15 +29,17 @@ export class ConsoleResultLogger implements ResultLogger {
   }
 
   log (evaluation: StoryEvaluation): void {
+    const with_padding = [Result.FAILED, Result.ERROR].includes(evaluation.result) || this._verbose
+    if (with_padding) console.log()
     this.#log_story(evaluation)
     this.#log_chapters(evaluation.prologues ?? [], 'PROLOGUES')
     this.#log_chapters(evaluation.chapters ?? [], 'CHAPTERS')
     this.#log_chapters(evaluation.epilogues ?? [], 'EPILOGUES')
-    console.log('\n')
+    if (with_padding) console.log()
   }
 
-  #log_story ({ result, full_path, description, display_path, message }: StoryEvaluation): void {
-    this.#log_evaluation({ result, message: message ?? full_path }, ansi.cyan(ansi.b(description ?? display_path)))
+  #log_story ({ result, full_path, display_path, message }: StoryEvaluation): void {
+    this.#log_evaluation({ result, message: message ?? full_path }, ansi.cyan(ansi.b(display_path)))
   }
 
   #log_chapters (evaluations: ChapterEvaluation[], title: string): void {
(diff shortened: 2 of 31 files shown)


tests/_core/search.yaml
```diff
@@ -12,8 +11,7 @@ prologues:
         director: Bennett Miller
         title: Moneyball
         year: 2011
-    response:
-      status: 201
+    status: [201]
```
dblock, line 14: Why would this return an array?

tools/src/tester/ResultLogger.ts
```diff
@@ -29,15 +29,17 @@ export class ConsoleResultLogger implements ResultLogger {
   }
 
   log (evaluation: StoryEvaluation): void {
+    const with_padding = [Result.FAILED, Result.ERROR].includes(evaluation.result) || this._verbose
+    if(with_padding) console.log()
```
dblock, line 33: Lack of space after `if` looks suspicious, missing linter?

tools/src/tester/SchemaValidator.ts
```diff
@@ -32,7 +32,7 @@ export default class SchemaValidator {
   validate (schema: OpenAPIV3.SchemaObject, data: any): Evaluation {
     const validate = this.ajv.compile(schema)
     const valid = validate(data)
-    if (! valid) {
+    if (!valid) {
```
dblock, line 35: Missing lint?

tools/src/tester/StoryValidator.ts
```diff
@@ -0,0 +1,55 @@
...
+import addFormats from 'ajv-formats'
+import { read_yaml } from "../helpers";
+
+export default class StoryValidator {
+  private static readonly SCHEMA_FILE = path.resolve("./json_schemas/test_story.schema.yaml")
+  private readonly ajv: Ajv2019
+  private readonly validate_schema: ValidateFunction
+
+  constructor() {
+    this.ajv = new Ajv2019({ allErrors: true, strict: false })
```
dblock, line 22: We use plain `ajv` elsewhere, why is this using a different one?

dblock (commented):
Looks good, I have some questions.

I have a general concern for lack of unit tests. I understand that we're testing the end-to-end flow, but for any new given class like `StoryValidator` I'd expect to see a `StoryValidator.test.ts` that checks the basics at the very least.

tests/_core/search.yaml
nhtruong: In prologues and epilogues, we don't evaluate any spec but rather simple run the HTTP request to setup and tear down. By default, the framework accepts `status: [200, 201]` as okay and other code will make the prologue or epilogue return an error. So in this case in particular, we could have omitted `status`. Specifying status is usually for when cleanup where we want to delete a resource created in one of the chapters but we want to account for 404 where the resource was not created successfully. In this case, it's usually `status: [200, 404]`

nhtruong: https://github.com/opensearch-project/opensearch-api-specification/pull/299#discussion_r1619208061

tools/src/tester/StoryValidator.ts
nhtruong: It's because of the use of [unevaluatedProperties](https://github.com/opensearch-project/opensearch-api-specification/pull/362/files#diff-a885c8c21d6578474539d576ffef321a7b78bdff03e878868a1e6d37242eece9R26), which is a new feature in JSON Schema draft 2019-09. We use this in `test_story.schema` to enable `additionalProperties: false` for `allOf` objects. This was how I caught [this infraction](https://github.com/opensearch-project/opensearch-api-specification/pull/362/files#diff-016beb89472fc47fd47a2e9c8a9077f02954b509dae821c4c7fb821e26cbc9c6L15-L16).

tools/src/tester/SchemaValidator.ts
nhtruong: Same issue with the previous PR: the rule was moved into a plugin.

tools/src/tester/ResultLogger.ts
nhtruong: Same issue with the previous PR: the rule was moved into a plugin.
I've resolved.

tests/_core/search.yaml
nhtruong: Also added unit tests for StoryValidator

tools/src/tester/StoryValidator.ts
dblock: Should we upgrade all schema checks?

dblock (approved):
Get it to green and it’s good to go.

PR #363, opened 2024-06-26
Added support for application/yaml response payloads.

dblock:
### Description

On top of https://github.com/opensearch-project/opensearch-api-specification/pull/360, Added application/yaml support and responses to cat APIs. This one requires deserializing the response body into an object to validate.

### DIFF
--- a/spec/schemas/cat.indices.yaml
+++ b/spec/schemas/cat.indices.yaml
@@ -29,13 +29,13 @@ components:
           type: string
         docs.count:
           description: available docs
-          oneOf:
+          anyOf:
             - type: string
             - nullable: true
               type: string
         docs.deleted:
           description: deleted docs
-          oneOf:
+          anyOf:
             - type: string
             - nullable: true
               type: string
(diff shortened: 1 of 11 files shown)


spec/schemas/cat.indices.yaml
```diff
@@ -29,13 +29,13 @@ components:
           type: string
         docs.count:
           description: available docs
-          oneOf:
+          anyOf:
             - type: string
             - nullable: true
               type: string
```
nhtruong, line 35: probably scope creeping but `nullable` is not a proper JSON schema keyword. Even if it was, this is redundant because nullable string already encapsulates non-nullable string.

In this case and a few below, we can repalce `anyOf: ...` with `type: [string, null]` according to [this](https://json-schema.org/understanding-json-schema/reference/type)

dblock: I tried this but it failed with some other error. Care to open an issue with some details to replace?

nhtruong (dismissed):

nhtruong (approved):

nhtruong (approved):

nhtruong:
This is stuck at "Checking for ability to merge automatically…"

PR #365, opened 2024-06-27
Missing required fields

Bzil:
### Description
Related to issue https://github.com/opensearch-project/opensearch-java/issues/1040 

### Issues Resolved
[Bug](https://github.com/opensearch-project/opensearch-java/issues/1040)


dblock:
Thanks! This is a good start, keep iterating on the test for a basic nodes info response.

Please amend your commit for DCO with `git commit --amend -s`.

tests/_core/nodes.yaml
```diff
@@ -0,0 +1,13 @@
+$schema: ../../json_schemas/test_story.schema.yaml
+
+description: Test nodes endpoints.
+chapters:
+  - synopsis: Nodes with a json response.
+    path: /_cat/aliases
```
dblock, line 6: This should be `_nodes`.

dblock (changes requested):
The test doesn't seem to have the right path. Run it, make sure it fails without your fix and passes with it.

dblock:
Looks like a legit spec test failure.

dblock (approved):
Turned out to be a lot more than just a required/optional field! Good work.

Want to try and add tests/fix the [remainder of nodes API](https://opensearch.org/docs/latest/api-reference/nodes-apis/index/)? :)

Bzil:
Indeed, I will try to do this, but it's will depend of my time/work

PR #368, opened 2024-06-28
Add api spec for tiering of indices from hot to warm

neetikasinghal:
### Description

### Issues Resolved

### DIFF
--- a/spec/namespaces/indices.yaml
+++ b/spec/namespaces/indices.yaml
@@ -938,6 +938,24 @@ paths:
       responses:
         '200':
           $ref: '#/components/responses/indices.delete@200'
+  /{index}/_tier/warm:
+    post:
+      operationId: indices.tier_warm.0
+      x-operation-group: indices.tier
+      x-version-added: '1.0'
+      description: Tier the index from hot to warm in the cluster.
+      externalDocs:
+        url: XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
+      parameters:
+        - $ref: '#/components/parameters/indices.tier_warm::path.index'
+        - $ref: '#/components/parameters/indices.tier_warm::query.allow_no_indices'
+        - $ref: '#/components/parameters/indices.tier_warm::query.cluster_manager_timeout'
+        - $ref: '#/components/parameters/indices.tier_warm::query.ignore_unavailable'
+        - $ref: '#/components/parameters/indices.tier_warm::query.timeout'
+        - $ref: '#/components/parameters/indices.tier_warm::query.wait_for_completion'
+      responses:
+        '200':
+          $ref: '#/components/responses/indices.tier_warm@200'
   /{index}/_alias:
     get:
       operationId: indices.get_alias.2
@@ -2602,6 +2620,29 @@ components:
             required:
               - _shards
               - _all
+    indices.tier_warm@200:
+      description: 'Hot to warm tiering response'
+      content:
+        application/json:
+          schema:
+            type: object
+            properties:
+              acknowledged:
+                type: boolean
+              failed_indices:
+                type: array
+                items:
+                  type: object
+                  properties:
+                    key:
+                      type: string
+                    value:
+                      type: string
+                  required:
+                    - key
+                    - value
+            required:
+              - acknowledged
     indices.update_aliases@200:
       description: ''
       content:
@@ -4953,6 +4994,60 @@ components:
       schema:
         $ref: '../schemas/_common.yaml#/components/schemas/Level'
       style: form
+
+    indices.tier_warm::path.index:
+      in: path
+      name: index
+      description: |-
+        Comma-separated list of data streams, indices, and aliases to tier from hot to warm.
+        Supports wildcards (`*`).
+        To search all data streams or indices, omit this parameter or use `*` or `_all`.
+      required: true
+      schema:
+        $ref: '../schemas/_common.yaml#/components/schemas/Indices'
+      style: simple
+    indices.tier_warm::query.allow_no_indices:
+        in: query
+        name: allow_no_indices
+        description: |-
+          If `false`, the request returns an error if any wildcard expression, index alias, or `_all` value targets only missing or closed indices.
+          This behavior applies even if the request targets other open indices.
+        schema:
+          type: boolean
+          default: false
+        style: form
+    indices.tier_warm::query.cluster_manager_timeout:
+        name: cluster_manager_timeout
+        in: query
+        description: Operation timeout for connection to cluster-manager node.
+        schema:
+          $ref: '../schemas/_common.yaml#/components/schemas/Duration'
+        x-version-added: 2.0.0
+    indices.tier_warm::query.ignore_unavailable:
+        in: query
+        name: ignore_unavailable
+        description: If `false`, the request returns an error if it targets a missing or closed index.
+        schema:
+          type: boolean
+          default: false
+        style: form
+    indices.tier_warm::query.timeout:
+        in: query
+        name: timeout
+        description: |-
+          Period to wait for a response.
+          If no response is received before the timeout expires, the request fails and returns an error.
+        schema:
+          $ref: '../schemas/_common.yaml#/components/schemas/Duration'
+        style: form
+    indices.tier_warm::query.wait_for_completion:
+        name: wait_for_completion
+        in: query
+        description: Should this request wait until the operation has completed before returning.
+        schema:
+          type: boolean
+          default: true
+          description: Should this request wait until the operation has completed before returning.
     indices.update_aliases::query.cluster_manager_timeout:
       name: cluster_manager_timeout
       in: query
--- a/tests/indices/index.yaml
+++ b/tests/indices/index.yaml
@@ -67,6 +67,12 @@ chapters:
     parameters:
       index: books
 
+  - synopsis: Tier the `books` and `games` indices from hot to warm.
+    path: /{index}/_tier/warm
+    method: POST
+    parameters:
+      index: books,games
+
   - synopsis: Delete the `books` and `games` indices.
     path: /{index}
     method: DELETE


neetikasinghal:
@dblock please review, let me know if there is anything else that needs to be added.

spec/namespaces/indices.yaml
```diff
@@ -938,6 +938,24 @@ paths:
       responses:
         '200':
           $ref: '#/components/responses/indices.delete@200'
+  /{index}/_tier/warm:
+    post:
+      operationId: indices.tier_warm.0
+      x-operation-group: indices.tier
+      x-version-added: '1.0'
```
dblock, line 945: This should be the actual version when this is added.

spec/namespaces/indices.yaml
```diff
@@ -938,6 +938,24 @@ paths:
...
+      description: Tier the index from hot to warm in the cluster.
+      externalDocs:
+        url: XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
```
dblock, line 948: Omit if there's no doc.

dblock (changes requested):
Glad you're adding an API with specs! 

Will need working tests as part of the PR, otherwise looking ok.

spec/namespaces/indices.yaml
neetikasinghal: since this will be behind a feature flag for now, we havent released it yet - should this pr not be merged until its released?

neetikasinghal:
> Glad you're adding an API with specs!
> 
> Will need working tests as part of the PR, otherwise looking ok.

is there any other tests you are looking for me to be added apart from what i added or just making sure that all tests are successful?

dblock:
> > Glad you're adding an API with specs!
> > Will need working tests as part of the PR, otherwise looking ok.
> 
> is there any other tests you are looking for me to be added apart from what i added or just making sure that all tests are successful?

You'll want to exercise all the new API's query parameters.

spec/namespaces/indices.yaml
```diff
@@ -938,6 +938,24 @@ paths:
       responses:
         '200':
           $ref: '#/components/responses/indices.delete@200'
+  /{index}/_tier/warm:
+    post:
+      operationId: indices.tier_warm.0
+      x-operation-group: indices.tier
```
nhtruong, line 944: the `x-operation-group` will be used to generate the client method. In this case it will be `client.indices.tier()`. Usually the function is a verb if it's performing some action. Is `tier` appropriate for this context?

spec/namespaces/indices.yaml
neetikasinghal: @dblock what do u suggest here?

dblock: I think we need a bit of work to add support for experimental features, I opened https://github.com/opensearch-project/opensearch-api-specification/issues/400. Locally you can boot a next version of OpenSearch with this feature flag enabled to get a PR finished for now.

spec/namespaces/indices.yaml
neetikasinghal: makes sense, thanks for your input, let me update this.

neetikasinghal:
@dblock @nhtruong do u mind taking another look at the pr?

CHANGELOG.md
```diff
@@ -55,6 +55,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Added `plugins` to NodeInfoSettings ([#442](https://github.com/opensearch-project/opensearch-api-specification/pull/442)) 
 - Added test coverage ([#443](https://github.com/opensearch-project/opensearch-api-specification/pull/443))
 - Added `--opensearch-version` to `merger` that excludes schema elements per semver ([#428](https://github.com/opensearch-project/opensearch-api-specification/pull/428))
+- Added api spec for tiering of indices from hot to warm ([#368](https://github.com/opensearch-project/opensearch-api-specification/pull/368))
```
dblock, line 58: Let's say "Added `/{index}/_tier`" so that it's clear.

spec/namespaces/indices.yaml
```diff
@@ -1793,6 +1793,22 @@ paths:
       responses:
         '200':
           $ref: '#/components/responses/indices.stats@200'
+  /{index}/_tier/warm:
```
dblock, line 1796: Is `warm` not a parameter, will there be other tiers and/or will there be a `/index/_tier/hot` that takes a different set of parameters than `/index/_tier/warm`?

If the answer is yes, logically everything that's `tier_hot_to_warm` should just be `tier`, including file names, paths, parameters, etc.

spec/namespaces/indices.yaml
```diff
@@ -2743,6 +2759,22 @@ components:
             required:
               - _all
               - _shards
+    indices.tier_hot_to_warm@200:
+      description: Hot to warm tiering response
+      content:
+        application/json:
+          schema:
+            type: object
+            properties:
```
dblock, line 2768: Take the time to add `description` to all the properties. These will hopefully be used to generate docs soon.

spec/namespaces/indices.yaml
```diff
@@ -5126,6 +5158,59 @@ components:
...
+      description: If `false`, the request returns an error if it targets a missing or closed index.
+      schema:
+        type: boolean
+        default: false
+      style: form
+    indices.tier_hot_to_warm::query.timeout:
+      in: query
+      name: timeout
+      description: |-
+        Period to wait for a response.
```
dblock, line 5201: What's the unit for this timeout?

dblock (changes requested):
Good progress, close! See below and get all tests to pass.

neetikasinghal:
> Good progress, close! See below and get all tests to pass.

thanks @dblock for reviewing, successful run of tests is dependent on this pr: https://github.com/opensearch-project/opensearch-api-specification/pull/454

spec/namespaces/indices.yaml
neetikasinghal: added

spec/namespaces/indices.yaml
neetikasinghal: It depends on the unit specified by the user. Default is 30s. Something like: https://opensearch.org/docs/latest/api-reference/index-apis/close-index/

spec/namespaces/indices.yaml
neetikasinghal: there will be a `/index/_tier/hot` that could take different set of parameters

neetikasinghal: there could be some parameters which would be same, so added them under tier.

`/index/_tier/warm` could be shared for indices targeted to the warm tier - hot->warm/cold->warm

tests/indices/index.yaml
```diff
@@ -83,6 +83,15 @@ chapters:
     parameters:
       index: books
 
+  - synopsis: Tier the `books` and `games` indices from hot to warm.
```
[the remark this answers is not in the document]
neetikasinghal: done

spec/namespaces/indices.yaml
dblock: > there will be a `/index/_tier/hot` that could take different set of parameters

If you are not sure, consider that as written the clients will get a method generated like `index.tier_wam()`, so if you think it would make sense to generalize to something like `index.tier(Tier::WARM)` it would be better to do it early.

spec/namespaces/indices.yaml
```diff
@@ -5126,6 +5160,59 @@ components:
       schema:
         $ref: '../schemas/_common.yaml#/components/schemas/Level'
       style: form
+    indices.tier::path.index:
+      in: path
+      name: index
+      description: |-
+        Comma-separated list of data streams, indices, and aliases to tier from hot to warm.
+        Supports wildcards (`*`).
+        To search all data streams or indices, omit this parameter or use `*` or `_all`.
```
dblock, line 5169: To tier?

spec/schemas/tier.yaml
```diff
@@ -0,0 +1,17 @@
...
+  description: Schemas of tier_hot_to_warm category
+  version: 1.0.0
+paths: {}
+components:
+  schemas:
+    FailedIndex:
+      type: object
+      properties:
+        index:
+          description: Index name
```
dblock, line 13: Add a period.

spec/schemas/tier.yaml
```diff
@@ -0,0 +1,17 @@
...
+          $ref: '_common.yaml#/components/schemas/IndexName'
+        error:
+          description: Error message
```
dblock, line 16: Add a period.

tests/indices/tier/warm.yaml
```diff
@@ -0,0 +1,38 @@
...
+description: Test endpoints relevant to tiering of index to warm.
+epilogues:
+  - path: /books
+    method: DELETE
+    status: [200, 404]
+  - path: /games
+    method: DELETE
+    status: [200, 404]
+chapters:
+  - synopsis: Create an indices named `books` and `games` with default settings.
```
dblock, line 12: Should go into prologues.

tests/indices/tier/warm.yaml
```diff
@@ -0,0 +1,38 @@
...
+    path: /{index}
+    method: PUT
+    parameters:
+      index: books,games
+
+  - synopsis: Check if the indices `games` and `books` exists. They should.
```
dblock, line 18: This is not necessary for this test.

tests/indices/tier/warm.yaml
```diff
@@ -0,0 +1,38 @@
...
+      index: books,games
+      allow_no_indices: true
+      expand_wildcards: none
+      flat_settings: true
+      include_defaults: true
+      ignore_unavailable: true
+      local: true
+
+  - synopsis: Tier the `books` and `games` indices from hot to warm.
+    version: '>= 2.16'
```
dblock, line 31: Since this tier test is entirely >= 2.16 move this `version` into the top level of the test, and it will be fully skipped.

tests/indices/tier/warm.yaml
```diff
@@ -0,0 +1,38 @@
...
+    path: /{index}/_tier/warm
+    method: POST
+    parameters:
+      index: books,games
+    response:
+      status: 200
```
dblock, line 37: Is there a way to know what tier my index is in?

dblock (changes requested):
Looks good, see some relatively minor asks below.

tests/indices/tier/warm.yaml
neetikasinghal: didnt know that, let me update this

tests/indices/tier/warm.yaml
neetikasinghal: that will be available via _cat/indices later or _cat/tiering api, changes to which are not done yet. This test can be updated once its available.

spec/namespaces/indices.yaml
neetikasinghal: I will keep the API separate for hot/warm instead of a generic tier. This will help to support the new parameters relevant to that target tier movement in its specific request type. With same API it will start to become too convoluted when parameters doesn't make sense for all the tiers. For example: In hot to warm case, we can provide an option to perform force merge in hot/warm nodes, which can be controlled by the request level options. Similarly for warm to hot, we can choose to provide a mechanism to control whether say we want to apply some new index settings such as increase replica count.

neetikasinghal:
> > Good progress, close! See below and get all tests to pass.
> 
> thanks @dblock for reviewing, successful run of tests is dependent on this pr: #454

@dblock should we add a higher version on the tests as the feature flag pr(https://github.com/opensearch-project/opensearch-api-specification/pull/454) is blocked on the other fields being added to the spec first? thoughts?

dblock:
> @dblock should we add a higher version on the tests as the feature flag pr(#454) is blocked on the other fields being added to the spec first? thoughts?

That would be semantically incorrect. Let's sort these PRs one after another.

neetikasinghal:
> > @dblock should we add a higher version on the tests as the feature flag pr(#454) is blocked on the other fields being added to the spec first? thoughts?
> 
> That would be semantically incorrect. Let's sort these PRs one after another.

@dblock raised a pr in opensearch core package to fix the bug:  https://github.com/opensearch-project/OpenSearch/pull/15076

dblock:
@neetikasinghal Looks like the OpenSearch PR got merged, want to finish this and add a 2.17 to CI?

neetikasinghal:
> @neetikasinghal Looks like the OpenSearch PR got merged, want to finish this and add a 2.17 to CI?

sure, what would be the @sha256 for 2.17? 
Also, the spell checker complains of the Dopensearch here(https://github.com/opensearch-project/opensearch-api-specification/actions/runs/10205504557/job/28236397605?pr=368), any suggestions to avoid this?

dblock:
> > @neetikasinghal Looks like the OpenSearch PR got merged, want to finish this and add a 2.17 to CI?
> 
> sure, what would be the @sha256 for 2.17? 

I look at https://hub.docker.com/r/opensearchstaging/opensearch/tags for a tag.

> Also, the spell checker complains of the Dopensearch here(https://github.com/opensearch-project/opensearch-api-specification/actions/runs/10205504557/job/28236397605?pr=368), any suggestions to avoid this?

Just add it to .cspell.

neetikasinghal:
@dblock there would be some more needs for the test to return a status of 200 - e.g. it needs a node with search role configured, else the API would throw a 400.

dblock:
> @dblock there would be some more needs for the test to return a status of 200 - e.g. it needs a node with search role configured, else the API would throw a 400.

Yeah, so this would go into docker-compose. 

There's a similar ask in https://github.com/opensearch-project/opensearch-api-specification/pull/464 for a bunch of configuration. I think we need to find a way not to cram all possible combinations of feature flags and tests into one place, or it will get totally unmanageable. 

What do you think about adding folders to `tests/`, for example `tests/features/experimental/tiered_remote_index` in which we'd locate a `docker-compose.yml` and YAML tests just for that feature? We can add another GHA workflow that only deals with things in `tests/features` differently from the main tests that we have now?

neetikasinghal:
> > @dblock there would be some more needs for the test to return a status of 200 - e.g. it needs a node with search role configured, else the API would throw a 400.
> 
> Yeah, so this would go into docker-compose.
> 
> There's a similar ask in #464 for a bunch of configuration. I think we need to find a way not to cram all possible combinations of feature flags and tests into one place, or it will get totally unmanageable.
> 
> What do you think about adding folders to `tests/`, for example `tests/features/experimental/tiered_remote_index` in which we'd locate a `docker-compose.yml` and YAML tests just for that feature? We can add another GHA workflow that only deals with things in `tests/features` differently from the main tests that we have now?

Thats not a bad idea, however these features would be pulled in as well into the main code someday, so there should be an easy way to remove the feature flags and still have the tests working as is. Else, there will be a lot of work to remove the feature flags. So, not sure if we should isolate the features completely. 

On a separate note: For tiering specifically, its needed for indices to be backed by remote store. Is there a current test that tests the remote store functionality? If not, that is something, which would need to be added for long run also(not just for experimental features.)

dblock:
> Thats not a bad idea, however these features would be pulled in as well into the main code someday, so there should be an easy way to remove the feature flags and still have the tests working as is. Else, there will be a lot of work to remove the feature flags. So, not sure if we should isolate the features completely.

I implemented it in https://github.com/opensearch-project/opensearch-api-specification/pull/472 and it got merged. When a feature goes from needing a custom docker-compose to not, we can move the test files to the `default` folder and delete the customer docker-compose.
 
> On a separate note: For tiering specifically, its needed for indices to be backed by remote store. Is there a current test that tests the remote store functionality? If not, that is something, which would need to be added for long run also(not just for experimental features.)

AFAIK not.

neetikasinghal:
> > Thats not a bad idea, however these features would be pulled in as well into the main code someday, so there should be an easy way to remove the feature flags and still have the tests working as is. Else, there will be a lot of work to remove the feature flags. So, not sure if we should isolate the features completely.
> 
> I implemented it in #472 and it got merged. When a feature goes from needing a custom docker-compose to not, we can move the test files to the `default` folder and delete the customer docker-compose.
> 
> > On a separate note: For tiering specifically, its needed for indices to be backed by remote store. Is there a current test that tests the remote store functionality? If not, that is something, which would need to be added for long run also(not just for experimental features.)
> 
> AFAIK not.

@dblock to unblock the specs being present in the package, can we merge this change without the tests and i can add follow-up issues to add the setup for the tests required to run successfully? what do u think?

dblock:
> @dblock to unblock the specs being present in the package, can we merge this change without the tests and i can add follow-up issues to add the setup for the tests required to run successfully? what do u think?

I'd rather not. What's stopping you from making the tests work?

neetikasinghal:
> > @dblock to unblock the specs being present in the package, can we merge this change without the tests and i can add follow-up issues to add the setup for the tests required to run successfully? what do u think?
> 
> I'd rather not. What's stopping you from making the tests work?

I was just avoiding a lot of changes as part of this pr only as those could be commonly used by other API spec tests as well when they are added and was trying to unblock the tiering API specs(this pr) because of that.

However, I understand that we should not ship the code without coverage for it.

Issue #352, opened 2024-06-21
[FEATURE] Sort spec fields (e.g. parameters)


dblock:
### Is your feature request related to a problem?

Looking at the missing parameter in #273 we find unsorted refs.

```
        - $ref: '#/components/parameters/search::query.allow_partial_search_results'
        - $ref: '#/components/parameters/search::query.typed_keys'
        - $ref: '#/components/parameters/search::query.version'
        - $ref: '#/components/parameters/search::query.seq_no_primary_term'
```

It's difficult to see which fields are missing when comparing to the docs in https://opensearch.org/docs/latest/api-reference/search/.

### What solution would you like?

Similar to #272, sort everything that makes sense to be sorted alphabetically, starting with parameters.

PR #369, opened 2024-06-28
Added a linter for order of parameters.

dblock:
### Description

Require `parameters` to be sorted.

Here's a silly Ruby script to apply a sort for next linter that needs sorted something.

```ruby
Dir["spec/**/*.yaml"].each do |filename|
    puts filename
    lines = []
    File.open(filename, "r") do |f|
        params = []
        in_params = false
        f.each_line do |line|
            if line.include?('parameters:')
                lines << line
                in_params = true
            elsif in_params && line.include?("- $ref: ")
                params << line
            elsif in_params && params.size > 0
                lines.concat(params.sort)
                lines << line
                params = []
                in_params = false
            else
                lines << line
            end
        end
    end
    File.open(filename, "w+") do |f|
        f.puts(lines)
    end
end
```

### Issues Resolved

Closes https://github.com/opensearch-project/opensearch-api-specification/issues/352.


package.json
```diff
@@ -12,6 +12,7 @@
     "lint--fix": "eslint . --fix --report-unused-disable-directives",
     "merge": "ts-node tools/src/merger/merge.ts",
     "test": "npm run test:unit && npm run test:integ",
+    "jest": "jest",
```
dblock, line 15: I put this back because to run an individual test you can't do `npm run test:unit -- ...test.ts`, it has `testMatch` flags, but you can do `npm run jest -- ...test.ts`.

Xtansia (approved):

PR #377, opened 2024-07-04
[WIP] Add test node

Bzil:
TODO 

### Description

### Issues Resolved

### DIFF
--- a/spec/namespaces/nodes.yaml
+++ b/spec/namespaces/nodes.yaml
@@ -268,6 +268,21 @@ paths:
       responses:
         '200':
           $ref: '#/components/responses/nodes.info@200'
+  /_nodes/{metric}/stats:
+    get:
+      operationId: nodes.info.1
+      x-operation-group: nodes.info
+      x-version-added: '1.0'
+      description: Returns information about nodes in the cluster.
+      externalDocs:
+        url: https://opensearch.org/docs/latest/api-reference/nodes-apis/nodes-info/
+      parameters:
+        - $ref: '#/components/parameters/nodes.info::path.metric'
+        - $ref: '#/components/parameters/nodes.info::query.flat_settings'
+        - $ref: '#/components/parameters/nodes.info::query.timeout'
+      responses:
+        '200':
+          $ref: '#/components/responses/nodes.info@200'
   /_nodes/{node_id}:
     get:
       operationId: nodes.info.2
--- a/spec/schemas/nodes.info.yaml
+++ b/spec/schemas/nodes.info.yaml
@@ -22,6 +22,7 @@ components:
         - indices
         - search_pipelines
         - _all
+        - _cluster_manager
     ResponseBase:
       allOf:
         - type: object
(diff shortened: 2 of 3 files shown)


spec/namespaces/nodes.yaml
```diff
@@ -268,6 +268,21 @@ paths:
       responses:
         '200':
           $ref: '#/components/responses/nodes.info@200'
+  /_nodes/{metric}/stats:
```
Bzil, line 271: // Read https://opensearch.org/docs/latest/api-reference/nodes-apis/index/

tests/_core/nodes.yaml
```diff
@@ -7,3 +7,53 @@ chapters:
...
+      metric:
+        - _cluster_manager
+    response:
+      status: 200
+  - synopsis: Nodes _cluster_manager metric with a json response.
+    path: /_nodes/{metric}/stats
+    method: GET
+    parameters:
+      metric:
+        - data:true
```
Bzil, line 48: @dblock do you have any idea how to implement a fix for this test ?

dblock: This error?

```
FAILED  Nodes _cluster_manager metric with a json response. 
            FAILED  PARAMETERS 
                FAILED  metric (data/0 must be equal to one of the allowed values)
```

The `data:true` value doesn't seem like a valid metric.

Bzil: it's seems weird yes, but it's define in documentation

dblock: Is it read like a JSON value? Quote it, `"data:true"`?

Bzil: if I do that, that implies that I have to put it in the enum and therefore do I have to do the same for `data:false` and the other filters?

dblock: I suppose that yes? There's really no such thing as booleans in URL paths AFAIK, this is a very creative use of the API.

spec/schemas/nodes.info.yaml
```diff
@@ -21,21 +21,24 @@ components:
...
+                - type: object
+                  $ref: '#/components/schemas/NodeInfo'
+                - type: object
+                  additionalProperties: {}
             _nodes:
                 $ref: 'nodes._common.yaml#/components/schemas/NodesResponseBase'
           required:
             - cluster_name
-            - nodes
             - _nodes
```
Xtansia, line 42: These changes are likely due to the duplicate path and having the wrong schemas, please revert them and test again with other changes made.

tests/_core/nodes.yaml
```diff
@@ -7,3 +7,53 @@ chapters:
     method: GET
```
Xtansia, line 7: This file should be moved under `tests/nodes/` and split into `info.yaml` and `stats.yaml`

tests/_core/nodes.yaml
Xtansia: These should be using `/_nodes/{node_id}/stats`

tests/_core/nodes.yaml
```diff
@@ -7,3 +7,53 @@ chapters:
...
+    path: /_nodes/{metric}
+    method: GET
+    parameters:
+      metric:
+        - jvm
+        - thread_pool
+    response:
+      status: 200
+  - synopsis: Nodes _all metric with a json response.
+    path: /_nodes/{metric}
```
Xtansia, line 28: This one is technically `/_nodes/{node_id}`

Xtansia (changes requested):

dblock:
Last tests from this PR made it into https://github.com/opensearch-project/opensearch-api-specification/pull/455.

PR #378, opened 2024-07-05
Refactored AJV usages with an AJV wrapper that provides better error JSON validation messages.

### LINKED ISSUE #364: [FEATURE] Failing story validation does not tell which property is unevaluated

After #362 if a story has an unevaluated property the error doesn't say which property is unevaluated.

In the following example `synopsis` is invalid.

```
prologues:
  - synopsis: Create an index named `books` with mappings and settings.
    path: /{index}
    method: PUT
    parameters:
      index: books
    request_body:
      payload: {}
```

When running tests.

```
ERROR   cat/indices.yaml (Invalid Story: data/prologues/0 must NOT have unevaluated properties)
```

Similar to the use of `ajv_errors` in the response body/schema validator.

nhtruong:
JsonSchemaValidator
- Wrapper for AJV to streamline Json validation logic.
- Collapse multiple `required` errors into one
- Collapse multiple `additionalProperties` and `unevaluatedProperties` errors into one, and print out which properties are prohibited.
- Make `enum` errors print out the supported enum values.

### Issues Resolved
closes #364 


tools/src/_utils/AjvErrorsParser.ts
```diff
@@ -0,0 +1,98 @@
...
+    const allowed_values = errors[0].params.allowedValues.join(', ')
+    return {
+      keyword: 'enum',
+      instancePath: errors[0].instancePath,
+      schemaPath: errors[0].schemaPath,
+      params: errors[0].params,
+      message: `MUST be equal to one of the allowed values: ${allowed_values}`
+    }
+  }
+
```
dblock, line 96: All this space ;)

tools/src/linter/SchemasValidator.ts
```diff
@@ -30,17 +26,19 @@ export default class SchemasValidator {
...
     this.root_folder = root_folder
     this.ajv = new AJV({ strict: true, discriminator: true })
     addFormats(this.ajv)
     for (const keyword of ADDITIONAL_KEYWORDS) this.ajv.addKeyword(keyword)
+    this.errors_parser = new AjvErrorsParser(this.ajv, {})
   }
 
   validate (): ValidationError[] {
-    this.spec = new OpenApiMerger(this.root_folder, this.logger).merge().components as Record<string, any>
+    this.spec = new OpenApiMerger(this.root_folder, new Logger(LogLevel.error)).merge().components as Record<string, any>
```
dblock, line 41: Why would we not want to pass down the top level logger?

tools/src/tester/SchemaValidator.ts
```diff
@@ -35,6 +37,7 @@ export default class SchemaValidator {
     ajv_errors(this.ajv, { singleError: true })
     const schemas = spec.components?.schemas ?? {}
     for (const key in schemas) this.ajv.addSchema(schemas[key], `#/components/schemas/${key}`)
+    this.errors_parser = new AjvErrorsParser(this.ajv)
```
dblock, line 40: This is an OO anti-pattern here. The error parser is really something that only applies to `this.ajv`, so it would be better to further subclass `AJV` into a `JsonSchemaValidator` (a generic JSON schema validator with error parsing), and make error parser a member of that, possibly hiding away more implementation details that are in `SchemaValidator` here and having its own tests. It would replace a few copy-pastes below. Can be done later of course.

dblock (changes requested):
I like it. The `new Logger` is the only must have.

tools/src/tester/SchemaValidator.ts
nhtruong: I also notice the same issue with AJV itself. That pattern is kinda already been broken by AJV when you have to do `ajv.errorsText(validate_func.errors)` instead of `validate_function.errorsText`. I can make a subclass for AJV to override its `errorsText` function but that means we will be locked to a specific JSON Schema draft (AJV vs AVJ2019 vs AJV2020).

tools/src/linter/SchemasValidator.ts
nhtruong: Because we're printing out warnings for the mergers when linting, while those are not the concerns of the linter. Those warnings are already being printed out by running the merger itself. The goal was reducing noise for the linter.

tools/src/_utils/AjvErrorsParser.ts
nhtruong: Im surprised the linter didn't complain :)

tools/src/tester/SchemaValidator.ts
nhtruong: Actually it's not hard to subclass JsonSchemaValidator into JsonSchemaValidator2019 and JsonSchemaValidator2020.

tools/src/linter/SchemasValidator.ts
dblock: I see, makes sense.

nhtruong:
@dblock This is now ready for review

tools/src/_utils/AjvErrorsParser.ts
```diff
@@ -0,0 +1,96 @@
...
+  parse(errors: ErrorObject[] | undefined | null): string {
+    const error_groups = this.#group_errors(errors ?? [])
+    const parsed_errors = [
+      this.#prohibited_property_error(error_groups.prohibited),
+      this.#required_property_error(error_groups.required),
+      this.#enum_error(error_groups.enum),
+      ...error_groups.others
+    ].filter(e => e != null) as ErrorObject[]
```
dblock, line 29: I think `_.compact()` will do the same, possibly cleaner.

tools/src/_utils/AjvErrorsParser.ts
```diff
@@ -0,0 +1,96 @@
...
+    return this.ajv.errorsText(parsed_errors, this.options)
+  }
+
+  #group_errors(errors: ErrorObject[]): { required: ErrorObject[], prohibited: ErrorObject[], enum: ErrorObject[], others: ErrorObject[] } {
```
dblock, line 33: Looks like it didn't get formatted properly ... { ... }

tools/src/linter/SchemasValidator.ts
```diff
@@ -100,16 +92,10 @@ export default class SchemasValidator {
...
-    try {
-      if (key != null) this.ajv.addSchema(schema, key)
-      this.ajv.compile(schema)
-    } catch (_e: any) {
-      const error = _e as Error
-      for (const prefix of IGNORED_ERROR_PREFIXES) if (error.message.startsWith(prefix)) return
-      return error
-    }
+    const message = this.json_validator.validate_schema(schema)
+    if (message != null) return message
```
dblock, line 98: Is this because of `null` vs. `undefined`? Should be able to just `return this.json_validator.validate_schema(schema)`.

dblock (approved):
I made some nitty comments, address them before merging?

tools/src/_utils/AjvErrorsParser.ts
nhtruong: I wish it was like `Array.compact()` because this is more `()` nesting 😅

tools/src/linter/SchemasValidator.ts
nhtruong: Nope just some brain fart when refactoring. Thanks for catching that.

dblock (approved):

PR #385, opened 2024-07-08
Fix: GET _mapping with index in query.

dblock:
### Description

OpenSearch supports `GET _mapping?index=...`. If you specify it both on the path and query the query is ignored.

```
$ curl -u admin:$OPENSEARCH_PASSWORD -k https://localhost:9200/_mapping?index=games | jq

{
  "games": {
    "mappings": {}
  }
}
```

```
$ curl --silent -u admin:$OPENSEARCH_PASSWORD -k https://localhost:9200/games/_mapping?index=foobars | jq
{
  "games": {
    "mappings": {}
  }
}
```

### DIFF
--- a/spec/namespaces/indices.yaml
+++ b/spec/namespaces/indices.yaml
@@ -418,6 +418,7 @@ paths:
         - $ref: '#/components/parameters/indices.get_mapping::query.cluster_manager_timeout'
         - $ref: '#/components/parameters/indices.get_mapping::query.expand_wildcards'
         - $ref: '#/components/parameters/indices.get_mapping::query.ignore_unavailable'
+        - $ref: '#/components/parameters/indices.get_mapping::query.index'
         - $ref: '#/components/parameters/indices.get_mapping::query.local'
         - $ref: '#/components/parameters/indices.get_mapping::query.master_timeout'
       responses:
@@ -1302,6 +1303,7 @@ paths:
         - $ref: '#/components/parameters/indices.get_mapping::query.cluster_manager_timeout'
         - $ref: '#/components/parameters/indices.get_mapping::query.expand_wildcards'
         - $ref: '#/components/parameters/indices.get_mapping::query.ignore_unavailable'
+        - $ref: '#/components/parameters/indices.get_mapping::query.index'
         - $ref: '#/components/parameters/indices.get_mapping::query.local'
         - $ref: '#/components/parameters/indices.get_mapping::query.master_timeout'
       responses:
@@ -3828,6 +3830,17 @@ components:
       schema:
         type: boolean
       style: form
+    indices.get_mapping::query.index:
+      in: query
+      name: index
+      description: |-
+        Comma-separated list of data streams, indices, and aliases used to limit the request.
+        Supports wildcards (`*`).
+        To target all data streams and indices, omit this parameter or use `*` or `_all`.
+      required: true
+      schema:
+        $ref: '../schemas/_common.yaml#/components/schemas/Indices'
+      style: simple
     indices.get_mapping::query.local:
       in: query
       name: local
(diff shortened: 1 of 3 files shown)


spec/namespaces/indices.yaml
```diff
@@ -1288,23 +1289,23 @@ paths:
           $ref: '#/components/responses/indices.forcemerge@200'
   /{index}/_mapping:
     get:
-      operationId: indices.get_mapping.1
-      x-operation-group: indices.get_mapping
+      operationId: indices.get_index_mapping.0
+      x-operation-group: indices.get_index_mapping
```
Xtansia, line 1293: Renaming this will cause issues for client generation, as these two are already treated as the same operation merely picking between the paths based on index being specified or not

dblock: How should I solve this? The two APIs have effectively different parameters, as the `_mapping` API takes an index query parameter and the `{index}/_mapping` API does not. The linter complains that the two must have the same parameters.

Xtansia: .NET's logic is that it unions all parameters from all variants and just prefers using the path over the query string if two parameters with the same name but different locations exist. Maybe we could loosen the linter slightly to just validate the variants have the same set of named parameters regardless of location?

Is there any reason to ever use this query string variant over the path variant?

dblock: No reason, but the API supports it. We could of course not document it, but 🤷‍♂️ 

WDYT @nhtruong?

nhtruong: @dblock Our clients support optional path parameters. Operations of the same group are only required to share an identical set of querystring params. The only requirement for path params is that if 2 operations support the same path params, then those params must be identical.

nhtruong: The [Clients Generator Guide](https://github.com/opensearch-project/opensearch-api-specification/blob/main/CLIENT_GENERATOR_GUIDE.md) touched on this:
> If two operations have identical HTTP methods, but different paths: use the path that best matches the path parameters provided.

nhtruong: > How should I solve this? The two APIs have effectively different parameters, as the _mapping API takes an index query parameter and the {index}/_mapping API does not. The linter complains that the two must have the same parameters.

The clients prioritize path parameters. If `index` is provided in the `clients.indices.get_mapping({ index: 'books' })` then `index` will extracted as a path param, and `/{index}/_mapping` will be used.

nhtruong (approved):

spec/namespaces/indices.yaml
dblock: Ended up just adding it to possible query parameters, thanks both!

spec/namespaces/indices.yaml
```diff
@@ -3828,6 +3830,17 @@ components:
...
         type: boolean
       style: form
+    indices.get_mapping::query.index:
+      in: query
+      name: index
+      description: |-
+        Comma-separated list of data streams, indices, and aliases used to limit the request.
+        Supports wildcards (`*`).
+        To target all data streams and indices, omit this parameter or use `*` or `_all`.
+      required: true
```
nhtruong, line 3840: I missed this. it should have been false

dblock: Yikes

spec/namespaces/indices.yaml
nhtruong: @Xtansia The clients gen will need to account for this situation when there are path and query params that share the same names. Especially for method documentation and typing for the `params` object. Should we also enforce that params of the same name, regardless of whether they are path or query, should have identical description and schema?

Issue #372, opened 2024-07-02
[FEATURE] Add cat smile format responses w/tests


dblock:
### Is your feature request related to a problem?

OpenSearch `_cat` APIs support the [SMILE](https://github.com/FasterXML/smile-format-specification) format. 

### What solution would you like?

Similar to https://github.com/opensearch-project/opensearch-api-specification/pull/371, add support for `application/smile`.

### Do you have any additional context?

See https://code.dblock.org/2024/07/02/using-opensearch-cat-api-response-formats.html for a walkthrough of `_cat` API formats.

PR #386, opened 2024-07-08
Add support for application/smile

dblock:
### Description

Add support for application/smile.

Thanks to @ngyewch for the recent work on the smile-js library to get this over the line!

### Issues Resolved

Closes https://github.com/opensearch-project/opensearch-api-specification/issues/372.

### DIFF
--- a/tests/cat/health.yaml
+++ b/tests/cat/health.yaml
@@ -82,3 +82,16 @@ chapters:
           status: yellow
           node.data: '1'
           discovered_cluster_manager: 'true'
+  - synopsis: Cat in different formats (format=smile).
+    method: GET
+    path: /_cat/health
+    parameters:
+      format: smile
+    response:
+      status: 200
+      content_type: application/smile
+      payload:
+        - node.total: '1'
+          status: yellow
+          node.data: '1'
+          discovered_cluster_manager: 'true'
--- a/tools/tests/tester/fixtures/stories/passed.yaml
+++ b/tools/tests/tester/fixtures/stories/passed.yaml
@@ -44,3 +44,13 @@ chapters:
       content_type: application/cbor
       payload:
         - node.total: '1'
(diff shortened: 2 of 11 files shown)


tests/cat/health.yaml
```diff
@@ -82,3 +82,16 @@ chapters:
...
           node.data: '1'
           discovered_cluster_manager: 'true'
+  - synopsis: Cat in different formats (format=smile).
+    method: GET
+    path: /_cat/health
+    parameters:
+      format: smile
+    response:
+      status: 200
+      content_type: application/yaml
```
Xtansia, line 92: Are we not actually validating `content_type` against the response content-type header?

tools/tests/tester/fixtures/stories/passed.yaml
```diff
@@ -44,3 +44,13 @@ chapters:
...
       payload:
         - node.total: '1'
+  - synopsis: This GET /_cat/health chapter returns application/smile and should pass.
+    path: /_cat/health
+    parameters:
+      format: cbor
+    method: GET
+    response:
+      status: 200
+      content_type: application/smile
```
Xtansia, line 54: As above regarding validating `content_type` as test is supposed to be for `smile` but format is set to `cbor`

Xtansia (changes requested):

tests/cat/health.yaml
dblock: Looks like it, https://github.com/opensearch-project/opensearch-api-specification/issues/392.

tools/tests/tester/fixtures/stories/passed.yaml
dblock: Updated to be `format: smile`.

Xtansia (approved):

PR #394, opened 2024-07-10
Validate response content types

### LINKED ISSUE #392: [BUG] Response content-type in spec tests is not validated
### What is the bug?

Coming from https://github.com/opensearch-project/opensearch-api-specification/pull/386#discussion_r1671365618, looks like `response/content-type` is not validated.

In tests.

```yaml
    response:
      status: 200
      content_type: x/z // doesn't matter what you put here
```

Fail if the response type is incorrect.

Xtansia:
### Description
Validate response content types

### Issues Resolved
Fixes #392

### DIFF
--- a/tests/cat/health.yaml
+++ b/tests/cat/health.yaml
@@ -76,7 +76,7 @@ chapters:
       format: cbor
     response:
       status: 200
-      content_type: application/yaml
+      content_type: application/cbor
       payload:
         - node.total: '1'
           status: yellow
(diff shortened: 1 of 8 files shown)


dblock (changes requested):
Looks good. Add a test fixture into invalid_data.yml (or a new fixture, might be worth reorganizing those files a little) that tests both possible new failures in integration tests.

dblock (approved):

PR #397, opened 2024-07-10
Update release description with CHANGELOG and upload an artifact with every build.

### LINKED ISSUE #380: [FEATURE] Release artifact permalink

Coming from https://github.com/opensearch-project/opensearch-java/pull/1065#discussion_r1668654405 we are trying to link to an artifact in the main pre-release in https://github.com/opensearch-project/opensearch-api-specification/releases. This repo makes/updates a pre-release with every commit, but doesn't actually make a release, so the pre-release is actually a moving target.

Change the release process to produce artifacts that have permalinks. This could be a product of a build with an artifact or a release that can be permalinked, but we likely don't want to make a release with every build. Eventually we will be making releases that match OpenSearch releases with full API coverage.

### What alternatives have you considered?

Do nothing, have projects download pre-release artifacts. Wait till we have real releases of API specification, or just start making releases.

dblock:
### Description

Update release description with CHANGELOG, and upload an artifact with every build.

Before:

Screenshot 2024-07-10 at 3 09 00 PM

After:

https://github.com/dblock/opensearch-api-specification/releases/tag/release-permalink-latest

Screenshot 2024-07-10 at 3 07 41 PM

Also uploads the build artifacts to the workflow run so it can be downloaded with a permalink if needed.

### Issues Resolved

Closes #380.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -28,13 +28,42 @@ jobs:
         shell: bash
         run: echo "name=${GITHUB_HEAD_REF:-${GITHUB_REF#refs/heads/}}" >> $GITHUB_OUTPUT
 
+      - name: Upload Artifacts
+        id: upload
+        uses: actions/upload-artifact@v4
+        with:
+          name: build
+          path: |
+            build/*
+
       - name: Release Specification to GitHub
+        id: release
         uses: marvinpinto/action-automatic-releases@v1.2.1
         with:
-          repo_token: "${{ secrets.GITHUB_TOKEN }}"
+          repo_token: ${{ secrets.GITHUB_TOKEN }}
           automatic_release_tag: ${{ steps.branch.outputs.name }}-latest
           prerelease: true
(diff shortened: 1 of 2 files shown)


Xtansia (approved):
I'm happy for this to be merged for now and to just re-visit it later, but I have some opinions:

I'm not super convinced on the value add of the build artifact compared to what's already associated with the release given it's not a true permalink and the only way to map the non-latest commit to an artifact is to query via the actions API.

I also thinks there's an opportunity to clean this workflow up as the automatic gh release action is now unmaintained and editing the release message in multiple steps seems error prone.

dblock:
> I'm not super convinced on the value add of the build artifact compared to what's already associated with the release given it's not a true permalink and the only way to map the non-latest commit to an artifact is to query via the actions API.

True. If nobody uses it we can delete it in the next iteration.
 
> I also thinks there's an opportunity to clean this workflow up as the automatic gh release action is now unmaintained and editing the release message in multiple steps seems error prone.

Yes indeed. Amazingly there's no action that exists that lets you create a release with the description right away. It might actually not be possible? ;)

PR #405, opened 2024-07-11
Fixed query DSL `match` that supports a field name and value

dblock:
### Description

Fixed query DSL match that supports a field name and value.
Added reindex tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -75,6 +75,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Fixed `/{index}/_dangling` that can return `nodes` and `cluster_name` ([#391](https://github.com/opensearch-project/opensearch-api-specification/pull/391))
 - Fixed `Metadata` schema ([#399](https://github.com/opensearch-project/opensearch-api-specification/pull/399))
 - Fixed `/_data_stream` health status and required fields ([#401](https://github.com/opensearch-project/opensearch-api-specification/pull/401))
+- Fixed query DSL `match` that supports a field name and value ([#405](https://github.com/opensearch-project/opensearch-api-specification/pull/405)) 
 
 ### Security
 
--- a/spec/schemas/_common.query_dsl.yaml
+++ b/spec/schemas/_common.query_dsl.yaml
@@ -69,16 +69,13 @@ components:
           description: |-
             Returns documents that match a provided text, number, date or boolean value.
             The provided text is analyzed before matching.
-          anyOf:
-            - type: object
-              additionalProperties:
-                $ref: '#/components/schemas/MatchQuery'
-              minProperties: 1
-              maxProperties: 1
-            - type: object
-              additionalProperties: true
-              minProperties: 1
-              maxProperties: 1
+          type: object
+          additionalProperties:
+            anyOf:
+              - $ref: '#/components/schemas/MatchQuery'
+              - true
+          minProperties: 1
+          maxProperties: 1
         match_all:
           $ref: '#/components/schemas/MatchAllQuery'
         match_bool_prefix:
--- a/spec/schemas/_common.query_dsl.yaml
+++ b/spec/schemas/_common.query_dsl.yaml
@@ -69,11 +69,16 @@ components:
           description: |-
             Returns documents that match a provided text, number, date or boolean value.
             The provided text is analyzed before matching.
-          type: object
-          additionalProperties:
-            $ref: '#/components/schemas/MatchQuery'
-          minProperties: 1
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -74,6 +74,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Fixed `_source` in `bulk` responses ([#375](https://github.com/opensearch-project/opensearch-api-specification/pull/375))
 - Fixed `/{index}/_dangling` that can return `nodes` and `cluster_name` ([#391](https://github.com/opensearch-project/opensearch-api-specification/pull/391))
 - Fixed `Metadata` schema ([#399](https://github.com/opensearch-project/opensearch-api-specification/pull/399))
+- Fixed query DSL `match` that supports a field name and value ([#405](https://github.com/opensearch-project/opensearch-api-specification/pull/405))
```
nhtruong, line 77: I usually pick a random location in the change logs listing instead of the last entry to avoid merging conflict :)

spec/schemas/_common.query_dsl.yaml
```diff
@@ -69,11 +69,16 @@ components:
...
+          anyOf:
+            - type: object
+              additionalProperties:
+                $ref: '#/components/schemas/MatchQuery'
+              minProperties: 1
+              maxProperties: 1
+            - type: object
+              additionalProperties: true
+              minProperties: 1
+              maxProperties: 1
```
nhtruong, line 81: doesn't the later (`additionalProperties: true`)  already encapsulate the former? or is it a placeholder for when we can accurately define another format for this?

tests/_core/reindex.yaml
```diff
@@ -0,0 +1,212 @@
...
+description: Test reindex.
+epilogues:
+  - path: /movies
+    method: DELETE
+    status: [200, 404]
+  - path: /films
+    method: DELETE
+    status: [200, 404]
+  - path: /videos
+    method: DELETE
```
nhtruong, line 12: we can shorten this with
```yaml
  - path: /movies,films,videos
    method: DELETE
    status: [200, 404]
```

tests/_core/reindex.yaml
```diff
@@ -0,0 +1,212 @@
...
+      payload:
+        title: Beauty and the Beast
+        year: 91
+    response:
+      status: 201
+  - synopsis: Refresh the index.
+    path: /{index}/_refresh
+    method: POST
+    parameters:
+      index: movies
```
nhtruong, line 33: The first 2 chapters should be part of `prologues` if they are not the operations to be tested in this story. According to the name, only `_reindex` operations are being tested in this story right?

tests/_core/reindex.yaml
```diff
@@ -0,0 +1,212 @@
...
+        processors:
+          - split:
+              field: title
+              separator: '\s+'
+              target_field: words
+          - script:
+              lang: painless
+              source: 'ctx.length = ctx.words.length'
+          - remove:
+              field: year
```
nhtruong, line 145: This should probably be part of the prologues too.

tests/_core/reindex.yaml
dblock: No, because then if 1 of these doesn't exist the entire request fails.

spec/schemas/_common.query_dsl.yaml
dblock: We have an `additionalProperties: true` and 

```
additionalProperties:
   type: object
```

This is quite different. The first one allows you to do `match: x`, and the second

```
match:
  x: y
```

nhtruong: I'm curious if the following is allowed in JSON schema.
```yaml
type: object
additionalProperties:
  anyOf:
     - true
     - $ref: '#/components/schemas/MatchQuery'
minProperties: 1
maxProperties: 1
```

nhtruong: Just tried it with the new JsonSchemaValidator:
```js
validator.validate_schema({ additionalProperties: { anyOf: [{type: 'string'}, { $ref: '#something'}]} })
```
It's allowed :D

nhtruong (approved):

spec/schemas/_common.query_dsl.yaml
dblock: Indeed! `is_ref` needed some tweaking, https://github.com/opensearch-project/opensearch-api-specification/pull/405/commits/922d1543a49243c87bd7058378602df5aebaf43d

nhtruong (approved):

PR #409, opened 2024-07-11
Run integration tests against OpenSearch 2.16.

dblock:
### Description

1. Run integration tests against multiple versions of OpenSearch (latest and next).
2. Actively wait for OpenSearch docker container to come up instead of a fixed 60s.
3. Display OpenSearch version tests are being run against.

Screenshot 2024-07-11 at 9 41 46 AM

### Issues Resolved

### DIFF
--- a/.github/workflows/test-spec.yml
+++ b/.github/workflows/test-spec.yml
@@ -2,25 +2,32 @@ name: Test Spec
 
 on:
   push:
-    branches: ['**']
     paths:
-      - 'package*.json'
+      - .github/workflows/test-spec.yml
+      - package*.json
       - tsconfig.json
-      - 'tools/src/tester/**'
-      - 'spec/**'
+      - tools/src/tester/**'
+      - spec/**
   pull_request:
-    branches: ['**']
     paths:
-      - 'package*.json'
+      - .github/workflows/test-spec.yml
+      - package*.json
       - tsconfig.json
-      - 'tools/src/tester/**'
-      - 'spec/**'
+      - tools/src/tester/**
+      - spec/**
 
 jobs:
   test-opensearch-spec:
+    strategy:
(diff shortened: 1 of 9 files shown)


.github/workflows/test-spec.yml
```diff
@@ -2,25 +2,32 @@ name: Test Spec
...
   push:
-    branches: ['**']
     paths:
-      - 'package*.json'
+      - .github/workflows/test-spec.yml
+      - package*.json
       - tsconfig.json
-      - 'tools/src/tester/**'
-      - 'spec/**'
+      - tools/src/tester/**'
```
nhtruong, line 9: the last `'` should be removed too

.github/workflows/test-spec.yml
```diff
@@ -2,25 +2,32 @@ name: Test Spec
...
+      - tools/src/tester/**
+      - spec/**
 
 jobs:
   test-opensearch-spec:
+    strategy:
+      matrix:
+        entry:
+          - {version: 2.15.0, hub: 'opensearchproject'}
+          - {version: 2.16.0, hub: 'opensearchstaging'}
```
nhtruong, line 25: This is great. My concern is whether we should test multiple versions of OS now when the test framework cant differentiate which features in the spec can be tested against which version of OS. Say if we add a new query param in 2.17 and update the spec accordingly, then all matrix entries before 2.17 will likely fail stories that test the affected endpoint.

Potential solution (in a different PR): Add `min_cluster_version` and `max_cluster_version` properties for test stories and the test framework will only test the story if the it's being tested against a valid cluster version.

dblock: I'm glad you asked, I got better in #410.

nhtruong (approved):

PR #412, opened 2024-07-12
Switch validating to use unevaluatedProperties

Xtansia:
### Description
Switch validating to use unevaluatedProperties and fix errors in specification


tools/src/_utils/index.ts
```diff
@@ -25,6 +31,33 @@ export function is_primitive_schema (schema: OpenAPIV3.SchemaObject): boolean {
     schema.type === 'string'
 }
 
+export function determine_possible_schema_types (doc: OpenAPIV3.Document, schema: MaybeRef<OpenAPIV3.SchemaObject>): SchemaObjectType[] {
```
dblock, line 34: This could use a test.

spec/namespaces/cluster.yaml
```diff
@@ -514,12 +514,10 @@ components:
             properties:
               persistent:
                 type: object
-                additionalProperties:
-                  type: object
+                additionalProperties: { }
```
dblock, line 517: For later, linter should catch `{}` vs. `{ }`.

dblock (approved):
Looks good, leaving some comments, feel free to add later/not.

PR #416, opened 2024-07-12
Removed one of the duplicate '/_nodes/{metric}' and '/_nodes/{node_id}' paths.

dblock:
### Description

Was trying to load the spec with the Ruby openapi3_parser and it failed with this error.

```
/Users/dblock/.rvm/gems/ruby-3.1.2/gems/openapi3_parser-0.9.2/lib/openapi3_parser/node_factory/map.rb:206:in `validate': Invalid data for #/paths: There are paths that conflict: '/_nodes/{metric}', '/_nodes/{node_id}' (Openapi3Parser::Error::InvalidData)
```

Now properly detecting dups.

### Issues Resolved

### DIFF
--- a/spec/namespaces/nodes.yaml
+++ b/spec/namespaces/nodes.yaml
@@ -542,7 +542,9 @@ components:
         items:
           anyOf:
             - $ref: '../schemas/_common.yaml#/components/schemas/NodeIds'
+              title: node_id
             - $ref: '../schemas/nodes.info.yaml#/components/schemas/Metric'
+              title: metric
       style: simple
     nodes.info::path.node_id:
       in: path
--- a/spec/namespaces/nodes.yaml
+++ b/spec/namespaces/nodes.yaml
@@ -253,7 +253,7 @@ paths:
       responses:
         '200':
           $ref: '#/components/responses/nodes.usage@200'
-  /_nodes/{node_id}:
+  /_nodes/{node_id_or_metric}:
     get:
       operationId: nodes.info.1
       x-operation-group: nodes.info
@@ -262,7 +262,7 @@ paths:
       externalDocs:
         url: https://opensearch.org/docs/latest/api-reference/nodes-apis/nodes-info/
       parameters:
-        - $ref: '#/components/parameters/nodes.info::path.node_id'
+        - $ref: '#/components/parameters/nodes.info::path.node_id_or_metric'
         - $ref: '#/components/parameters/nodes.info::query.flat_settings'
         - $ref: '#/components/parameters/nodes.info::query.timeout'
       responses:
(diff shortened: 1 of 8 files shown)


nhtruong:
This will cause a breaking change in the clients. The clients accepts both `metric` and `node_id` as params for `client.nodes.info()` function ([js](https://github.com/opensearch-project/opensearch-js/blob/main/api/api/nodes.js), [ruby](https://github.com/opensearch-project/opensearch-ruby/blob/main/lib/opensearch/api/actions/nodes/info.rb)) and uses the following logic to determine which path to use:
- both `metric` and `node_id` are provided: `/nodes/{node_id}/{metric}`
- only `node_id` is provided: `/nodes/{node_id}`
- only `metric` is provided: `/nodes/{metric}`
- neither is provided: `/nodes`
This is another quirk of OpenSearch API.
The server treats`/nodes/{node_id}` and  `/nodes/{metric}` as an identical endpoint. So this endpoint is actually `/nodes/{node_id_or_metric}`. But this will cause the clients to read `node_id_or_metric` as the 3rd path param. We can collapse those 2 endpoints into `/nodes/{node_id_or_metric}` but we will have to add a new `x-` extension to tell the generator how to handle this edge case.

tools/tests/linter/fixtures/folder_validators/namespaces/invalid_folder/dup_path_d.yaml
```diff
@@ -0,0 +1,3 @@
+paths:
+  '/nodes/{metric}': {}
+  '/nodes/{node_id}': {}
```
nhtruong, line 3: We should also account for when the paths appear in different namespace files.

dblock: https://github.com/opensearch-project/opensearch-api-specification/pull/416/commits/b3da76c27c337fce57973102e77e65eb8ed75635

dblock:
> I'm thinking `x-union-param: ['node_id', 'metric']` for the `node_id_or_metric` path param

I made the change by introducing `/_nodes/{node_id_or_metric}`.

I am not sure we're doing the right thing by introducing parameter hints. As written in this PR the parameter is `node_id_or_metric`, and it can now be an array of either a `node_id` or a `metric` (`anyOf`). Perhaps, the right way is for  clients to patch this behavior to be backwards compatible instead of the spec requiring a hint?

spec/namespaces/nodes.yaml
```diff
@@ -545,15 +530,21 @@ components:
...
+    nodes.info::path.node_id_or_metric:
       in: path
-      name: metric
-      description: Limits the information returned to the specific metrics. Supports a comma-separated list, such as http,ingest.
+      name: node_id_or_metric
+      description: |
+        Limits the information returned to a list of node IDs or specific metrics.
+        Supports a comma-separated list, such as node1,node2 or http,ingest.
       required: true
       schema:
```
Xtansia, line 540: NodeIds is already an array, so need to lift up the anyOf
```suggestion
      schema:
        anyOf:
          - title: node_id
            $ref: '../schemas/_common.yaml#/components/schemas/NodeIds'
          - title: metric
            type: array
            items:
              $ref: '../schemas/nodes.info.yaml#/components/schemas/Metric'
```

Xtansia (changes requested):

dblock:
> We should update https://github.com/opensearch-project/opensearch-api-specification/blob/main/CLIENT_GENERATOR_GUIDE.md to alert devs about this edge case as well.

Done, https://github.com/opensearch-project/opensearch-api-specification/pull/416/commits/d4f3a55c9fd8cab683faa1c30416e25936724a39.

nhtruong (approved):

spec/namespaces/nodes.yaml
```diff
@@ -545,15 +530,28 @@ components:
...
+            items:
+              $ref: '../schemas/nodes.info.yaml#/components/schemas/Metric'
         type: array
         items:
-          $ref: '../schemas/nodes.info.yaml#/components/schemas/Metric'
+          anyOf:
+            - $ref: '../schemas/_common.yaml#/components/schemas/NodeIds'
+              title: node_id
+            - $ref: '../schemas/nodes.info.yaml#/components/schemas/Metric'
+              title: metric
```
Xtansia, line 554: Need to remove these duplicated parts

Xtansia (changes requested):

spec/namespaces/nodes.yaml
dblock: 🤦

Xtansia (approved):

PR #418, opened 2024-07-15
Renamed files and operations for consistency.

dblock:
### Description

I am not sure it makes sense to rename operation IDs, but this is what it would look like for `dangling`, `remotestore` and `transform`. Currently the files and the APIs don't match the operation IDs, not sure whether that was intentional?


dblock:
Nevermind, this creates more problems.

@nhtruong LMK if you think any of it makes sense and I can run it down.

PR #419, opened 2024-07-15
set `value` required in constant_keyword field type

kkewwei:
### Description
set value required in constant_keyword field type

### Issues Resolved


dblock:
Looks like tests are failing.

Organization-wise I wonder where we should put this test. Is this a `_doc` or `mapping` thing? Maybe we want to create subfolders accordingly under `tests`? WDYT?

tests/mapping/constant_keyword.yaml
```diff
@@ -0,0 +1,37 @@
+$schema: ../../json_schemas/test_story.schema.yaml
+
+description: Test constant_keyword field type and retrieving a doc.
+
+prologues:
+  - path: /{index}
```
dblock, line 6: This index is not needed.

spec/schemas/_common.mapping.yaml
```diff
@@ -789,13 +789,14 @@ components:
         - type: object
           properties:
             value:
-              type: object
+              type: string
```
dblock, line 792: Is this always a string? Can the value not be an integer, an object, etc?

dblock (changes requested):

tests/mapping/constant_keyword.yaml
```diff
@@ -0,0 +1,33 @@
...
+    method: PUT
+    parameters:
```
dblock, line 8: Btw, in prologues you can do `path: /movies` and not use `parameters`. That's because we don't test schema here. Makes it a little shorter, NBD.

dblock (changes requested):
Looks like 2.0 test failed. Which version was this added in? You will need a `version: '>= ...'` in the test.

Also add a line to CHANGELOG. This is a fix for the constant_keyword mapping type.

spec/schemas/_common.mapping.yaml
kkewwei: I have added the test to test integer in https://github.com/opensearch-project/OpenSearch/pull/14651.

CHANGELOG.md
```diff
@@ -82,6 +82,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Fixed query DSL `match` that supports a field name and value ([#405](https://github.com/opensearch-project/opensearch-api-specification/pull/405)) 
 - Fixed `/_mapping` with `index` in query ([#385](https://github.com/opensearch-project/opensearch-api-specification/pull/385))
 - Fixed duplicate `/_nodes/{node_id}` path ([#416](https://github.com/opensearch-project/opensearch-api-specification/pull/416))
+- Fixed Set value required in constant_keyword field type [#419](https://github.com/opensearch-project/opensearch-api-specification/pull/419)
```
dblock, line 85: Sorry to be annoying, I promise, last one, could you please make it look like other lines? (lowercase set, quote `constant_keyword`, PR in ()). 

"Fixed required value for `constant_keyword` field type ([#419](https://github.com/opensearch-project/opensearch-api-specification/pull/419))"

dblock (changes requested):
One nit for CHANGELOG.

The 2.16 daily is failing to startup, I'll look into it.

https://github.com/opensearch-project/opensearch-api-specification/actions/runs/9967302688/job/27562098545

CHANGELOG.md
kkewwei: I apologize for my carelessness. I just copy, and don't know why the different. I have updated

dblock (approved):

dblock:
> @dblock, The source code has the bug( [opensearch-project/OpenSearch#14651](https://github.com/opensearch-project/OpenSearch/pull/14651)), is this caused by the bug?

Dailies don't have all plugins when they fail to build, looks like today security plugin didn't work, so the version on docker hub doesn't have security/SSL, and that's what's failing. I locked it to a specific revision in #431 and we can update it regularly.

dblock:
Merging since the 2.16 problem is unrelated. Thanks @kkewwei!

dblock:
@kkewwei Did I make a mistake merging this? This test is not supposed to work in 2.15 either because of https://github.com/opensearch-project/OpenSearch/pull/14651 and we have to wait for your fix to be merged there? I can revert.

dblock:
@kkewwei Reopen this when the bug is fixed with `>= 2.16` (will need an update of the first image that has the fix in https://github.com/opensearch-project/opensearch-api-specification/blob/main/.github/workflows/test-spec.yml#L29).

Issue #84, opened 2023-04-06
[PROPOSAL] Publish versioned API specs


dblock:
### What are you proposing?

Every version of OpenSearch has its own version of the API and therefore the API specs should reflect that.

### What users have asked for this feature?

In https://github.com/opensearch-project/opensearch-clients/issues/58 we plan to auto-generate clients. Therefore we need a way to both version API specs and a way to know the exact API spec for a given version of server.

### What problems are you trying to solve?

- Easily identify the exact spec for an exact version of OpenSearch.
- Ability to generate specs for flavors of deployments of OpenSearch, such as the Amazon managed service, which may have APIs added or removed.

### What is the developer experience going to be?

I see several options.

a) The specs tag when APIs were added/removed/changed, see #80, and publish a single version of specs with the options/tags.
b) The repo is branched in ways that follows OpenSearch and each branch publishes a version of the specs. 

### What is the user experience going to be?

I'd like to be able to say "where's the 2.7.0 version of OpenSearch specs" and consume it directly.

wbeckler:
I would propose that the API spec include version ranges as a field for a method. A given client should be able to speak with multiple servers, and by knowing which server version we're talking to, or by specifying it in the request, a single client with a multi-version spec could handle that situation.

reta:
> a) The specs tag when APIs were added/removed/changed, see https://github.com/opensearch-project/opensearch-api-specification/issues/80, and publish a single version of specs with the options/tags.

I kind of like this approach: continuously evolving the spec. I would like to suggest a combined approach, keep one spec for a release line (1.x, 2.x, 3.x). Why is that: within the release line, the spec should evolve in a backward compatible way but not between major releases. I have difficulties expressing the API in spec that was introduced in 1.3.7, deprecated in 2.7.0 and removed in 3.0.0 (we could only express the latter essentially). The only option with single spec I see is when we always evolve the spec in backward compatible way.

For cases where there is a need to have multiple clients, non issue in scope of same release line, but if the support of multiple major versions is needed, several clients could be generated easily (one for 2.x and one for 3.x fe).

What do you think?

wbeckler:
[...]

PR #424, opened 2024-07-16
Test against OpenSearch 1.3.17.

dblock:
### Description

This passes tests against OpenSearch 1.x.

Since 1.x is still supported I think we should also support it in the spec. Clients can choose to ignore anything that's `x-version-removed=2.0`. This gives us a way to evolve backwards incompatible changes in 3.0.

The other option is a 1.x branch and we can have backports. It's possibly easier, and so if you think it's better, then I can PR spec and tests with `>= 2.0` separately so that the diff is smaller, and make a 1.x branch.

WDYT @nhtruong @Xtansia ?

### Issues Resolved

Part of https://github.com/opensearch-project/opensearch-api-specification/issues/84.

### DIFF
--- a/spec/schemas/_core.get.yaml
+++ b/spec/schemas/_core.get.yaml
@@ -9,6 +9,8 @@ components:
     GetResult:
       type: object
       properties:
+        _type:
+          $ref: '_common.yaml#/components/schemas/Type'
         _index:
           $ref: '_common.yaml#/components/schemas/IndexName'
         fields:
--- a/.github/workflows/test-spec.yml
+++ b/.github/workflows/test-spec.yml
@@ -23,6 +23,7 @@ jobs:
     strategy:
       matrix:
         entry:
+          - {version: 1.3.17, admin_password: admin}
           - {version: 2.0.0, admin_password: admin}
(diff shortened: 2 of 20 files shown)


spec/schemas/_core.get.yaml
```diff
@@ -9,6 +9,9 @@ components:
     GetResult:
       type: object
       properties:
+        _type:
+          $ref: '_common.yaml#/components/schemas/Type'
+          x-version-removed: '2.0'
```
nhtruong, line 14: This is redundant since `_common.yaml#/components/schemas/Type` already has this property.

dblock: What's the right place for it? In `components/schemas/Type` or here?

nhtruong:
This is fine. The changes between major versions we see now are mainly adding and removing components so this works.

spec/schemas/_core.get.yaml
nhtruong: The entire concept of doc type was removed in 2.0 so putting it in `components/schemas/Type` makes the most sense

nhtruong (approved):

PR #428, opened 2024-07-16
Delete schema elements that don't match target OpenSearch version.

### LINKED ISSUE #425: [FEATURE] Remove properties with x-version-added/removed when checking schema

Coming from https://github.com/opensearch-project/opensearch-api-specification/pull/424 we are adding `x-version-removed` in various fields. However, these fields are still included when checking schema against a server version >= version removed.

When running tests against OpenSearch X, schema verification should delete fields that are x-version-removed <= X, and fields that are x-version-added > X.

dblock:
### Description

This change allows us to generates a merged spec that targets a specific version without all the unnecessary parts. It accomplishes a number of goals.

1. It removes fields from the spec in integration testing that have been `x-version-removed` before the target server version or `x-version-added` after, effectively testing that the values of these fields are correct.
2. It enables generating specs for a given version of OpenSearch without branching. We could publish a separate 1.x spec and a 2.x spec instead of 1 combined spec, removing the burden of having to check `x-version` fields  from consumers of the spec. This would allow us to maintain stable releases of clients that are only compatible with OpenSearch 1.x for example with only fixes, without adding any 2.x methods to them.

```
npm run merge -- --opensearch-version=1.3 --output build/v1.3.yaml
npm run merge -- --opensearch-version=2.0 --output build/v2.0.yaml
diff build/v1.3.yaml build/v2.0.yaml
```

This removes non-matching parameters.

```
132a134
>         - $ref: '#/components/parameters/indices.put_alias::query.cluster_manager_timeout'
```

and entire operations

```
>     cat.templates::query.cluster_manager_timeout:
>       name: cluster_manager_timeout
>       in: query
>       description: Operation timeout for connection to cluster-manager node.
>       schema:
>         $ref: '#/components/schemas/_common:Duration'
>       x-version-added: '2.0'
```

### Issues Resolved

Closes #425.


tools/src/merger/OpenApiMerger.ts
```diff
@@ -40,17 +45,27 @@ export default class OpenApiMerger {
     }
   }
 
-  merge (output_path: string = ''): OpenAPIV3.Document {
+  write_to(output_path: string): OpenApiMerger {
+    if (!this._merged) { throw "Call merge() first." }
```
nhtruong, line 49: Why doesn't `write_to()` just call `merge()` itself in this case?

tools/src/merger/OpenApiMerger.ts
```diff
@@ -40,17 +45,27 @@ export default class OpenApiMerger {
...
+    this.logger.info(`Writing ${output_path} ...`)
+    write_yaml(output_path, this._spec)
+    return this
+  }
+
+  merge (): OpenApiMerger {
+    if (this._merged) { throw "Spec already merged." }
```
nhtruong, line 56: Why doesn't merge just return in this case?

tools/src/merger/OpenApiMerger.ts
```diff
@@ -59,21 +74,83 @@ export default class OpenApiMerger {
...
+          removed.push(key)
+          // eslint-disable-next-line @typescript-eslint/no-dynamic-delete
+          delete obj[key]
+        } else {
+          removed = _.concat(removed, this.#remove_per_semver(obj[key]))
+        }
+      }
+    }
+
+    return removed
```
nhtruong, line 144: Isolation-of-concerns wise, I think this logic should be of its own class (for better readability and testing). The `OpenApiMerger`'s class' only duty is turning all the YAML files in the `spec` folder into a single OpenAPI yaml file without any modification. Then we have something like `ExtractSpecByVersion` class that uses the output of this  class to remove the irrelevant bits.

tools/src/merger/OpenApiMerger.ts
dblock: I got rid of `merge()` altogether, it just does it the first time when you call `spec()`.

tools/src/merger/OpenApiMerger.ts
dblock: I got rid of `merge()` altogether, it just does it the first time when you call `spec()`.

tools/src/merger/OpenApiMerger.ts
dblock: Extracted into a helper.

nhtruong: I meant this [entire code block](https://github.com/opensearch-project/opensearch-api-specification/pull/428/files#diff-707690e395baf72e22f4878dc5f5d802fc2a0a558ca4d56f9b49e67a8e3861fcR80-R126) should be its own class (say `ExtractSpecByVersion`) that consumes the output of the current `OpenApiMerger`. It makes the logic easier to follow and test as `OpenApiMerger` would be doing 2 jobs that intertwine with each other (merge all files in `./spec` and remove by `x-version...`).

Also, maybe I've missed it, but does this also remove items in `#components/schemas` that are no longer referenced after the removal of operation components?

dblock: I see. I think I disagree with that. For me, merging a spec targeting a different version is just a feature of a merger, not something that takes an existing merged spec to create another merged spec. LMK if you feel strongly about this and I'll extract it.

nhtruong: It's just a matter of opinion. It's easier to remove these components and the affected components (i.e. the unused items in `#components/schemas`) from a single-file spec than removing them from the individual spec file during merging and then figure out which other files are affected.

dblock: I ended up extracting this into a class as you suggested, you're right, it's cleaner.

nhtruong:
2 Things we should add to this PR:
- Tests for when an entire operation is removed (Then all of it's components in `#components` will also have to be removed)
- Logic go remove items in `#components/schemas` that are no longer used anywhere after the removal of other components.

The [tools/src/linter/SchemaRefsValidator.ts](https://github.com/opensearch-project/opensearch-api-specification/blob/main/tools/src/linter/SchemaRefsValidator.ts) might have something you need. Once you’ve removed all components that do not exist in your specified version of OS, the $refs pointing to those components (directly and indirectly) will become unresolvable.
1) The SchemaRefsValidator does this by traversing all operations in #path and their components (i.e. parameters, requestBodies…) whenever it sees a $ref, it adds that ref to the list of referenced refs then go to the definition of that $ref and keeps traversing (with an optimization step of marking which $ref it has traversed to avoid doing it twice and avoid infinite loops of self $ref) till it hits an unresolvable $ref.
2) It also traverses #components to build a list of available_refs from the names of the components. It then compares the diff of referenced_refs and available_refs to look for unused components.

You can modify (2) to remove components not existing in your version of OS. Note that (2) will find all items in `#components` that are not used in `#paths`, not just those are not directly referenced. So, you can use that to prune all parameters, requestBodies, responses, and schemas that should be removed if an operation is removed as well.

dblock:
@nhtruong I think I addressed all the feedback. The updated implementation will delete all refs that don't match semver, then purge all unused refs and finally purge all objects that reference all refs that were removed.

tools/src/merger/OpenApiVersionExtractor.ts
```diff
@@ -0,0 +1,104 @@
...
+  #remove_keys_not_matching_semver(): void {
+    if (this._target_version === undefined) return
+    delete_matching_keys(this._spec, this.#exclude_per_semver.bind(this))
+  }
+
+  #remove_unused(): void {
+    if (this._spec === undefined) return
+
+    // remove anything that's not referenced
+    var references: string[] = find_refs(this._spec)
```
nhtruong, line 82: `references`, in this case, will also include all indirectly referenced components whose grandparents have been removed (since `find_refs` will traverse `_spec.components` as well). That is, say `#/components/parameters/X` has been removed. but `X` is an array whose `items` is `$ref: #/components/schemas/Y` but `Y` is `anyOf: [A,B]`. In this case, after X is removed, Y will be removed by this function as a result, but A and B will remain.

On the other hand, `find_refs(this.spec.paths)` will exclude all `#components/schemas/...` references being used by other `#/components/<non-schemas>/...` 😅. That's not good either.

I'm afraid this strategy won't work. You'll have to follow (1) described [here](https://github.com/opensearch-project/opensearch-api-specification/pull/428#issuecomment-2243844698) to find all `$ref` actually being used by the remaining parts of `#paths` (directly or indirectly).

tools/src/merger/OpenApiVersionExtractor.ts
```diff
@@ -0,0 +1,104 @@
...
+    this.#remove_unused()
+  }
+
+  #exclude_per_semver(obj: any): boolean {
+    if (this._target_version === undefined) return false
+
+    const x_version_added = semver.coerce(obj['x-version-added'] as string)
+    const x_version_removed = semver.coerce(obj['x-version-removed'] as string)
+
+    if (x_version_added && !semver.satisfies(this._target_version, `>=${x_version_added?.toString()}`)) {
```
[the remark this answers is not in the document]
dblock: Unfortunately not.

```
 55:73  error  Invalid type "SemVer" of template literal expression  @typescript-eslint/restrict-template-expressions
  57:81  error  Invalid type "SemVer" of template literal expression  @typescript-eslint/restrict-template-expressions
```

tools/src/helpers.ts
```diff
@@ -65,6 +64,33 @@ export function sort_array_by_keys (values: any[], priorities: string[] = []): s
...
+  if (obj !== undefined && obj !== null && obj.$ref !== undefined) {
+    results.push(obj.$ref as string)
+  }
+
+  if (_.isObject(obj)) {
+    results = _.concat(results, _.flatMap(obj, (v, _k) => find_refs(v as Record<string, any>)))
+  }
+
+  return _.uniq(results)
+}
```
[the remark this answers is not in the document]
dblock: I dislike this because we're passing the accumulator as an argument which makes the function semantically weird. But ok.

tools/src/merger/OpenApiVersionExtractor.ts
dblock: https://github.com/opensearch-project/opensearch-api-specification/pull/428/commits/ed18c4d145a39c8a3c3cf2894dcfa66175102df8 will traverse refs recursively.

dblock: resolved

dblock: > On the other hand, find_refs(this.spec.paths) will exclude all #components/schemas/... references being used by other #/components/<non-schemas>/... 😅. That's not good either.

This is now handled. If a component is referenced from within `paths`, all its refs are now safe.

nhtruong (approved):

PR #429, opened 2024-07-17
Add  is_hidden parameter for create or update alias API

gaobinlong:
### Description

In this [PR](https://github.com/opensearch-project/OpenSearch/pull/14719), we fix the bug of create or update alias API doesn't throw exception for unsupported parameters, and also add the missing request body parameter `is_hidden`, so we need to update the api specification.


CHANGELOG.md
```diff
@@ -46,6 +46,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Added missing fields to `/_nodes/stats` ([#415](https://github.com/opensearch-project/opensearch-api-specification/pull/415))
 - Added missing metrics options to `/_nodes/stats` ([#422](https://github.com/opensearch-project/opensearch-api-specification/pull/422))
 - Added tests against OpenSearch 1.3 ([#424](https://github.com/opensearch-project/opensearch-api-specification/pull/424))
+- Add `is_hidden` parameter for create or update alias API
```
dblock, line 49: Make it look like the rest of the lines.

"Added `is_hidden` to `/{index}/_alias/` ([#...](...))"

spec/namespaces/indices.yaml
```diff
@@ -1860,6 +1860,11 @@ components:
                 $ref: '../schemas/_common.yaml#/components/schemas/Routing'
               search_routing:
                 $ref: '../schemas/_common.yaml#/components/schemas/Routing'
+              is_hidden:
+                x-version-added: '2.16'
+                description: |-
+                  If `true`, the alias will be hidden, defaults to `false`.
+                type: boolean
```
dblock, line 1867: Add `default: false`, and no need to say "defaults to ..." in the text. We're going to clean these up at some point.

tests/indices/create_or_update_alias.yaml
```diff
@@ -0,0 +1,42 @@
+$schema: ../../json_schemas/test_story.schema.yaml
```
dblock, line 1: We already have `alias.yaml` that has a lot of similar tests. Either add one more test for `is_hidden` there, or create a folder called `tests/indices/alias/is_hidden.yaml` and include only functionality related to this new is_hidden into it. I think the former seems simpler.

You also need to add `version: '>= 2.16'` into the test so it doesn't run with older OpenSearch (I expect it to fail).

dblock (dismissed):
Thanks! See below.

So the tests for <= 2.15 pass because we weren't checking parameters? Do update them to have the correct `version` if there are fields used in tests against a version that doesn't support them, so we don't imply they can work.

gaobinlong:
Thanks @dblock , please help to take a look at the new change.

Xtansia (approved):

Issue #366, opened 2024-06-27
Replace `{ "type": "string", "nullable": true }` with `{ "type": ["string", "null"] }`


nhtruong:
We have many schemas describing a datatype that is either string or null with `{ "type": "string", "nullable": true }`. However the correct way to describe this datatype is `{ "type": ["string", "null"] }`. Moreover, Using `anyOf` on `{ "type": ["string", "null"] }` and `{ "type": "string" }` is redundant because the former already encapsulates the latter.

Spawning off of this [comment](https://github.com/opensearch-project/opensearch-api-specification/pull/363#discussion_r1655534588)

PR #436, opened 2024-07-18
spec: replace nullable value with null type

Jakob3xD:
### Description
- replace all nullable values with type null
- replace anyOf and oneOf with type string, null 

### Issues Resolved
Closes https://github.com/opensearch-project/opensearch-api-specification/issues/366

### DIFF
--- a/spec/schemas/_common.aggregations.yaml
+++ b/spec/schemas/_common.aggregations.yaml
@@ -113,21 +113,14 @@ components:
     KeyedPercentiles:
       type: object
       additionalProperties:
-        oneOf:
-          - type: string
-          - type: number
-          - nullable: true
-            type: string
+        type: ['null', 'number', 'string']
     ArrayPercentilesItem:
       type: object
       properties:
         key:
           type: string
         value:
-          oneOf:
-            - type: number
-            - nullable: true
-              type: string
+          type: ['null', 'number', 'string']
         value_as_string:
           type: string
       required:
@@ -162,10 +155,7 @@ components:
               description: |-
                 The metric value. A missing value generally means that there was no data to aggregate,
                 unless specified otherwise.
-              oneOf:
-                - type: number
-                - nullable: true
-                  type: string
+              type: ['null', 'number', 'string']
             value_as_string:
               type: string
           required:
@@ -226,20 +216,11 @@ components:
             count:
               type: number
             min:
-              oneOf:
-                - type: number
-                - nullable: true
-                  type: string
+              type: ['null', 'number', 'string']
             max:
-              oneOf:
-                - type: number
-                - nullable: true
-                  type: string
+              type: ['null', 'number', 'string']
             avg:
-              oneOf:
-                - type: number
-                - nullable: true
-                  type: string
+              type: ['null', 'number', 'string']
             sum:
               type: number
             min_as_string:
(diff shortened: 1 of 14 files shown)


Jakob3xD:
I am wondering. Do we need this:
```
oneOf:
    - type: number
    - type:
        - 'null'
        - string
``` 
or would following also work:
```
oneOf:
    - type:
        - 'null'
        - number
```

dblock:
> I am wondering. Do we need this:
> 
> ```
> oneOf:
>     - type: number
>     - type:
>         - 'null'
>         - string
> ```
> 
> or would following also work:
> 
> ```
> oneOf:
>     - type:
>         - 'null'
>         - number
> ```

I think `oneOf` here is not required.

dblock:
The quoted `'null'` feels a little awkward, does it get treaded differently from a `null` (without quotes)?

spec/schemas/_common.aggregations.yaml
```diff
@@ -114,10 +114,10 @@ components:
       type: object
       additionalProperties:
         oneOf:
-          - type: string
           - type: number
-          - nullable: true
-            type: string
+          - type:
+              - 'null'
+              - string
```
dblock, line 120: What does this actually mean?

```
oneOf:
  - type: number
  - type:
     - 'null'
     - string
```

Is it not the same as 

```
type:
  - oneOf: 
    - number
    - string
    - 'null'
```

or even

```
type: [number, string, 'null']
```

?

Jakob3xD:
> The quoted `'null'` feels a little awkward, does it get treaded differently from a `null` (without quotes)?

Yes, otherwise it get interpreted by yaml.

spec/schemas/_common.aggregations.yaml
Jakob3xD: If my research is correct then both following definitions are valid.
```
 oneOf:
    - type: 'null'
    - type: number
    - type: string
```
```
type:
    - 'null'
    - number
    - string
``` 

According to https://blog.stoplight.io/difference-between-open-v2-v3-v31
> With OpenAPI 3.1, the specification now supports type as an array

Jakob3xD: Should I change all of those to the second definition with type as an array?

dblock:
> > The quoted `'null'` feels a little awkward, does it get treaded differently from a `null` (without quotes)?
> 
> Yes, otherwise it get interpreted by yaml.

So, can we introduce a `NullableString` so that we don't accidentally confuse the two?

spec/schemas/_common.aggregations.yaml
dblock: I don't think that's necessary, but collapsing the nested structure probably is.

nhtruong:
```yaml
type:
    - 'null'
    - number
    - string
```
Is the best option IMO.

dblock: I think it would be clearer if we created a type for `NullableString` because `null` can be ambiguous if you're not paying attention. I'm ok with what @nhtruong proposes though, @Jakob3xD your call.

Jakob3xD: `null` without quotes is not valid and will fail the validation test.
https://github.com/Jakob3xD/opensearch-api-specification/actions/runs/10003423255/job/27650310114

I'll do the suggested type array.

Jakob3xD: When using the type array we need to set allowUnionTypes for Ajv but I am not sure where I can configure it.

dblock (approved):

PR #440, opened 2024-07-19
Add API spec for Search Response Processors

### LINKED ISSUE #437: [BUG] Specs for Search Pipeline SearchResponseProcessors are missing
### What is the bug?

Coming from  https://github.com/opensearch-project/OpenSearch/issues/14800#issuecomment-2237118907

I need to add specs/test for two new search response processors.  However, non of the existing processors exist to serve as an example. [A search for `rerank`](https://github.com/search?q=repo%3Aopensearch-project%2Fopensearch-api-specification%20rerank&type=code) has no code results.  Similarly [for `rename_field`](https://github.com/search?q=repo%3Aopensearch-project%2Fopensearch-api-specification%20rename_field&type=code).   

It does look like these [used to be part of the smithy spec.](https://github.com/Tokesh/opensearch-api-specification/blob/1a901cdfd2db877de93529c386fe4c9c924ac7f7/model/search_pipeline/search_pipeline.smithy#L82-L89) but that code has been removed.

It looks like the request processors are included in https://github.com/opensearch-project/opensearch-api-specification/blob/main/spec/schemas/search_pipeline._common.yaml but I don't see any of the response types there 
   - [this](https://github.com/opensearch-project/opensearch-api-specification/blob/0453dbe35080bf31eebac1d75f8022667d019e9f/spec/schemas/search_pipeline._common.yaml#L26) may be a typo

### Do you have any additional context?

The PRs for the two new processor types are targeted for 2.16.0 release.

I don't have the bandwidth to manually recreate all the response processors prior to 2.16.0 code freeze, and I'm hopeful automation already exists to translate them from the old Smithy specs.

dbwiddis:
### Description

- Adds the missing Search Response Processors from Smithy spec migration
- Adds new processors for Sort and Split
  - https://github.com/opensearch-project/OpenSearch/pull/14785
  - https://github.com/opensearch-project/OpenSearch/pull/14800

### Issues Resolved

Fixes #437

### DIFF
--- a/tests/_core/search/pipeline.yaml
+++ b/tests/_core/search/pipeline.yaml
@@ -17,13 +17,13 @@ chapters:
       payload:
         response_processors:
           - sort:
-              field: 'message'
-              order: 'asc'
-              target_field: 'sorted_message'
+              field: message
+              order: asc
+              target_field: sorted_message
           - split:
-              field: 'message'
+              field: message
               separator: ', '
-              target_field: 'split_message'
+              target_field: split_message
     response:
       status: 200
       payload:
--- a/tests/_core/search/pipeline.yaml
+++ b/tests/_core/search/pipeline.yaml
@@ -1,4 +1,4 @@
-$schema: ../../json_schemas/test_story.schema.yaml
+$schema: ../../../json_schemas/test_story.schema.yaml
 
 description: |
   Test the creation of a search pipeline with sort and split response processors.
@@ -17,13 +17,13 @@ chapters:
       payload:
         response_processors:
           - sort:
-              field: "message"
-              order: "asc"
-              target_field: "sorted_message"
+              field: 'message'
+              order: 'asc'
+              target_field: 'sorted_message'
           - split:
-              field: "message"
-              separator: ", "
-              target_field: "split_message"
+              field: 'message'
+              separator: ', '
+              target_field: 'split_message'
     response:
       status: 200
       payload:
--- a/tests/_core/search/pipeline.yaml
+++ b/tests/_core/search/pipeline.yaml
@@ -1,0 +1,37 @@
+$schema: ../../json_schemas/test_story.schema.yaml
+
+description: |
+  Test the creation of a search pipeline with sort and split response processors.
+epilogues:
+  - path: /_search/pipeline/sorting_pipeline
+    method: DELETE
+    status: [200, 404]
+version: '>= 2.16'
+chapters:
+  - synopsis: Create search pipeline for sorting and splitting.
+    path: /_search/pipeline/{id}
+    method: PUT
+    parameters:
+      id: sorting_pipeline
+    request_body:
+      payload:
+        response_processors:
+          - sort:
+              field: "message"
+              order: "asc"
+              target_field: "sorted_message"
+          - split:
+              field: "message"
+              separator: ", "
+              target_field: "split_message"
+    response:
+      status: 200
+      payload:
+        acknowledged: true
+  - synopsis: Query created pipeline.
+    path: /_search/pipeline/{id}
+    method: GET
+    parameters:
+      id: sorting_pipeline
+    response:
+      status: 200
--- a/.cspell
+++ b/.cspell
@@ -30,6 +30,8 @@ OSCPU
 Oversample
 Rebalance
 Reindex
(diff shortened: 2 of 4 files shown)


tests/_core/search/pipeline.yaml
```diff
@@ -0,0 +1,37 @@
+$schema: ../../../json_schemas/test_story.schema.yaml
+
+description: |
+  Test the creation of a search pipeline with sort and split response processors.
+epilogues:
+  - path: /_search/pipeline/sorting_pipeline
+    method: DELETE
+    status: [200, 404]
+version: '>= 2.16'
```
dblock, line 9: The CI failed because we lock the build to a specific version on dockerhub [here](https://github.com/opensearch-project/opensearch-api-specification/blob/main/.github/workflows/test-spec.yml#L29) - update it to the latest one that also has the new pipeline added.

dblock (changes requested):
Looks great! 

Update CI to pass.

Maybe we can have coverage of all the other processors (a folder with `tests/_core/search/pipeline/response_processors.yaml`, etc.)? Eventually they will be used as samples for the documentation.

tests/_core/search/pipeline.yaml
dbwiddis: > The CI failed because we lock the build to a specific version on dockerhub [here](https://github.com/opensearch-project/opensearch-api-specification/blob/main/.github/workflows/test-spec.yml#L29) - update it to the latest one that also has the new pipeline added.

The PR adding this feature has not been merged yet.  What is the usual pattern for writing a test for unmerged code, since submitting this API PR is a prerequisite to merging the PR? 🐔 🥚

dbwiddis:
> Maybe we can have coverage of all the other processors (a folder with `tests/_core/search/pipeline/response_processors.yaml`, etc.)? 

I'm really trying to timebox how much I'm doing on this. I'll follow up with another PR if necessary but I really just need to get some 2.16.0 features in.

> Eventually they will be used as samples for the documentation.

I had that thought but then it just seemed like integration testing and this spec didn't seem the right place to do that.  Isn't the goal here just to make sure the REST request works?

I wrote detailed samples [here](https://github.com/opensearch-project/documentation-website/pull/7767) and it seems having a link in the API docs to one authoritative source is better than duplicating it?

tests/_core/search/pipeline.yaml
dblock: I think you have two choices.

1. Wait till the PR is merged and there's a build with it.
2. Split this PR in 2, one for existing features and one for the new one, and (1) for the new one.

dblock:
> I had that thought but then it just seemed like integration testing and this spec didn't seem the right place to do that. Isn't the goal here just to make sure the REST request works?

The goal here is to make sure that the schema matches between requests and responses. When you write integration tests here the OpenAPI spec is validated, not that the server works.

dblock:
> I wrote detailed samples [here](https://github.com/opensearch-project/documentation-website/pull/7767) and it seems having a link in the API docs to one authoritative source is better than duplicating it?

The problem with the documentation as written is that there's no guarantee that it's correct, so our [documentation contains a ton of bugs and typos](https://code.dblock.org/2024/07/12/fixing-bugs-in-opensearch-api-documentation-using-api-spec-tests.html). With running samples in this repo we will be able to export both the reference documentation (https://github.com/opensearch-project/documentation-website/issues/7700) and working samples in 8 different programming languages.

tests/_core/search/pipeline.yaml
dbwiddis: I had already assumed option 1 was the expectation.

dblock: Thanks @dbwiddis, looking forward to getting all this closed out!

dbwiddis:
> The spec should be that source.

I like this idea (I *really* like it) and have been trying to figure out the best way on Flow Framework to "integ test" our templates, which are "configuration as code".  I think a test framework like you have here is a great solution and I'll be writing an issue summarizing it shortly.

In the meantime to keep tracking this ask, feel free to open a new issue to build test/demo setup for Search Pipelines and assign it to me.  I may not get to it immediately but it'll be a fun weekend project sometime in the next month.

Issue #441, opened 2024-07-19
[FEATURE] _nodes/settings misses the plugins field


Jakob3xD:
### Is your feature request related to a problem?
When doing the spec test against the Opensearch started by the opensearch-go project,
I noticed the test complaining about a missing field `plugins`.

`curl https://localhost:9200/_nodes/settings -k -u admin`
```json
{
  "_nodes": {
    "total": 1,
    "successful": 1,
    "failed": 0
  },
  "cluster_name": "docker-cluster",
  "nodes": {
    "6fYrI3t8QWeUjBzYb2t05w": {
      "name": "60bf0b26a1cf",
      "transport_address": "172.19.0.2:9300",
      "host": "172.19.0.2",
      "ip": "172.19.0.2",
      "version": "2.15.0",
      "build_type": "tar",
      "build_hash": "61dbcd0795c9bfe9b81e5762175414bc38bbcadf",
      "roles": [
        "cluster_manager",
        "data",
        "ingest",
        "remote_cluster_client"
      ],
      "attributes": {
        "shard_indexing_pressure_enabled": "true"
      },
      "settings": {
        "cluster": {
          "name": "docker-cluster"
        },
        "node": {
          "attr": {
            "shard_indexing_pressure_enabled": "true"
          },
          "max_local_storage_nodes": "3",
          "name": "60bf0b26a1cf"
        },
        "path": {
          "logs": "/usr/share/opensearch/logs",
          "home": "/usr/share/opensearch",
          "repo": [
            "/usr/share/opensearch/mnt"
          ]
        },
        "discovery": {
          "type": "single-node"
        },
        "plugins": {
          "index_state_management": {
            "job_interval": "1"
          }
        },
        "client": {
          "type": "node"
        },
        "http": {
          "compression": "false",
          "type": "org.opensearch.security.http.SecurityHttpServerTransport",
          "type.default": "netty4"
        },
        "index": {
          "store": {
            "hybrid": {
              "mmap": {
                "extensions": [
                  "nvd",
                  "dvd",
                  "tim",
                  "tip",
                  "dim",
                  "kdd",
                  "kdi",
                  "cfs",
                  "doc",
                  "vec",
                  "vex"
                ]
              }
            }
          }
        },
        "bootstrap": {
          "memory_lock": "true"
        },
        "transport": {
          "type": "org.opensearch.security.ssl.http.netty.SecuritySSLNettyTransport",
          "type.default": "netty4"
        },
        "network": {
[...]

PR #442, opened 2024-07-19
spec: add plugins to NodeInfoSettings

Jakob3xD:
### Description
Add plugins to NodeInfoSettings

### Issues Resolved
Closes https://github.com/opensearch-project/opensearch-api-specification/issues/441

### DIFF
--- a/.github/opensearch-cluster/docker-compose.yml
+++ b/.github/opensearch-cluster/docker-compose.yml
@@ -10,3 +10,4 @@ services:
       - 'OPENSEARCH_INITIAL_ADMIN_PASSWORD=${OPENSEARCH_PASSWORD:-myStrongPassword123!}'
       - discovery.type=single-node
       - path.repo=/tmp/opensearch/repo
+      - plugins.index_state_management.job_interval=1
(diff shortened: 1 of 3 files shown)


dblock (changes requested):
Thanks! Care to add a test that has a response with this field?

Jakob3xD:
> Thanks! Care to add a test that has a response with this field?

I am not quite sure how to do this as the response contains as map with a random node id.

But I added the option to the docker env that is responsible for the plugins property to show up.

dblock:
> I am not quite sure how to do this as the response contains as map with a random node id.

You don't actually need to verify the content. As long as the response has the missing field it will validate it against the spec. Try without the fix and it should fail, then with the fix it will pass.

Jakob3xD:
> You don't actually need to verify the content. As long as the response has the missing field it will validate it against the spec. Try without the fix and it should fail - I imagine with the change to the docker options it's failing now? With the fix it will pass.

Yup, fails without the plugins.
https://github.com/Jakob3xD/opensearch-api-specification/actions/runs/10036637738/job/27734769941

dblock (approved):

dblock:
Merged, thank you!

dblock:
@Jakob3xD why did we add `plugins.index_state_management.job_interval=1` here?

Jakob3xD:
> @Jakob3xD why did we add `plugins.index_state_management.job_interval=1` here?

The setting gets returned on node info calls. With this it gets returned by the api and we can test against it. Otherwise it would be omitted.

dblock:
Can it be set via an API like [here](https://github.com/opensearch-project/opensearch-api-specification/blob/main/tests/sql/explain.yaml#L15)? I am thinking we need some way to set and unset these and other settings across tests.

dblock:
I tried to do this in https://github.com/opensearch-project/opensearch-api-specification/pull/471 in which I added a test that ensures the response contains the `job_interval` value. You can set cluster settings, but it has no effect on the response from nodes unfortunately. There's no way to set a node setting via API.

PR #444, opened 2024-07-21
Add Persian stem token filter

ebraminio:
### Description
Add Persian stem token filter API documentation

### Issues Resolved
Related to https://github.com/opensearch-project/OpenSearch/pull/14847

This is based on PorterStemTokenFilter


dblock (changes requested):
Thanks!

Please write a test for it and add to CHANGELOG.

ebraminio:
@dblock Added the changelog 😊 I'm yet figuring out how to add a test as there is not any test for any other filter as far as I can see

dblock:
> @dblock Added the changelog 😊 I'm yet figuring out how to add a test as there is not any test for any other filter as far as I can see

Yes, appreciate if you could write some for other filters as well, start with something simple and work from there. We have a testing guide: https://github.com/opensearch-project/opensearch-api-specification/blob/main/TESTING_GUIDE.md, too.

Since this is in the next version you will have to bump the docker hub ref in CI as well, https://github.com/opensearch-project/opensearch-api-specification/blob/main/.github/workflows/test-spec.yml#L29.

dblock:
Finishing this in https://github.com/opensearch-project/opensearch-api-specification/pull/592.

PR #454, opened 2024-07-30
Add experimental features flag option for tests

neetikasinghal:
### Description
As of now, there is no way in the tests to turn on the experimental feature flags for the tests, this pr aims to address it.

### Issues Resolved
Related prs: https://github.com/opensearch-project/opensearch-api-specification/pull/368

### DIFF
--- a/.github/opensearch-cluster/docker-compose.yml
+++ b/.github/opensearch-cluster/docker-compose.yml
@@ -8,6 +8,7 @@ services:
       - '9600:9600'
     environment:
       - 'OPENSEARCH_INITIAL_ADMIN_PASSWORD=${OPENSEARCH_PASSWORD:-myStrongPassword123!}'
+      - OPENSEARCH_JAVA_OPTS=${OPENSEARCH_JAVA_OPTS}
       - discovery.type=single-node
       - path.repo=/tmp/opensearch/repo
       - plugins.index_state_management.job_interval=1
--- a/.github/workflows/test-spec.yml
+++ b/.github/workflows/test-spec.yml
@@ -26,7 +26,9 @@ jobs:
           - {version: 1.3.17, admin_password: admin}
           - {version: 2.0.0, admin_password: admin}
           - {version: 2.15.0}
-          - {version: 2.16.0, hub: opensearchstaging, ref: '@sha256:50fbfe3b95c41e92a113ada3e80513ba4524dfc8a25dc6aaeff2bbe1e1145d5f'}
+          - version: 2.16.0
+            hub: opensearchstaging
+            ref: '@sha256:50fbfe3b95c41e92a113ada3e80513ba4524dfc8a25dc6aaeff2bbe1e1145d5f'
 
     name: test-opensearch-spec (version=${{ matrix.entry.version }}, hub=${{ matrix.entry.hub || 'opensearchproject' }})
     runs-on: ubuntu-latest
@@ -36,6 +38,7 @@ jobs:
       OPENSEARCH_DOCKER_REF: ${{ matrix.entry.ref }}
       OPENSEARCH_VERSION: ${{ matrix.entry.version }}
       OPENSEARCH_PASSWORD: ${{ matrix.entry.admin_password || 'myStrongPassword123!' }}
+      OPENSEARCH_JAVA_OPTS: ${{ matrix.entry.opts }}
 
     steps:
       - name: Checkout Repo
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Added test coverage ([#443](https://github.com/opensearch-project/opensearch-api-specification/pull/443))
 - Added `--opensearch-version` to `merger` that excludes schema elements per semver ([#428](https://github.com/opensearch-project/opensearch-api-specification/pull/428))
 - Added `retry` to `tester` to support asynchronous tasks ([453](https://github.com/opensearch-project/opensearch-api-specification/pull/453))
+- Added passing OPENSEARCH_JAVA_OPTS into the docker container used for tests ([#454](https://github.com/opensearch-project/opensearch-api-specification/pull/454))
 
 ### Changed
 


.github/workflows/test-spec.yml
```diff
@@ -26,7 +26,10 @@ jobs:
           - {version: 1.3.17, admin_password: admin}
           - {version: 2.0.0, admin_password: admin}
           - {version: 2.15.0}
-          - {version: 2.16.0, hub: opensearchstaging, ref: '@sha256:50fbfe3b95c41e92a113ada3e80513ba4524dfc8a25dc6aaeff2bbe1e1145d5f'}
+          - version: 2.16.0
+            hub: opensearchstaging
+            ref: '@sha256:50fbfe3b95c41e92a113ada3e80513ba4524dfc8a25dc6aaeff2bbe1e1145d5f'
+            features: '-Dopensearch.experimental.feature.tiered_remote_index.enabled=true'
```
dblock, line 32: This feels like just a parameter, either make this `opts` (for `OPENSEARCH_JAVA_OPTS`) or (this might be harder but nicer)

```yaml
experimental_features:
  - tiered_remote_index: true
```

neetikasinghal: updated to opts

.github/opensearch-cluster/docker-compose.yml
```diff
@@ -8,6 +8,7 @@ services:
       - '9600:9600'
     environment:
       - 'OPENSEARCH_INITIAL_ADMIN_PASSWORD=${OPENSEARCH_PASSWORD:-myStrongPassword123!}'
+      - 'OPENSEARCH_JAVA_OPTS=${OPENSEARCH_EXPERIMENTAL_FEATURES}'
```
dblock, line 11: The list of opts is not a list of experimental features, so this should be `OPENSEARCH_JAVA_OPTS=${OPENSEARCH_JAVA_OPTS}`.

.github/workflows/test-spec.yml
```diff
@@ -36,6 +39,7 @@ jobs:
       OPENSEARCH_DOCKER_REF: ${{ matrix.entry.ref }}
       OPENSEARCH_VERSION: ${{ matrix.entry.version }}
       OPENSEARCH_PASSWORD: ${{ matrix.entry.admin_password || 'myStrongPassword123!' }}
+      OPENSEARCH_EXPERIMENTAL_FEATURES: ${{ matrix.entry.opts || 'null' }}
```
dblock, line 42: This should be `OPENSEARCH_JAVA_OPTS` because it's not just a list of experimental features. 

Does it default to a string "null"? That's not the same thing as undefined. IMO it should read this:

```
OPENSEARCH_JAVA_OPTS: ${{ matrix.entry.opts }}
```

CHANGELOG.md
```diff
@@ -55,6 +55,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Added `plugins` to NodeInfoSettings ([#442](https://github.com/opensearch-project/opensearch-api-specification/pull/442)) 
 - Added test coverage ([#443](https://github.com/opensearch-project/opensearch-api-specification/pull/443))
 - Added `--opensearch-version` to `merger` that excludes schema elements per semver ([#428](https://github.com/opensearch-project/opensearch-api-specification/pull/428))
+- Added experimental features flag option for tests ([#454](https://github.com/opensearch-project/opensearch-api-specification/pull/454))
```
dblock, line 58: "Added passing OPENSEARCH_JAVA_OPTS into the docker container used for tests"

dblock (commented):
The code passes generic `OPENSEARCH_JAVA_OPTS` but claims this is `OPENSEARCH_EXPERIMENTAL_FEATURES` which it's not, it can be any opts. Get rid of mentions of "experimental features".

An alternate and likely a nicer implementation could be to actually pass a list of experimental features.

In test-spec.yml.

```yaml
- version: 2.16.0
   features:
     - tiered_remote_index: true
```

To convert this into `OPENSEARCH_JAVA_OPTS` you could use [join](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/expressions#join).

.github/workflows/test-spec.yml
neetikasinghal: will update this

.github/opensearch-cluster/docker-compose.yml
neetikasinghal: updated

neetikasinghal:
> The code passes generic `OPENSEARCH_JAVA_OPTS` but claims this is `OPENSEARCH_EXPERIMENTAL_FEATURES` which it's not, it can be any opts. Get rid of mentions of "experimental features".
> 
> An alternate and likely a nicer implementation could be to actually pass a list of experimental features.
> 
> In test-spec.yml.
> 
> ```yaml
> - version: 2.16.0
>    features:
>      - tiered_remote_index: true
> ```
> 
> To convert this into `OPENSEARCH_JAVA_OPTS` you could use [join](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/expressions#join).

I tried the following:

```
    steps:
      - name: Set OPENSEARCH_JAVA_OPTS
        run: |
          # Initialize options array
          opts=()
          prefix="-Dopensearch.experimental.feature"
          suffix="enabled"
          
          # Extract features directly from the matrix
          features=${{ matrix.entry.features }}
          
          # Split features into key-value pairs
          IFS=',' read -r -a feature_pairs <<< "$features"
          
          # Process each key-value pair
          for pair in "${feature_pairs[@]}"; do
            IFS='=' read -r key value <<< "$pair"
            opts+=("${prefix}.${key}.${suffix}=${value}")
          done
          
          # Set the environment variable
          echo "OPENSEARCH_JAVA_OPTS=$(IFS=','; echo "${opts[*]}")" >> $GITHUB_ENV
          echo $(cat $GITHUB_ENV)
```

However, its coming as `OPENSEARCH_JAVA_OPTS=-Dopensearch.experimental.feature.Array.enabled=`

refer: https://github.com/opensearch-project/opensearch-api-specification/actions/runs/10189736051/job/28188313025

So, I am going to go with opts for now unless u have another suggestion.

dblock (approved):

dblock:
Looks like the update SHA is bringing in a new feature that's not documented in the spec. Either add it (looks easy), or change the SHA back to a previous one for now.

dblock:
I updated the 2.16 ref to the latest in #459 and tests didn't fail the way they did here, so I looked at the `@sha256:50fbfe3b95c41e92a113ada3e80513ba4524dfc8a25dc6aaeff2bbe1e1145d5f` SHA and locally all my tests passed. So I added `-Dopensearch.experimental.feature.tiered_remote_index.enabled=true` and it failed like in this PR. Looks like the `cache_utilized` and `file_cache` properties are only enabled with this experimental feature enabled.

.github/opensearch-cluster/docker-compose.yml
```diff
@@ -8,6 +8,7 @@ services:
       - '9600:9600'
     environment:
       - 'OPENSEARCH_INITIAL_ADMIN_PASSWORD=${OPENSEARCH_PASSWORD:-myStrongPassword123!}'
+      - 'OPENSEARCH_JAVA_OPTS=${OPENSEARCH_JAVA_OPTS}'
```
dblock, line 11: In my experiments these quotes are unnecessary on both `OPENSEARCH_JAVA_OPTS` and `OPENSEARCH_INITIAL_ADMIN_PASSWORD`, so you can remove them.

neetikasinghal:
> I updated the 2.16 ref to the latest in #459 and tests didn't fail the way they did here, so I looked at the `@sha256:50fbfe3b95c41e92a113ada3e80513ba4524dfc8a25dc6aaeff2bbe1e1145d5f` SHA and locally all my tests passed. So I added `-Dopensearch.experimental.feature.tiered_remote_index.enabled=true` and it failed like in this PR. Looks like the `cache_utilized` and `file_cache` properties are only enabled with this experimental feature enabled.

Looks like its coming from here: https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/node/Node.java#L385 - this doesnt look right.

dblock:
> > I updated the 2.16 ref to the latest in #459 and tests didn't fail the way they did here, so I looked at the `@sha256:50fbfe3b95c41e92a113ada3e80513ba4524dfc8a25dc6aaeff2bbe1e1145d5f` SHA and locally all my tests passed. So I added `-Dopensearch.experimental.feature.tiered_remote_index.enabled=true` and it failed like in this PR. Looks like the `cache_utilized` and `file_cache` properties are only enabled with this experimental feature enabled.
> 
> Looks like its coming from here: https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/node/Node.java#L385

Are you looking into it? Open an issue at least?

dblock:
> > > > I updated the 2.16 ref to the latest in #459 and tests didn't fail the way they did here, so I looked at the `@sha256:50fbfe3b95c41e92a113ada3e80513ba4524dfc8a25dc6aaeff2bbe1e1145d5f` SHA and locally all my tests passed. So I added `-Dopensearch.experimental.feature.tiered_remote_index.enabled=true` and it failed like in this PR. Looks like the `cache_utilized` and `file_cache` properties are only enabled with this experimental feature enabled.
> > > 
> > > 
> > > Looks like its coming from here: https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/node/Node.java#L385
> > 
> > 
> > Are you looking into it? Open an issue at least?
> 
> yeah checking with @andrross to understand his thought process behind this change - doesnt look super right to me as If the tiered remote index feature flag is on and there is not search node it the cluster it would set the NODE_SEARCH_CACHE_SIZE_SETTING to 80%.

For this PR, remove `opts: -Dopensearch.experimental.feature.tiered_remote_index.enabled=true` from CI and it will go green, so we can then merge it.

In your other PR where you actually need `-Dopensearch.experimental.feature.tiered_remote_index.enabled=true` we need a fix. If it's in OpenSearch we can wait for it.

dblock:
The intermittent failure is an OOM in the sql plugin, I changed the threshold in another PR.

dblock (approved):

Issue #231, opened 2024-04-11
[FEATURE] Add specs for `ppl` namespace


nhtruong:
Part of [Adding missing API specs issue](https://github.com/opensearch-project/opensearch-api-specification/issues/168)

- [ ] /_opendistro/_ppl: ["POST"]
- [ ] /_opendistro/_ppl/_explain: ["POST"]
- [ ] /_opendistro/_ppl/stats: ["GET","POST"]
---
- [ ] /_plugins/_ppl: ["POST"]
- [ ] /_plugins/_ppl/_explain: ["POST"]
- [ ] /_plugins/_ppl/stats: ["GET","POST"]

Tokesh:
Hello! I would like to work on this issue :)

dblock:
Please!

Tokesh:
I think we can close this issue by [#460]

PR #460, opened 2024-08-01
adding ppl namespace api specs

Tokesh:
### Description
adding ppl namespace specs

### Issues Resolved
[#231]


Tokesh:
Still a draft. I want to refactor the structures, as there is an intersection with the SQL plugin. But the main work is done and the tests are written.

spec/namespaces/ppl.yaml
```diff
@@ -0,0 +1,165 @@
...
+    ppl.get_stats::query.format:
+      name: format
+      in: query
+      description: A short version of the Accept header, e.g. json, yaml.
+      schema:
+        type: string
+    ppl.get_stats::query.sanitize:
+      name: sanitize
+      in: query
+      description: Specifies whether to escape special characters in the results
```
dblock, line 105: Let's add periods at the end of all these sentences.

spec/schemas/_plugins._common.yaml
```diff
@@ -0,0 +1,78 @@
+openapi: 3.1.0
+info:
+  title: Schemas of _plugins._common category.
+  description: Schemas of _plugins._common category.
+  version: 1.0.0
+paths: {}
+components:
+  schemas:
+    Query:
```
dblock, line 9: Are these common to all plugins or just SQL? Feels like sql, so should it go to `_plugins/sql/_common.yaml`?

tests/ppl/explain.yaml
```diff
@@ -0,0 +1,22 @@
...
+  - path: /books
+    method: DELETE
+    status: [200, 404]
+chapters:
+  - synopsis: Get explain of SQL Query
+    path: /_plugins/_ppl/_explain
+    method: POST
+    request_body:
+      payload:
+        query: search source=books;
```
dblock, line 20: Curious, does it require that semicolon?

tests/ppl/explain.yaml
```diff
@@ -0,0 +1,22 @@
...
+  - path: /books
+    method: PUT
+    request_body:
+      payload: {}
+epilogues:
+  - path: /books
+    method: DELETE
+    status: [200, 404]
+chapters:
+  - synopsis: Get explain of SQL Query
```
dblock, line 15: Add a period. We should really add it to the linter :)

dblock (changes requested):
Nice. Mostly nitty asks.

tests/ppl/explain.yaml
Tokesh: Yes, I will try to finish the linter, because for me it happens so often :)

tests/ppl/explain.yaml
Tokesh: I just checked, it works without semicolons.
There are no semicolons in the examples of working with the OS API. That's why I removed them in the commit

spec/schemas/_plugins._common.yaml
Tokesh: I got this, you're right! I fixed it

tests/ppl/explain.yaml
dblock: I had a PR almost ready for some of it, https://github.com/opensearch-project/opensearch-api-specification/pull/463.

dblock (approved):

PR #473, opened 2024-08-07
Document fielddata.yaml fail, ignore vim swp files

awick:
### Description
* Ignore test files that don't end with .yaml so vim swp files are ignored
* fielddata.yaml test files if opensearch cluster was stopped with docker kill instead of docker stop, document the issue
* upgrade remaining actions/checkout@v3 to v4 to stop warnings

### Issues Resolved
N/A


dblock (changes requested):
It's a fine workaround, but maybe we should collect all `.yaml` files instead? Add a test?

dblock:
@awick I needed to exclude `docker-compose.yml` in another PR, you can build on top of https://github.com/opensearch-project/opensearch-api-specification/pull/472/commits/140c0fd1579e131a3d89f69fa549cdd14866aeda.

awick:
@dblock  I think this is good now, and I don't think that failure is me.

dblock (dismissed):

tools/src/tester/TestRunner.ts
```diff
@@ -69,6 +69,9 @@ export default class TestRunner {
     const path = file === '' ? folder : `${folder}/${file}`
     const next_prefix = prefix === '' ? file : `${prefix}/${file}`
     if (fs.statSync(path).isFile()) {
+      if (!path.endsWith('.yaml')) {
+        return []
+      }
```
dblock, line 74: This is fine, but if you look below there's `EXCLUDED_FILES.includes()`, feels like this is where we should be checking the .yaml extension as well to keep things in one place?

awick: Happy to move it, when I looked at it that seemed to be when it was reading files in a directory, so this will catch even the case when a file was used on the command line.

So this would not run:
npm run test:spec -- --opensearch-insecure --tests-path tests/cat/badfile

but that is an edge case I guess. Let me know which you prefer.

awick: or maybe i should move the EXCLUDE? Or actually the current list is a file that won't run anymore anyway, lol.

dblock: Right. So we definitely don't want both checks, pick your favorite implementation and update?

dblock (approved):

PR #456, opened 2024-07-31
Adding SQL Settings specs

Tokesh:
### Description
adding sql settings specs

### Issues Resolved
[#235]

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #474, opened 2024-08-07
Observability namespace

### LINKED ISSUE #230: [FEATURE] Add specs for `observability` namespace
Part of [Adding missing API specs issue](https://github.com/opensearch-project/opensearch-api-specification/issues/168)

- [ ] /_plugins/_observability/_local/stats: ["GET"]
- [ ] /_plugins/_observability/object: ["GET","POST","DELETE"]
- [ ] /_plugins/_observability/object/{objectId}: ["GET","PUT","DELETE"]
---
- [ ] /_plugins/_query/_datasources: ["GET","PUT","POST"]
- [ ] /_plugins/_query/_datasources/{dataSourceName}: ["GET","DELETE"]
- [ ] /_plugins/_query/settings: ["PUT"]

aabeshov:
### Description

Adding observability namespace specs.

### Issues Resolved

Closes #230.


dblock (changes requested):
Great start! Needs tests. 

You can add these specs piecemeal instead of everything at once, maybe start with something simple with tests, changelog, etc.?

nhtruong:
Agree with @dblock. This is a lot in one PR esp once you add the tests. Do break it into smaller PRs and thank you so much for the hard work :)

aabeshov:
> Agree with @dblock. This is a lot in one PR esp once you add the tests. Do break it into smaller PRs and thank you so much for the hard work :)

It's okay ,thanks, I guess I am close to the finish line with these specs
Check Links is failing, is it connected with test stories that I create ?
Because my test stories are passing locally
image

dblock:
Looks like only the link checker is failing against `http://localhost:9090` which is mentioned in a test. 

Add an exclude file like [here](https://github.com/opensearch-project/OpenSearch/blob/main/.github/workflows/links.yml#L18C90-L18C121) or change the test?

tests/observability/observability.yaml
```diff
@@ -0,0 +1,162 @@
...
+    path: /_plugins/_observability/object
+    method: GET
+    response:
+      status: 200
+      payload:
+        startIndex: 0
+        totalHits: 0
+        totalHitRelation: eq
+        observabilityObjectList: []
+
```
dblock, line 155: Nit: let's get rid of double empty lines ;)

aabeshov:
I changed the lychee ignore. --exclude-file is deprecated.
BTW, I haven't implemented  /_plugins/_query/settings: ["PUT"] this api , because I guess @Tokesh has already made it in this PR [#456 ]

.cspell
```diff
@@ -185,4 +185,6 @@ urldecode
 vectory
 whoamiprotected
 wordnet
-Yrtsd
\ No newline at end of file
+Yrtsd
+Refn
```
dblock, line 189: We actually removed this guy.

.lycheeignore
```diff
@@ -1 +1,3 @@
 https://localhost:9200*
+http://localhost:9090
+http://localhost:9091
```
dblock, line 3: Maybe `https://localhost:*` ?

spec/namespaces/observability.yaml
```diff
@@ -0,0 +1,224 @@
...
+          schema:
+            $ref: '../schemas/observability._common.yaml#/components/schemas/ObservabilityObject'
+    observability.update_object:
+      content:
+        application/json:
+          schema:
+            $ref: '../schemas/observability._common.yaml#/components/schemas/ObservabilityObject'
+  responses:
+    observability.list_objects@200:
+      description: Successful response of retrieving all Observability Objects
```
dblock, line 101: Add a period. (Note to self, we need a validator.)

tests/observability/datasources.yaml
```diff
@@ -0,0 +1,94 @@
+$schema: ../../json_schemas/test_story.schema.yaml
```
dblock, line 1: This is for `query`, so it should live in `tests/query/datasources.yml`, right?

dblock (changes requested):
This looks good! I have some small asks and let's get this merged.

tests/observability/datasources.yaml
aabeshov: I will move datasource test story to /tests/default/query/. But do you think I should move observability to default ?  Cause as far as I know observability is an external service

dblock: This is a bit new in this repo.

Folders like `default` and `snapshot` are there to have separate docker-compose.yml, not because something is external or internal. These are "test suites". So you should move into `tests/default/query` and `tests/default/observability`. 

If observability required some fancy setup, you could create a new folder for it and add a job to the GHA, but it is hopefully not necessary.

aabeshov:
I don't know why but there is a bad request when test story is making DELETE /_plugins/_query/_datasources/{datasource_name}

dblock:
> @dblock ,I don't know why but there is a 400 bad request when test story is making DELETE /_plugins/_query/_datasources/{datasource_name} However locally it's okay

We should show the detailed output for these errors (please open an issue). 

For now, add a `--verbose` into the GHA workflow to see the details?

aabeshov:
> > @dblock ,I don't know why but there is a 400 bad request when test story is making DELETE /_plugins/_query/_datasources/{datasource_name} However locally it's okay
> 
> We should show the detailed output for these errors (please open an issue).
> 
> For now, add a `--verbose` into the GHA workflow to see the details? Note that it fails on older versions, maybe a quick start of a container with 1.3.x locally will reproduce it?

[INFO] => DELETE /_plugins/_query/_datasources/test_datasource ({}) [application/json] 
[INFO] <= 400 (application/json) | "no handler found for uri [/_plugins/_query/_datasources/test_datasource] and method [DELETE]"
[INFO] => DELETE /_plugins/_query/_datasources/test_datasource ({}) [application/json] 
[INFO] <= 400 (application/json) | "no handler found for uri [/_plugins/_query/_datasources/test_datasource] and method [DELETE]"

ERROR   query/datasources.yaml (/home/runner/work/opensearch-api-specification/opensearch-api-specification/tests/default/query/datasources.yaml)
    ERROR   PROLOGUES
        ERROR   DELETE /_plugins/_query/_datasources/{datasource_name} (Bad Request)
    SKIPPED CHAPTERS
        SKIPPED Create a new query datasource.
        SKIPPED Retrieve the list of all query datasources.
        SKIPPED Retrieve a specific query datasource by name.
        SKIPPED Update an existing query datasource.
        SKIPPED Retrieve the updated query datasource by name.
        SKIPPED Delete the query datasource by name.
    ERROR   EPILOGUES
        ERROR   DELETE /_plugins/_query/_datasources/{datasource_name} (Bad Request)
        
        Here is the details. I am not sure, but maybe there is not such api in old versions

spec/namespaces/observability.yaml
```diff
@@ -0,0 +1,224 @@
...
+info:
+  title: OpenSearch Observability Object API
+  description: API for retrieving and managing Observability Objects.
+  version: 1.0.0
+paths:
+  /_plugins/_observability/_local/stats:
+    get:
+      operationId: observability.get_localstats.0
+      x-operation-group: observability.get_localstats
+      x-version-added: '1.1'
```
Tokesh, line 11: please recheck `x-version-added`. i think it can fix your ci issue

aabeshov: @Tokesh committed changes , you can check

.github/workflows/test-spec.yml
```diff
@@ -76,6 +76,7 @@ jobs:
             --opensearch-insecure \
             --opensearch-version=${{ matrix.entry.version }} \
             --coverage coverage/test-spec-coverage-${{ matrix.entry.version }}-${{ matrix.entry.tests || 'default' }}.json \
+            --verbose \
```
dblock, line 79: Let's not leave this here, it spews a lot of logs. Remove.

.lycheeignore
```diff
@@ -1 +1,2 @@
 https://localhost:9200*
+http://localhost:*
```
dblock, line 2: Let's consolidate this into a single `https://localhost:*` and change the prometheus URL below to an `https://` one so it catches both?

tests/default/query/datasources.yaml
```diff
@@ -0,0 +1,95 @@
...
+    path: /_plugins/_query/_datasources
+    method: POST
+    request:
+      payload:
+        name: test_datasource
+        description: ''
+        connector: PROMETHEUS
+        allowedRoles: []
+        properties:
+          prometheus.uri: 'http://localhost:9090'
```
dblock, line 23: This (and a few other instances below) is the URL that gets tripped by the link checker. Change it to `https://...`

I am surprised the URL is actually not checked, is this expected? I'd think the API fails if prometheus is not available?

dblock (changes requested):
Looks good. I have 2 asks below.

tests/default/query/datasources.yaml
aabeshov: It sound like it should fail, but I tested API in the playground with turned off prometheus and got 200 without errors

dblock (approved):

dblock:
Looks like one of the tests here may fail intermittently:

```
        FAILED  Retrieve list of Observability objects.
            PASSED  REQUEST BODY
            PASSED  RESPONSE STATUS
            FAILED  RESPONSE PAYLOAD BODY (expected totalHits='0', got '1')
            PASSED  RESPONSE PAYLOAD SCHEMA
```

https://github.com/opensearch-project/opensearch-api-specification/actions/runs/10354531782/job/28660019203?pr=493

@aabeshov care to take a look please?

dblock:
Thanks! I fixed it in https://github.com/opensearch-project/opensearch-api-specification/pull/497 since it was failing often. Does that look right? Next time don't hold back and just make a PR :)

aabeshov:
> Thanks! I fixed it in #497 since it was failing often. Does that look right? Next time don't hold back and just make a PR :)



> Thanks! I fixed it in #497 since it was failing often. Does that look right? Next time don't hold back and just make a PR :)

Thanks , but I am really confused because I chapters first I do get request to specific object and it return that It exists then right after this step I do request to get list of all objects and sometimes it returns that list is empty that why it is expected for index to be 0 in some cases.

dblock:
Writing something doesn't guarantee that it will be read right away, it's eventually consistent. Since these options are backed by an index, refreshing means waiting for a full refresh cycle where everything written gets read consistently, making it predictable. Does this help?

https://opensearch.org/docs/latest/api-reference/index-apis/flush/

aabeshov:
> Writing something doesn't guarantee that it will be read right away, it's eventually consistent. Since these options are backed by an index, refreshing means waiting for a full refresh cycle where everything written gets read consistently, making it predictable. Does this help?
> 
> https://opensearch.org/docs/latest/api-reference/index-apis/refresh/

aaa okay now I got this, thanks!

PR #476, opened 2024-08-08
Added support for SigV4.

dblock:
### Description

Enables running tests against Amazon OpenSearch.


eslint.config.mjs
```diff
@@ -78,6 +80,8 @@ export default [
       ],
       '@typescript-eslint/require-await': 'error',
       '@typescript-eslint/no-unused-vars': ['error', { argsIgnorePattern: '^_' }],
+      '@typescript-eslint/unbound-method': 'off',
```
nhtruong, line 83: Should we leave this on and evaluate unbound method case by case?

tools/src/OpenSearchHttpClient.ts
```diff
@@ -72,26 +115,58 @@ export interface OpenSearchInfo {
...
+          service: 'es'
+        },
+        credentials: {
+          accessKeyId: opts.aws_access_key_id,
+          secretAccessKey: opts.aws_access_secret_key,
+          sessionToken: opts.aws_access_session_token,
+        }
+      });
+    } else {
+      this._logger?.warn(`No credentials provided, did you forget to set OPENSEARCH_PASSWORD or AWS_ACCESS_KEY_ID?`)
```
nhtruong, line 150: Instead of checking individual components of different authentication methods, we should leverage Typescript's typing to stream line this:
```ts
export interface BasicAuth {
  username: string
  password: string
}

export interface AwsAuth {
  aws_access_key_id: string
  aws_access_secret_key: string
  aws_access_session_token: string
  aws_region: string
  aws_service: string
}

export interface OpenSearchHttpClientOptions {
  url?: string
  insecure?: boolean
  responseType?: ResponseType
  logger?: Logger
  basic_auth: BasicAuth
  aws_auth: AwsAuth
}

export class OpenSearchHttpClient {
  private readonly _axios: AxiosInstance
  private readonly _opts: OpenSearchHttpClientOptions
  private readonly _logger?: Logger

  constructor (opts: OpenSearchHttpClientOptions = {}) { // Note that I also give opts the default value so we don't have to keep typechecking it for null
    this._opts = opts
    this._logger = opts?.logger

    let auth = undefined
    let sigv4_interceptor = undefined

    if (opts.basic_auth != null) {
      this._logger?.info(`Authenticating with ${opts.username} ...`)
      auth = opts.basic_auth
    } else if (opts.aws_auth != null) {
      this._logger?.info(`Authenticating using SigV4 with ${opts.aws_access_key_id} (${opts.aws_region}) ...`)
      ...
    } else {
      this._logger?.warn(`No credentials provided, did you forget to set OPENSEARCH_PASSWORD or AWS_ACCESS_KEY_ID?`)
    }
...
```
This also makes it easier to use the Client as if you decide to use aws_auth to instantiate the client, TS will make sure that it has all 6 required components before the user runs the code.

tools/src/OpenSearchHttpClient.ts
```diff
@@ -72,26 +115,58 @@ export interface OpenSearchInfo {
 export class OpenSearchHttpClient {
   private readonly _axios: AxiosInstance
   private readonly _opts?: OpenSearchHttpClientOptions
+  private readonly _logger?: Logger
```
nhtruong, line 118: I prefer having a default null-logger than having to typecheck if `_logger` is null everywhere it is used.

nhtruong (approved):

PR #477, opened 2024-08-08
Add Segments API test

Naarcha-AWS:
This PR adds an API test for the segments API.

I'm not sure how much detail I should add to the response part of the test, could use some guidance there.


CHANGELOG.md
```diff
@@ -66,6 +66,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Added `creation_date` field to `DanglingIndex` ([#462](https://github.com/opensearch-project/opensearch-api-specification/pull/462))
 - Added doc on `cluster create-index blocked` workaround ([#465](https://github.com/opensearch-project/opensearch-api-specification/pull/465))
 - Added support for reusing output variables as keys in payload expectations ([#471](https://github.com/opensearch-project/opensearch-api-specification/pull/471))
+- Added test for `indices/_segments` ([#1714](https://github.com/opensearch-project/opensearch-api-specification/pull/477))
```
dblock, line 69: You can remove this. We usually add changes that are publicly facing only (spec changes).

dblock (commented):
I labeled this `skip-changelog`, remove the CHANGELOG line below.

> I'm not sure how much detail I should add to the response part of the test, could use some guidance there.

Generally it's not needed, but it's good practice if there's specific data that otherwise would not be present. Is there something in the response of `GET /index/_segments` that can confirm that the segment belongs to the index?

dblock (approved):

PR #478, opened 2024-08-08
Catch response deserialization errors.

dblock:
### Description

After doing #476 I ran tests against AOS 2.7 and ran into an error deserializing CBOR. The implementation here will catch deserialization errors and bubble them up instead of failing unexpectedly.

Before:

```
$ npm run test:spec -- --tests tests/default/cat/health.yaml 

> ts-node tools/src/tester/test.ts --tests tests/default/cat/health.yaml

OpenSearch 2.7.0

/Users/dblock/source/opensearch-project/opensearch-api-specification/dblock-opensearch-api-specification/node_modules/cbor/lib/decoder.js:469
          throw new Error(`Additional info not implemented: ${ai}`)
                ^
Error: Additional info not implemented: 29
    at Decoder._parse (/Users/dblock/source/opensearch-project/opensearch-api-specification/dblock-opensearch-api-specification/node_modules/cbor/lib/decoder.js:469:17)
    at _parse.next (<anonymous>)
    at Object.decodeFirstSync (/Users/dblock/source/opensearch-project/opensearch-api-specification/dblock-opensearch-api-specification/node_modules/cbor/lib/decoder.js:228:22)
    at ChapterReader.#deserialize_payload (/Users/dblock/source/opensearch-project/opensearch-api-specification/dblock-opensearch-api-specification/tools/src/tester/ChapterReader.ts:112:69)
    at /Users/dblock/source/opensearch-project/opensearch-api-specification/dblock-opensearch-api-specification/tools/src/tester/ChapterReader.ts:53:48
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async ChapterReader.read (/Users/dblock/source/opensearch-project/opensearch-api-specification/dblock-opensearch-api-specification/tools/src/tester/ChapterReader.ts:41:5)
    at async ChapterEvaluator.#evaluate (/Users/dblock/source/opensearch-project/opensearch-api-specification/dblock-opensearch-api-specification/tools/src/tester/ChapterEvaluator.ts:65:22)
    at async ChapterEvaluator.evaluate (/Users/dblock/source/opensearch-project/opensearch-api-specification/dblock-opensearch-api-specification/tools/src/tester/ChapterEvaluator.ts:50:16)
    at async StoryEvaluator.#evaluate_chapters (/Users/dblock/source/opensearch-project/opensearch-api-specification/dblock-opensearch-api-specification/tools/src/tester/StoryEvaluator.ts:89:28)
```

After:

```
$ npm run test:spec -- --tests tests/default/cat/health.yaml 

> ts-node tools/src/tester/test.ts --tests tests/default/cat/health.yaml

OpenSearch 2.7.0


        ERROR   Cat in different formats (format=cbor).
            PASSED  PARAMETERS
                PASSED  format
            PASSED  REQUEST BODY
            ERROR   RESPONSE STATUS (Expected status 200, but failed to read the response. Additional info not implemented: 29)
            SKIPPED RESPONSE PAYLOAD BODY
```

### DIFF
--- a/tools/src/tester/ChapterEvaluator.ts
+++ b/tools/src/tester/ChapterEvaluator.ts
@@ -122,7 +122,7 @@ export default class ChapterEvaluator {
 
   #evaluate_status(chapter: Chapter, response: ActualResponse): Evaluation {
     const expected_status = chapter.response?.status ?? 200
-    if (response.status === expected_status) return { result: Result.PASSED }
+    if (response.status === expected_status && response.error === undefined) return { result: Result.PASSED }
 
     let result: Evaluation = {
       result: Result.ERROR,
--- a/tools/src/tester/ChapterEvaluator.ts
+++ b/tools/src/tester/ChapterEvaluator.ts
@@ -124,13 +124,15 @@ export default class ChapterEvaluator {
     const expected_status = chapter.response?.status ?? 200
     if (response.status === expected_status) return { result: Result.PASSED }
 
-    const result: Evaluation = {
+    let result: Evaluation = {
       result: Result.ERROR,
       message: _.join(_.compact([
-        `Expected status ${expected_status}, but received ${response.status}: ${response.content_type}.`,
+        expected_status == response.status ?
+          `Received ${response.status ?? 'none'}: ${response.content_type ?? 'unknown'}.` :
+          `Expected status ${expected_status}, but received ${response.status ?? 'none'}: ${response.content_type ?? 'unknown'}.`,
         response.message
       ]), ' ')
-    };
+    }
 
     if (response.error !== undefined) {
       result.error = response.error as Error
--- a/tools/src/tester/ChapterReader.ts
+++ b/tools/src/tester/ChapterReader.ts
@@ -54,7 +54,7 @@ export default class ChapterReader {
       if (payload !== undefined) response.payload = payload
       this.logger.info(`<= ${r.status} (${r.headers['content-type']}) | ${to_json(response.payload)}`)
     }).catch(e => {
-      if (e.response === undefined) {
+      if (e.response == null) {
         this.logger.info(`<= ERROR: ${e}`)
         response.message = e.message
         response.error = e
@@ -64,7 +64,6 @@ export default class ChapterReader {
         const payload = this.#deserialize_payload(e.response.data, response.content_type)
         if (payload !== undefined) response.payload = payload.error
         response.message = payload.error?.reason ?? e.response.statusText
-        response.error = e
         this.logger.info(`<= ${response.status} (${response.content_type}) | ${response.payload !== undefined ? to_json(response.payload) : response.message}`)
       }
     })
--- a/tools/src/tester/ChapterReader.ts
+++ b/tools/src/tester/ChapterReader.ts
@@ -54,18 +54,19 @@ export default class ChapterReader {
       if (payload !== undefined) response.payload = payload
       this.logger.info(`<= ${r.status} (${r.headers['content-type']}) | ${to_json(response.payload)}`)
     }).catch(e => {
-      if (e.response == null) {
+      if (e.response === undefined) {
         this.logger.info(`<= ERROR: ${e}`)
-        throw e
+        response.message = e.message
+        response.error = e
+      } else {
+        response.status = e.response.status
+        response.content_type = e.response.headers['content-type']?.split(';')[0]
+        const payload = this.#deserialize_payload(e.response.data, response.content_type)
+        if (payload !== undefined) response.payload = payload.error
+        response.message = payload.error?.reason ?? e.response.statusText
+        response.error = e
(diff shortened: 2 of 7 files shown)


tools/src/tester/StoryEvaluator.ts
```diff
@@ -86,12 +86,24 @@ export default class StoryEvaluator {
...
+          }
+          evaluations.push(evaluation)
+        } catch (e) {
+          has_errors = true
+          const title = chapter.synopsis || `${chapter.method} ${chapter.path}`
+          const error = e as Error
+          evaluations.push({
+            title,
+            overall: {
+              result: Result.ERROR, message: error.message, error
```
nhtruong, line 103: This is a very broad `catch`. We should limit it to the errors that we know and still throw the other errors as usual.

dblock: Updated to treat errors as failures instead of re-throwing.

tools/tests/tester/fixtures/evals/error/prologue_error.yaml
```diff
@@ -16,7 +16,6 @@ prologues:
     overall:
       result: ERROR
       message: no such index [does_not_exists]
-      error: Request failed with status code 404
```
dblock, line 19: These aren't actually errors per-se, they get parsed and properly reflected in the failures of the test.

tools/src/tester/ChapterReader.ts
```diff
@@ -54,18 +54,18 @@ export default class ChapterReader {
       if (payload !== undefined) response.payload = payload
       this.logger.info(`<= ${r.status} (${r.headers['content-type']}) | ${to_json(response.payload)}`)
     }).catch(e => {
-      if (e.response == null) {
+      if (e.response === undefined) {
```
nhtruong, line 57: Just curious. Maybe I've gotten it wrong.  Isn't `== null` the same as `=== undefined || ===null` and `==null` is still valid in this case?

dblock: I am just too scared of `== null`, it can only be undefined here AFAIK.

tools/src/tester/ChapterEvaluator.ts
```diff
@@ -122,15 +122,17 @@ export default class ChapterEvaluator {
 
   #evaluate_status(chapter: Chapter, response: ActualResponse): Evaluation {
     const expected_status = chapter.response?.status ?? 200
-    if (response.status === expected_status) return { result: Result.PASSED }
+    if (response.status === expected_status && response.error === undefined) return { result: Result.PASSED }
```
nhtruong, line 125: Is checking `response.error === undefined` necessary? From the surface when reading this line, I interpret the business logic as: Even if the status is 403 as the test story expects, the test will still fail regardless because it's an error. 

Meanwhile the implementation [here](https://github.com/opensearch-project/opensearch-api-specification/pull/478/files#diff-92cb1226311bbe71a379072f350975f933601c69605c601ec44d7bf7aaf019a6R59-R60) says other wise. It guarantees that only either `response.status` or `response.error` is set, but not both. That means if `response.status === expected_status (which is not null)` then `response.error === undefined` is also true.

It a small thing to nit-pick but keeping the guards simple and close to the high-level logic makes it a lot easier to maintain :)

tools/src/tester/ChapterReader.ts
nhtruong: I think `== null` is favorable in this case, and most situations. Consider these 2 scenarios:
```ts
const x = {}
x.foo = null
console.log(x.foo) // null
console.log(x.foo == null) // true
console.log(x.foo === undefined) // false
```
```ts
const x= {}
consolee.log(x.foo) // undefined
console.log(x.foo == null) // true
console.log(x.foo === undefined) // true
```

As you can see, if `x.foo` is explicitly set to `null`, then `=== undefined` will fail, while `== null` accounts for both whether `x.foo` is unset or explicitly set to `null`

tools/src/tester/ChapterEvaluator.ts
dblock: If the server responded with the expected status (200), but data failed to deserialize, an exception will be thrown and you will end up with a matching status, but no `response.error`.

nhtruong: In this [snippet](https://github.com/opensearch-project/opensearch-api-specification/pull/478/files#diff-92cb1226311bbe71a379072f350975f933601c69605c601ec44d7bf7aaf019a6R57-R68), `response.status` and `response.error` are assigned in different code paths (i.e. mutual exclusive), how do you get them both at the same time?

It's a just a small nitpick tho. If it does happens like you describe then it's all good :)

nhtruong (approved):

tools/src/tester/ChapterEvaluator.ts
dblock: Is that `catch` re-throwing into itself?!

Issue #398, opened 2024-07-10
[FEATURE] Run tests against multiple versions of OpenSearch


dblock:
### Is your feature request related to a problem?

In https://github.com/opensearch-project/opensearch-api-specification/pull/356 we are adding an API spec for the **next** version of OpenSearch. The tests running currently against 2.15 will fail.

### What solution would you like?

1. Expand the test matrix to run against a stable and the next version of OpenSearch.
2. Add a feature that lets one mark the minimum version of server required for tests.
3. Only run tests against any specific version of OpenSearch that are compatible.

Use nightly docker images https://hub.docker.com/r/opensearchstaging/opensearch/tags pointing to next version of opensearch. As it is nightly, it might not have all the plugins installed.

### What alternatives have you considered?

Always run tests against the next version of OpenSearch.

PR #483, opened 2024-08-08
Added support for testing multiple distributions.

dblock:
### Description

With this change, `npm run test:spec -- --opensearch-distribution=aos` passes against AOS 2.13.

### Issues Resolved

Part of #398.
 

nhtruong:
Should we have `opensearch.org` as the default distribution for the test stories as well since it's the default in the spec?

DEVELOPER_GUIDE.md
```diff
@@ -146,6 +146,7 @@ This repository includes several OpenAPI Specification Extensions to fill in any
 - `x-ignorable`: Denotes that the operation should be ignored by the client generator. This is used in operation groups where some operations have been replaced by newer ones, but we still keep them in the specs because the server still supports them.
 - `x-global`: Denotes that the parameter is a global parameter that is included in every operation. These parameters are listed in the [spec/_global_parameters.yaml](spec/_global_parameters.yaml).
 - `x-default`: Contains the default value of a parameter. This is often used to override the default value specified in the schema, or to avoid accidentally changing the default value when updating a shared schema.
+- `x-distributions`: Contains a list of distributions known to include the API. Use `opensearch.org` for the official distribution, `aos` for Amazon Managed OpenSearch, and `aoss` for Amazon OpenSearch Serverless.
```
Xtansia, line 149: I think having `aos` and `aoss` as shortened versions is not immediately grokkable to someone without some AWS specific context. Even as someone with that context having one be a prefix of the other immediately hurts readability. Could we expand these to `amazon-managed` & `amazon-serverless` given we have a full form `opensearch.org`?

We should think ahead for what precedent we set if in future we end up with contributions related to things like [Oracle Cloud Infrastructure Search with OpenSearch](https://www.oracle.com/cloud/search/). (`oci`, `ocisos`, `oracle`, `oracle-managed`, `oci-managed`, etc)

dblock: Good point.  One thought I had is to use URIs, e.g. `aws.amazon.com/opensearch-service`? WDYT?

Xtansia:
> Should we have `opensearch.org` as the default distribution for the test stories as well since it's the default in the spec?

I feel like we might really want both an "allow"/"include" and "deny"/"exclude" variant. As in the majority of cases all or most distributions will support an API and just `aoss` doesn't. So my thought is you'd generally by default want new tests/operations to be automatically opted into for all/new distributions and then whittled down as necessary. (Assuming we get automated testing against the managed distributions)

DEVELOPER_GUIDE.md
Xtansia: Could be a bit verbose, AOSS would be `aws.amazon.com/opensearch-service/features/serverless` presumably.

Either way we'd likely need to define an enum somewhere and validate correctness, as spelling mistakes will be a PITA

dblock: I tried to use `ajv.addKeyword` to define a custom type for the new `x-distribution` options, but couldn't make it work. Maybe @nhtruong can help me make these values check against an enum on top of this PR.

nhtruong: We're only using AJV to validate JSON Schemas in the spec. `x-distribution` is not used in a schema, but rather an operation. So it should be validated similar to how we validate `x-operation-group` and other extensions.

We should create a json-schema for the namespace files and another for the category files in the `./spec` folder. Even though we currently only require our `yaml` files to be OpenAPI valid, we actually require a much stricter json-schema for these files:
-  `x-operation-group` is supported but a typo like `x-opperation-group` should not. 
-  Someone uses a new and valid OpenAPI keyword that the clients should handle accordingly but it got through the PR review without updating the Client Gen doc.
- We can enforce validation on the extensions for free through AJV.

dblock:
I've added include and exclude variants, etc., but no value validation in this PR. Please take another look?

DEVELOPER_GUIDE.md
dblock: https://github.com/opensearch-project/opensearch-api-specification/issues/492

dblock:
> > Should we have `opensearch.org` as the default distribution for the test stories as well since it's the default in the spec?
> 
> I feel like we might really want both an "allow"/"include" and "deny"/"exclude" variant. As in the majority of cases all or most distributions will support an API and just `aoss` doesn't. So my thought is you'd generally by default want new tests/operations to be automatically opted into for all/new distributions and then whittled down as necessary. (Assuming we get automated testing against the managed distributions)

On this note. Do you @nhtruong @Xtansia have a (strong) preference for:

1. `x-distributions-included`, `x-distributions-included` and `distributions:` in tests.
2. ... 
    ```
    distributions:
       included:
       excluded: 
   ```
   in tests.
3. ... 
    ```
    distributions-included:
    distributions-excluded: 
   ```
   in tests.

DEVELOPER_GUIDE.md
```diff
@@ -146,6 +146,10 @@ This repository includes several OpenAPI Specification Extensions to fill in any
 - `x-ignorable`: Denotes that the operation should be ignored by the client generator. This is used in operation groups where some operations have been replaced by newer ones, but we still keep them in the specs because the server still supports them.
 - `x-global`: Denotes that the parameter is a global parameter that is included in every operation. These parameters are listed in the [spec/_global_parameters.yaml](spec/_global_parameters.yaml).
 - `x-default`: Contains the default value of a parameter. This is often used to override the default value specified in the schema, or to avoid accidentally changing the default value when updating a shared schema.
+- `x-distributions-included`: Contains a list of distributions known to include the API.
+- `x-distributions-excluded`: Contains a list of distributions known to exclude the API.
+
+Use `opensearch.org` for the official distribution in `x-distributions-*`, `amazon-opensearch` for Amazon Managed OpenSearch, and `amazon-serverless` for Amazon OpenSearch Serverless.
```
Xtansia, line 152: This says `amazon-opensearch` but the TESTING_GUIDE.md says `amazon-managed`

tools/src/tester/StoryEvaluator.ts
```diff
@@ -76,15 +86,18 @@ export default class StoryEvaluator {
...
 
-  async #evaluate_chapters(chapters: Chapter[], has_errors: boolean, dry_run: boolean, story_outputs: StoryOutputs, version?: string): Promise<ChapterEvaluation[]> {
+  async #evaluate_chapters(chapters: Chapter[], has_errors: boolean, dry_run: boolean, story_outputs: StoryOutputs, version?: string, distribution?: string): Promise<ChapterEvaluation[]> {
     const evaluations: ChapterEvaluation[] = []
     for (const chapter of chapters) {
       if (dry_run) {
         const title = chapter.synopsis || `${chapter.method} ${chapter.path}`
         evaluations.push({ title, overall: { result: Result.SKIPPED, message: 'Dry Run', error: undefined } })
-      } else if (version != undefined && chapter.version !== undefined && !semver.satisfies(version, chapter.version)) {
+      } else if (version != undefined && chapter.version !== undefined && semver.satisfies(version, chapter.version)) {
```
Xtansia, line 95: This change doesn't look correct

tools/src/tester/TestRunner.ts
```diff
@@ -34,19 +34,25 @@ export default class TestRunner {
...
+  async run (story_path: string, version?: string, distribution?: string, dry_run: boolean = false): Promise<{ results: StoryEvaluations, failed: boolean }> {
     let failed = false
     const story_files = this.story_files(story_path)
     const results: StoryEvaluations = { evaluations: [] }
 
     if (!dry_run) {
-      const info = await this._http_client.wait_until_available()
-      console.log(`OpenSearch ${ansi.green(info.version.number)}\n`)
-      version = info.version.number
+      if (distribution === 'aoss') {
```
Xtansia, line 43:
```suggestion
      if (distribution === 'amazon-serverless') {
```

tools/src/types.ts
```diff
@@ -16,6 +16,8 @@ export interface OperationSpec extends OpenAPIV3.OperationObject {
   'x-version-deprecated'?: string
   'x-deprecation-message'?: string
   'x-ignorable'?: boolean
+  'x-distributions-included'?: string
+  'x-distributions-excluded'?: string
```
Xtansia, line 20:
```suggestion
  'x-distributions-included'?: string[]
  'x-distributions-excluded'?: string[]
```

tools/src/OpenSearchHttpClient.ts
```diff
@@ -22,6 +22,10 @@ export const OPENSEARCH_URL_OPTION = new Option('--opensearch-url <url>', 'URL a
   .default(DEFAULT_URL)
   .env('OPENSEARCH_URL')
 
+export const OPENSEARCH_DISTRIBUTION_OPTION = new Option('--opensearch-distribution <key>', 'OpenSearch distribution')
+  .default('opensearch.org')
+  .env('OPENSEARCH_DISTRIBUTION')
+
```
Xtansia, line 28: Is there a reason this is being added into the `OpenSearchHttpClientOptions`?

Xtansia (changes requested):

Xtansia:
I think ideally we'd want `distributions: {included: [], excluded: []}` in tests, could still have `distributions: ['foo']` that just shortcuts to `distributions: {included: ['foo']}`.

There might be better verbiage than `(in|ex)cluded` (something like `supported` maybe?) but I haven't come up with any other options I'm 100% happy with so not critical to this PR.

tools/src/OpenSearchHttpClient.ts
dblock: Because it will need to set the version to 2.1 for serverless because that doesn't have `GET /`.

dblock: Moved it.

dblock:
> I think ideally we'd want `distributions: {included: [], excluded: []}` in tests, could still have `distributions: ['foo']` that just shortcuts to `distributions: {included: ['foo']}`.
> 
> There might be better verbiage than `(in|ex)cluded` (something like `supported` maybe?) but I haven't come up with any other options I'm 100% happy with so not critical to this PR.

Made updates per comments and rebased. We can expand into includes and excludes for the distribution in a future PR.

tools/src/OpenSearchHttpClient.ts
```diff
@@ -64,6 +64,7 @@ export interface AwsAuth {
 
 export interface OpenSearchHttpClientOptions {
   url?: string
+  distribution?: string,
```
Xtansia, line 67: We also don't need these other changes here now as it's not actually used by `OpenSearchHttpClient`

Xtansia (changes requested):

tools/src/OpenSearchHttpClient.ts
dblock: Updated.

Xtansia (approved):

PR #489, opened 2024-08-11
Fixed schema refs for ml.yaml

nhtruong:
The body schemas for some bodies in the `ml` namespace were incorrectly referenced to the content-type instead of their respective `schema` property.

### DIFF
--- a/spec/namespaces/ml.yaml
+++ b/spec/namespaces/ml.yaml
@@ -159,8 +159,8 @@ components:
               task_id:
                 type: string
             required:
-              - task_id
               - status
+              - task_id
     ml.delete_model@200:
       content:
         application/json:
--- a/spec/namespaces/ml.yaml
+++ b/spec/namespaces/ml.yaml
@@ -152,7 +152,15 @@ components:
       content:
         application/json:
           schema:
-            $ref: '../schemas/ml._common.yaml#/components/schemas/Task'
+            type: object
+            properties:
+              task_id:
+                type: string
+              status:
+                type: string
+            required:
+              - task_id
+              - status
     ml.delete_model@200:
       content:
         application/json:
--- a/spec/namespaces/ml.yaml
+++ b/spec/namespaces/ml.yaml
@@ -147,7 +147,7 @@ components:
       content:
         application/json:
           schema:
-            $ref: '../schemas/ml._common.yaml#/components/schemas/ModelGroup'
+            $ref: '../schemas/_common.yaml#/components/schemas/WriteResponseBase'
     ml.register_model@200:
       content:
         application/json:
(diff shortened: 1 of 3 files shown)


dblock (approved):
Should the checker have caught this as additional fields or something?

dblock (changes requested):
Looks like ML tests started failing.

nhtruong:
> Should the checker have caught this as additional fields or something?

We should create JSON-schemas for the namespace and category files to catch this among others, especially on catching typos and enforcing rules for extensions.

dblock:
> > Should the checker have caught this as additional fields or something?
> 
> We should create JSON-schemas for the namespace and category files to catch this among others, especially on catching typos and enforcing rules for extensions.

Yes, you want to take it? https://github.com/opensearch-project/opensearch-api-specification/issues/492

The tests here are failing though, so you'll need to add a fix.

nhtruong:
> The tests here are failing though, so you'll need to add a fix.

I saw that. I'm on it.
It also reveals a potential bug in the test framework: When the response schema is not defined in the spec (At least the framework thought so in this case because the schema was referenced in the wrong place), it skips response payload test and still pass the chapter. I think it should throw an error (spec not found) in this case just like we will fail the test if a property is missing (In this case all fields are missing 😆 yet the test passed)

dblock:
> > The tests here are failing though, so you'll need to add a fix.
> 
> I saw that. I'm on it. It also reveals a potential bug in the test framework: When the response schema is not defined in the spec (At least the framework thought so in this case because the schema was referenced in the wrong place), it skips response payload test and still passes the chapter. I think it should throw an error (spec not found) in this case just like we will fail the test if a property is missing (In this case all fields are missing 😆 yet the test passed)

Feel free to open this one separately and I can take care of it.

nhtruong:
@dblock I think I've found another [bug](https://github.com/opensearch-project/opensearch-api-specification/actions/runs/10357629378/job/28670062845?pr=489#step:7:66):
> Skipped because version 2.17.0 does not satisfy >= 2.11, < 3.0.

spec/namespaces/ml.yaml
```diff
@@ -130,36 +130,52 @@ components:
...
       content:
         application/json:
-          $ref: '../schemas/ml._common.yaml#/components/schemas/ModelGroup'
+          schema:
+            $ref: '../schemas/_common.yaml#/components/schemas/WriteResponseBase'
     ml.register_model@200:
       content:
         application/json:
-          $ref: '../schemas/ml._common.yaml#/components/schemas/Task'
+          schema:
```
dblock, line 154: Is this not a task? In tests we register a model then wait it the task to complete.

dblock (changes requested):
See below.

spec/namespaces/ml.yaml
nhtruong: The response is not the entire Task object, which is different from `get_task`'s response. This only returns the task_id for you to check on it later.

PR #493, opened 2024-08-12
Request sort in rest_total_hits_as_int tests

awick:
### Description
Request sort in queries since @dblock said some tests were failing because result order wasn't guaranteed. Had to add a new order field to sort against. Could have made one of the current fields a keyword, but would have required setting a mapping.

### Issues Resolved
N/A


dblock:
I would have used different movies and sorted by year to be frugal :)

dblock (approved):

PR #497, opened 2024-08-12
Fix: correctly named folders/files.

dblock:
### Description

Some leftovers from https://github.com/opensearch-project/opensearch-api-specification/pull/439.

Refresh indexes to avoid flake in observability testing. Closes #495. 

With https://github.com/opensearch-project/opensearch-api-specification/pull/498 CI will fail without these changes.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #498, opened 2024-08-12
Skip only known or hidden files.

dblock:
### Description

One of the unfortunate side effects of https://github.com/opensearch-project/opensearch-api-specification/pull/473 is that it's very easy to miss a test file that is misnamed. For example in https://github.com/opensearch-project/opensearch-api-specification/pull/497 we have .yml, or files that aren't yml at all but get skipped.

In this PR we only skip known and hidden things.

cc: @awick 


awick:
lol, how I first had it :)

dblock:
> lol, how I first had it :)

Amen.

Xtansia (approved):

Issue #500, opened 2024-08-12
[BUG][TEST-FRAMEWORK] Version selector not handling patch version correctly


nhtruong:
A test story was [skipped](https://github.com/opensearch-project/opensearch-api-specification/actions/runs/10357629378/job/28670062845?pr=489#step:7:66) because:
> Skipped because version 2.17.0 does not satisfy >= 2.11, < 3.0.

This is most likely due to `2.17.0` was used instead of `2.17`.

dblock:
I'll take it.

dblock:
Looks like I already fixed it [here](https://github.com/opensearch-project/opensearch-api-specification/pull/483/files#diff-5fdfeeda9cf99e3d36e2e8c3fb434cd7389123d041be602c663fd30eaf149d1dR240) but I'll refactor it with explicit tests.

PR #501, opened 2024-08-12
Fix: semver comparisons for ranges.

dblock:
### Description

Fixes semver comparisons for ranges, using the default that the library supports (without ruby-style commas).
Raises an error when semver range is incorrectly specified (hence the new utils helpers).
Re-enables tests/default/ingest/pipeline.yaml that passes with the current 3.0.

### Issues Resolved

Closes https://github.com/opensearch-project/opensearch-api-specification/issues/500.

### DIFF
--- a/TESTING_GUIDE.md
+++ b/TESTING_GUIDE.md
@@ -195,11 +195,11 @@ You can also reuse output in payload expectations. See [tests/plugins/index_stat
 
 ### Managing Versions
 
-It's common to add a feature to the next version of OpenSearch. When adding a new API in the spec, make sure to specify `x-version-added`, `x-version-deprecated` or `x-version-removed`. Finally, specify a semver range in your test stories or chapters as follows.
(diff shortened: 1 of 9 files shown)


tools/tests/_utils/Semver.test.ts
```diff
@@ -0,0 +1,47 @@
...
+  })
+});
+
+describe('satisfies', () => {
+  it ('defaults', () => {
+    expect(semver.satisfies('', '>= 1.3, < 99.0')).toBe(true)
+  })
+
+  it ('semver', () => {
+    expect(semver.satisfies(semver.coerce('2.17.0'), '>= 1.3, < 99.0')).toBe(true)
```
Xtansia, line 27: Why do we need special comma handling when the correct way to write this according to the [semver package](https://docs.npmjs.com/cli/v6/using-npm/semver#ranges) is to just remove the comma: `>=1.3 <99.0`

dblock: Hmmmm, interesting. I actually didn’t know that’s correct. I’ll get rid of comma handling and make sure we raise errors if this is incorrectly specified.

dblock: Updated.

Xtansia (approved):

Issue #496, opened 2024-08-12
[BUG] Test Framework passing chapters with missing response bodies


nhtruong:
Spawning off of this [PR](https://github.com/opensearch-project/opensearch-api-specification/pull/489#issuecomment-2284467731)

When the response schema is NOT defined in the spec, the framework skips the response payload test and still passes the chapter. Even if the response is null or an empty object `{}`, it should still have to be explicitly defined and the framework should throw a `response schema not found` in this case.

PR #502, opened 2024-08-13
Require response bodies in tests.

dblock:
### Description

Require response bodies in tests.

### Issues Resolved

N/A


nhtruong:
https://github.com/opensearch-project/opensearch-api-specification/issues/496 refers to missing explicit response body definition in the spec. In the test we don't need to explicitly specify a response body per chapter because it's assumed to be the default `200 - application/json` as it applies to 99% of test cases.

dblock:
Oh I see. I removed "closes ..." from the PR, I think this still can be merged, yes?

PR #506, opened 2024-08-13
Add resolve API test

Naarcha-AWS:
### Description

Adds the test for the `_resolve` endpoint in the current index.yaml test. 

### DIFF
--- a/tests/default/indices/index.yaml
+++ b/tests/default/indices/index.yaml
@@ -85,4 +85,15 @@ chapters:
     parameters:
       index: books,games
       cluster_manager_timeout: 10s
+
+  - synopsis: Check if the index `books` exists. It should.
+    path: /_resolve/index/books
+    method: GET
+    parameters:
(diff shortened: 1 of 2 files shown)


Naarcha-AWS:
Curious if I put this test in the right location. Here is the endpoint in the spec:

https://github.com/opensearch-project/opensearch-api-specification/blob/09824d0d0913a27ed9deeedf71fa4fffb1bed8ff/spec/namespaces/indices.yaml#L573

dblock (changes requested):
Lets put this in its own path, `tests/default/indices/resolve.yaml`.

The `path` in the test will have to be `/_resolve/index/{name}` as in the spec and `{name}` will come in `parameters`. Pre-create the index in a `prologues` and delete it in `epilogues`. Also test that a response to an index that doesn't exist is a 404.

dblock (approved):

PR #510, opened 2024-08-14
Added distributions/included and excluded.

dblock:
### Description

Following up on #483, added `distributions` with `included` and `excluded`. I prefer to manage exclusions in tests so that a potential new distribution can run this with all tests enabled the first time, which makes it possible to immediately see what is broken. 

Removed `undefined` values from results, and use an exact comparison in integration tests, making it much easier to see what scenarios failed.

Passing `npm run test:spec -- --opensearch-distribution=amazon-managed` against Amazon Managed OpenSearch 2.13. The values of `ssl_provider_http`, `ssl_cipher` and `ssl_protocol` are coming back `null` from `sslinfo` and most security APIs are disabled. 

### DIFF
--- a/spec/namespaces/cat.yaml
+++ b/spec/namespaces/cat.yaml
@@ -291,6 +291,9 @@ paths:
       operationId: cat.nodeattrs.0
       x-operation-group: cat.nodeattrs
       x-version-added: '1.0'
+      x-distributions-excluded:
+        - amazon-managed
+        - amazon-serverless
       description: Returns information about custom node attributes.
       externalDocs:
         url: https://opensearch.org/docs/latest/api-reference/cat/cat-nodeattrs/
--- a/TESTING_GUIDE.md
+++ b/TESTING_GUIDE.md
@@ -228,13 +228,16 @@ OpenSearch consists of plugins that may or may not be present in various distrib
     description: Returns basic information about the cluster.
 ```
 
-Similarly, skip tests that are not applicable to a distribution by listing the distributions that support it.
+Similarly, skip tests that are not applicable to a distribution by listing the distributions that support or do not support it.
 
 ```yaml
 description: Test root endpoint.
 distributions:
-  - amazon-managed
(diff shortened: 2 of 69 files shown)


Xtansia (approved):

spec/namespaces/cat.yaml
```diff
@@ -291,6 +291,9 @@ paths:
       operationId: cat.nodeattrs.0
       x-operation-group: cat.nodeattrs
       x-version-added: '1.0'
+      x-distributions-excluded:
```
reta, line 294: @dblock  coming from https://github.com/opensearch-project/opensearch-java/pull/1152#discussion_r1726911495, I understand the bias towards supporting AWS, but I honestly believe the spec has to be vendor neutral. I believe it should be feasible to keep the distribution specific details outside of the spec and allow the distribution specific spec to be produced if needed

dblock: I think you can/should open an issue copy-pasting the above about removing x-distribution-excluded.

When I introduced this feature I did not intend it to be biased towards supporting AWS in any way, but I did want to run the tests against Amazon Managed so I annotated some APIs. We would welcome other keywords here, be it `oracle-oci`, `aiven-managed` or something platform-specific like `community-freebsd` in the short term. Therefore I believe the spec *is* vendor netural today. I *think* it's ok to mention vendors in the spec this way, but we can talk about it. 

What you are looking at is the source for the spec, not the spec. We currently publish the all-versions spec in a [release](https://github.com/opensearch-project/opensearch-api-specification/releases). This repo is capable of producing a version-specific spec (e.g. 1.3.x, without APIs added in/after 2.0 for example or vice-versa) (see https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md#spec-merger), and that could be extended to produce a vendor-specific spec like (e.g. `-amazon-managed.yaml`).

PR #513, opened 2024-08-15
Added detailed test coverage report.

dblock:
### Description

Added a new `--coverage-report` option that displays a hierarchy of paths that are not tested. This is helpful to see if an entire API is tested vs. what's missing.

For example, the `/_alias` API is missing 4 tests and `/_plugins` are missing 46, including 11 in `/_plugins/_knn`.

```
  /_alias (4)
    GET /_alias
   /{name} (3)
     GET /_alias/{name}
     POST /_alias/{name}
     HEAD /_alias/{name}
```

```
/_plugins (46)
   /_knn (11)
```

### DIFF
--- a/TESTING_GUIDE.md
+++ b/TESTING_GUIDE.md
@@ -16,6 +16,9 @@
     - [Warnings](#warnings)
       - [multiple-paths-detected](#multiple-paths-detected)
       - [Suppressing Warnings](#suppressing-warnings)
+  - [Collecting Test Coverage](#collecting-test-coverage)
+    - [Coverage Summary](#coverage-summary)
+    - [Coverage Report](#coverage-report)
 <!-- TOC -->
 
 # Spec Testing Guide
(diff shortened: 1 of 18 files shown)


TESTING_GUIDE.md
```diff
@@ -318,3 +321,32 @@ The test runner may generate warnings that can be suppressed with `warnings:`. F
...
+## Collecting Test Coverage
+
+### Coverage Summary
+
+The test tool can generate a test coverage summary using `--coverage <path>` with the number of evaluated verb + path combinations, a total number of paths and the % of paths tested. 
+
+```json
+{
+  "evaluated_paths_count": 214,
+  "paths_count": 550,
```
nhtruong, line 334: Nit: "paths_count" is ambiguous. Can we do `total_paths_count` or `exposed_paths_count`?

dblock: Possibly, but this is existing code, so let's not scope creep it.

The spec test coverage is defined as follows.

```
export interface SpecTestCoverage {
  paths_count: number
  evaluated_paths_count: number,
  evaluated_paths_pct: number
}
```

In this context the paths count represents the spec, so adding `spec_` would duplicate these semantics.

tools/src/tester/ResultLogger.ts
```diff
@@ -43,7 +44,30 @@ export class ConsoleResultLogger implements ResultLogger {
...
   log_coverage(results: TestResults): void {
     console.log()
-    console.log(`Tested ${results.evaluated_paths_count()}/${results.spec_paths_count()} paths.`)
+    console.log(`Tested ${results.evaluated_paths().length}/${results.spec_paths().length} paths.`)
+  }
+
+  log_coverage_report(results: TestResults): void {
+    console.log()
+    console.log(`${results.unevaluated_paths().length} paths remaining.`)
+    const groups = _.groupBy(results.unevaluated_paths(), (path) => path.split(' ', 2)[1].split('/')[1])
```
nhtruong, line 53: I was wondering where the space `' '`  character came from because an URL path doesn't contain any space character. Then I saw [this](https://github.com/opensearch-project/opensearch-api-specification/pull/513/files#diff-4535a96ee24bf0e5cabb704065f4bff83e72f485127bcc45184d72b674a7942aR45). The combination of an http verb and an url path constitutes an API operation. I think we should name these entities `unevaluated_operations` and such to avoid confusion.

dblock: Ok, that's a good suggestion, I'll do a bulk rename.

dblock: I've refactored the path parsing into an `Operation` object, too.

nhtruong (approved):

PR #518, opened 2024-08-19
Add info_win.yaml (cloned from info.yaml). Can be used in Windows env

marknyny:
### Description
Added the initial step for setting up an env. for running test.
Added a new info_win.yaml (identical into.yaml) test file for Windows

### Issues Resolved
lets run tests in windows:
'''
npm run test:spec -- --opensearch-insecure --tests tests/default/_core/info_win.yaml

> ts-node tools/src/tester/test.ts --opensearch-insecure --tests tests/default/_core/info_win.yaml

OpenSearch 2.16.0

FAILED  info_win.yaml (C:\Users\marka\Projects\OpenSearch\runTests\opensearch-api-specification\tests\default\_core\info_win.yaml)
    FAILED  CHAPTERS
        FAILED  Get server info. (Operation "GET \" not found in the spec.)
        FAILED  Get server info (pretty=false). (Operation "GET \" not found in the spec.)

Tested 0/550 paths.
'''

### DIFF
--- a/TESTING_GUIDE.md
+++ b/TESTING_GUIDE.md
@@ -29,6 +29,16 @@ We have devised our own test framework to test the spec against an OpenSearch cl
 
 ### Running Spec Tests Locally
 
+Downloading and installing latest version of Node.js and npm
+
+https://docs.npmjs.com/downloading-and-installing-node-js-and-npm
+
+```bash
+Note: to download the latest version of npm, on the command line, run the following command:
+
+npm install -g npm
+```
+
 Set up an OpenSearch cluster with Docker:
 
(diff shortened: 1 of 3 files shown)


dblock (changes requested):
We definitely don't want to duplicate all tests between Windows and *nix. The tooling needs to support both. 

To add Windows support, I recommend starting by running unit tests for tools with `npm run test:unit` and fixing any failing tests on Windows. Then `npm run test:integ` to fix integration tests.

dblock:
Closing this because we definitely don't want to duplicate all our tests for Windows. Would like a PR that adds Windows support with CI instead.

PR #520, opened 2024-08-21
Fix tasks namespace

Xtansia:
### Description

### Issues Resolved

### DIFF
--- a/spec/schemas/_common.yaml
+++ b/spec/schemas/_common.yaml
@@ -1415,25 +1415,6 @@ components:
       type: string
       enum:
         - auto
-    BulkIndexByScrollFailure:
-      type: object
-      properties:
-        cause:
-          $ref: '#/components/schemas/ErrorCause'
-        id:
-          $ref: '#/components/schemas/Id'
-        index:
-          $ref: '#/components/schemas/IndexName'
-        status:
-          type: number
-        type:
-          type: string
-      required:
-        - cause
-        - id
-        - index
-        - status
-        - type
     Retries:
       type: object
       properties:
@@ -1960,6 +1941,8 @@ components:
         - remote_cluster_client
         - transform
         - voting_only
+      x-deprecated-enums:
+        - master
     HttpHeaders:
       type: object
       additionalProperties:
@@ -1990,11 +1973,7 @@ components:
         transport_address:
           $ref: '#/components/schemas/TransportAddress'
       required:
-        - attributes
-        - host
-        - ip
         - name
-        - transport_address
     RankContainer:
       type: object
       properties:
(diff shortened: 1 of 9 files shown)


spec/schemas/_common.yaml
```diff
@@ -1927,6 +1908,8 @@ components:
         - remote_cluster_client
         - transform
         - voting_only
+      x-deprecated-enums:
```
dblock, line 1911: The problem with `x-deprected-enums` is that it doesn't express since when. Would this be possible?

```
- value: master
   x-deprecated-since: '2.0'
```

dblock (changes requested):

spec/schemas/_common.yaml
Xtansia: It's unfortunate that `enum:` only accepts a string list, so we can't properly dedupe the values and associate info directly. We could do something like:
```yaml
    NodeRole:
      type: string
      enum:
        - client
        - cluster_manager
        - coordinating_only
        - data
        - data_cold
        - data_content
        - data_frozen
        - data_hot
        - data_warm
        - ingest
        - master
        - ml
        - remote_cluster_client
        - transform
        - voting_only
      x-deprecated-enums:
        master:
          message: 'Please use inclusive language.'
          since: '2.0'
```

Or if we want to generalize adding any extensions against specific values:
```yaml
    NodeRole:
      type: string
      enum:
        - client
        - cluster_manager
        - coordinating_only
        - data
        - data_cold
        - data_content
        - data_frozen
        - data_hot
        - data_warm
        - ingest
        - master
        - ml
        - remote_cluster_client
        - transform
        - voting_only
      x-enum:
        cluster_manager:
          x-version-added: '2.0'
        master:
          x-deprecation-message: 'Please use inclusive language.'
          x-version-deprecated: '2.0'
```

Obviously would ideally need a validator/linter for it to ensure values match up etc.

dblock: Don't we transform these parts of the spec in the merger and could have special treatment for enums?

Otherwise I think x-enum is a better and more extensible idea.

nhtruong: The `x-vendor-extensions` are only applicable to OpenAPI properties, not JSON Schemas. Further more, I don't think this level of granularity is not necessary, in my opinion. We can add `description` property for the schema that says `The use of "master" as a node role has been deprecated. Use "cluster_manager" instead`

nhtruong: I also think that this repo also needs branches for major versions, as it will get more and more complex to keep track of all the diffs between major versions in the same spec.

dblock: @nhtruong Are you saying this PR is good as is? I'm ok with it, feel free to hit approve.

Xtansia: @nhtruong The `x-` are allowed on all schema objects according to OAS 3.1 (because "OpenAPI properties" are just schemas): https://spec.openapis.org/oas/v3.1.0.html#fixed-fields-19

The main reasoning for this is that we annotate these values as `@Deprecated` in Java (or `[Obsolete]` in .NET), so just putting it in the description is unhelpful.

dblock: I think we should do https://stackoverflow.com/a/78232576.

tools/src/_utils/JsonSchemaValidator.ts
```diff
@@ -38,7 +38,13 @@ export default class JsonSchemaValidator {
     addFormats(this.ajv);
     if (options.ajv_errors_opts != null) ajv_errors(this.ajv, options.ajv_errors_opts)
     for (const keyword of options.additional_keywords ?? []) this.ajv.addKeyword(keyword)
-    Object.entries(options.reference_schemas ?? {}).forEach(([key, schema]) => this.ajv.addSchema(schema, key))
+    Object.entries(options.reference_schemas ?? {}).forEach(([key, schema]) => {
+      try {
+        this.ajv.addSchema(schema, key);
+      } catch (e) {
+        throw new Error(`Failed to add schema ${key}: \`${JSON.stringify(schema)}\``, { cause: e })
```
dblock, line 45: Could use a test.

dblock (approved):
Looks great!

dblock:
@nhtruong you good with the enum solution?

dblock (approved):

PR #524, opened 2024-08-22
Added schema for cluster routing awareness.

dblock:
### Description

Added schema for cluster routing awareness.

I also tried to implement a test for `/_cluster/decommission/awareness/{awareness_attribute_name}/{awareness_attribute_value}` and so far failed.

```yaml
$schema: ../../../../json_schemas/test_story.schema.yaml

description: Test cluster decommission operations.
prologues:
  - path: /_cluster/settings
    method: PUT
    request:
      payload:
        transient:
          cluster.routing.allocation.awareness.attributes: zone
          cluster.routing.allocation.awareness.force.zone.values:
            - zone-a
            - zone-b
  - path: /_cluster/routing/awareness/zone/weights
    method: PUT
    request:
      payload:
        weights:
          zone-a: '0'
          zone-b: '1'
        _version: 1
chapters:
  - synopsis: Decommission zone-a.
    path: /_cluster/decommission/awareness/{awareness_attribute_name}/{awareness_attribute_value}
    method: PUT
    parameters:
      awareness_attribute_name: zone
      awareness_attribute_value: zone-a
    response:
      status: 200
```


tools/src/tester/ChapterOutput.ts
```diff
@@ -36,6 +37,9 @@ export class ChapterOutput {
           return { evaluation: { result: Result.ERROR, message: 'No payload found in response, but expected output: ' + path } }
         }
         value = _.get(response, path)
+        const rhs = path.replaceAll(' ', '').split('?', 2)
+        value = _.get(response, rhs[0])
+        if (value === undefined && rhs[1] !== undefined) value = YAML.parse(rhs[1])
```
nhtruong, line 42: What's `rhs`? 🤔

nhtruong, line 42: What's `rhs`? 🤔

tests/routing/cluster/decommission/awareness.yaml
```diff
@@ -0,0 +1,45 @@
+$schema: ../../../../json_schemas/test_story.schema.yaml
+
+description: Test cluster routing settings.
+version: '>= 2.16'
+prologues:
+  - id: weights
+    path: /_cluster/routing/awareness/zone/weights
+    method: GET
+    output:
+      version: payload._version ? -1
```
nhtruong, line 10: This works fine for simple default values, but not scalable for more complex values, and it's not obviously if `-1` is a number or a string, or if `true` is a truthful bool or `"true"`. To accommodate this, we should allow the output definition to be either a string (which holds the retrieval path from the response_payload like how it's currently used), or an object:
```yaml
output:
  version:
    path: payload._version
    default: -1
```
This also makes the parsing logic a lot simpler.

Side tracking: does `payload._version` hint that the output values can come from something outside of the response payload?

nhtruong (changes requested):

tools/src/tester/ChapterOutput.ts
dblock: right hand side

tests/routing/cluster/decommission/awareness.yaml
dblock: Yeah, not sure why I didn't do this from the start, thanks.

> does payload._version hint that the output values can come from something outside of the response payload?

Unfortunately not. If you assumed that the cluster was just started that the value is always -1, but the problem is that when testing locally these tests would fail on the second run.

dblock:
@nhtruong Care to take another look please?

nhtruong (approved):

Issue #535, opened 2024-08-27
[BUG] KnnQuery radial search spec inconsistency


alex-keeler:
### What is the bug?
The KnnQuery spec (located at schemas/_common.yaml#/components/schemas/KnnQuery) is inconsistent with the current implementation of the knn query, specifically in regards to radial search. It specifies a `similarity` parameter, whereas the implementation accepts a `min_score` or `max_distance` parameter to achieve the same end.

### Do you have any additional context?
Radial search docs: https://opensearch.org/docs/latest/search-plugins/knn/radial-search-knn/
Original issue: https://github.com/opensearch-project/k-NN/issues/814

PR #538, opened 2024-08-28
Update knn query spec

alex-keeler:
### Description
Updates the knn query spec to match how it is currently implemented in OpenSearch.

### Issues Resolved
Closes https://github.com/opensearch-project/opensearch-api-specification/issues/535

### DIFF
--- a/spec/schemas/_common.yaml
+++ b/spec/schemas/_common.yaml
@@ -591,23 +591,19 @@ components:
       type: number
     EmptyObject:
       type: object
-    KnnQuery:
+    KnnField:
       type: object
       properties:
-        field:
-          $ref: '#/components/schemas/Field'
-        query_vector:
+        vector:
           $ref: '#/components/schemas/QueryVector'
-        query_vector_builder:
-          $ref: '#/components/schemas/QueryVectorBuilder'
         k:
           description: The final number of nearest neighbors to return as top hits
           type: number
-        num_candidates:
-          description: The number of nearest neighbor candidates to consider per shard
+        min_score:
+          description: The minimum similarity score for a neighbor to be considered a hit
           type: number
-        boost:
-          description: Boost value to apply to kNN scores
+        max_distance:
+          description: The maximum physical distance in vector space for a neighbor to be considered a hit
           type: number
         filter:
           description: Filters for the kNN search query
@@ -616,34 +612,15 @@ components:
             - type: array
               items:
                 $ref: '_common.query_dsl.yaml#/components/schemas/QueryContainer'
-        similarity:
-          description: The minimum similarity for a vector to be considered a match
+        boost:
+          description: Boost value to apply to kNN scores
           type: number
       required:
-        - field
-        - k
-        - num_candidates
+        - vector
     QueryVector:
       type: array
       items:
         type: number
-    QueryVectorBuilder:
-      type: object
-      properties:
-        text_embedding:
-          $ref: '#/components/schemas/TextEmbedding'
-      minProperties: 1
-      maxProperties: 1
-    TextEmbedding:
-      type: object
-      properties:
-        model_id:
-          type: string
-        model_text:
-          type: string
-      required:
-        - model_id
-        - model_text
     SlicedScroll:
       type: object
       properties:
--- a/spec/schemas/_common.yaml
+++ b/spec/schemas/_common.yaml
@@ -597,16 +597,18 @@ components:
         vector:
           $ref: '#/components/schemas/QueryVector'
         k:
-          description: The final number of nearest neighbors to return as top hits
+          description: The final number of nearest neighbors to return as top hits.
           type: number
         min_score:
-          description: The minimum similarity score for a neighbor to be considered a hit
+          description: The minimum similarity score for a neighbor to be considered a hit.
           type: number
+          x-version-added: '2.14'
         max_distance:
-          description: The maximum physical distance in vector space for a neighbor to be considered a hit
+          description: The maximum physical distance in vector space for a neighbor to be considered a hit.
           type: number
+          x-version-added: '2.14'
         filter:
-          description: Filters for the kNN search query
+          description: Filters for the kNN search query.
           oneOf:
             - $ref: '_common.query_dsl.yaml#/components/schemas/QueryContainer'
             - type: array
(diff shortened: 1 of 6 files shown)


alex-keeler:
The change was a bit more involved than I had originally anticipated.
Based off what I can tell, it appears the existing spec was ripped from the equivalent Elasticsearch spec. The OpenSearch implementation differs pretty significantly from this spec -- notably that the `fields` parameter has been replaced with a nested object, which uses the field name as the key. This nested object is what contains the rest of the parameters. Such behavior provided some challenges in designing the OpenAPI spec, so if there's a better alternative to my solution I'd be happy to update it. I also split the API between the KnnQuery and a new KnnField component to try and simplify the nested object business -- that can also be consolidated if deemed necessary.

spec/schemas/_common.yaml
```diff
@@ -591,23 +591,19 @@ components:
...
           $ref: '#/components/schemas/QueryVector'
-        query_vector_builder:
-          $ref: '#/components/schemas/QueryVectorBuilder'
         k:
           description: The final number of nearest neighbors to return as top hits
           type: number
-        num_candidates:
-          description: The number of nearest neighbor candidates to consider per shard
+        min_score:
+          description: The minimum similarity score for a neighbor to be considered a hit
```
dblock, line 603: Nit: add periods at the end of these descriptions.

dblock (changes requested):
👏 great work

- Update CHANGELOG.
- Annotate the APIs with `x-version-added` and such, 1.3.x tests fail with "[knn] query does not support [min_score])" for example.

@nhtruong @Xtansia might have some suggestions around the spec itself.

dblock (dismissed):

dblock:
For lints, probably just easier to sort values.

```
/home/runner/work/opensearch-api-specification/opensearch-api-specification/tests/default/_core/search/knn.yaml
Error:    36:31  error  Expected sequence values to be in ascending order. '2.3' should be before '3.5'  yml/sort-sequence-values
```

Tests also need to be limited to `>= 2.14` with `version:`, it's not smart enough to know that you're trying to test APIs that were added later to skip them.

dblock:
Looks like the filter query doesn't work in these old versions either. Not sure whether that's expected.

```
[INFO] => POST /movies/_search ({}) [application/json] {
  "query": {
    "knn": {
      "embedding": {
        "vector": [
          1.4,
          2.3,
          3.5,
          4.1,
          9.2
        ],
        "k": 1,
        "filter": {
          "term": {
            "year": 2012
          }
        }
      }
    }
  }
}
[INFO] <= 400 (application/json) | {
  "root_cause": [
    {
      "type": "parsing_exception",
      "reason": "[knn] unknown token [START_OBJECT] after [filter]",
      "line": 1,
      "col": 77
    }
  ],
  "type": "parsing_exception",
  "reason": "[knn] unknown token [START_OBJECT] after [filter]",
  "line": 1,
  "col": 77
}
```

You can reproduce these failures locally pretty easily btw.

```
OPENSEARCH_VERSION=1.3.7 docker compose up
```

```
npm run test:spec--insecure -- --tests tests/default/_core/search/knn.yaml --verbose
```

alex-keeler:
Thanks for the heads up, I've only been testing against 2.16.0 locally. I'll go through and make sure all is well with earlier releases. I'm sure the failures are as annoying for you as they are for me.
I'll have to do some digging to see when the filter parameter was added. I didn't realize it came in a later release (I'm assuming, anyway).

dblock:
@navneet1v from k-nn might be interested in this and can help make sure things are properly version-annotated

navneet1v:
> Looks like the filter query doesn't work in these old versions either. Not sure whether that's expected.
> 
> ```
> [INFO] => POST /movies/_search ({}) [application/json] {
>   "query": {
>     "knn": {
>       "embedding": {
>         "vector": [
>           1.4,
>           2.3,
>           3.5,
>           4.1,
>           9.2
>         ],
>         "k": 1,
>         "filter": {
>           "term": {
>             "year": 2012
>           }
>         }
>       }
>     }
>   }
> }
> [INFO] <= 400 (application/json) | {
>   "root_cause": [
>     {
>       "type": "parsing_exception",
>       "reason": "[knn] unknown token [START_OBJECT] after [filter]",
>       "line": 1,
>       "col": 77
>     }
>   ],
>   "type": "parsing_exception",
>   "reason": "[knn] unknown token [START_OBJECT] after [filter]",
>   "line": 1,
>   "col": 77
> }
> ```
> 
> You can reproduce these failures locally pretty easily btw.
> 
> ```
> OPENSEARCH_VERSION=1.3.7 docker compose up
> ```
> 
> ```
> npm run test:spec--insecure -- --tests tests/default/_core/search/knn.yaml --verbose
> ```

The filter query specifically provided here doesn't work for old version because filters works for 2.4 and above.

alex-keeler:
> > Looks like the filter query doesn't work in these old versions either. Not sure whether that's expected.
> > ```
> > [INFO] => POST /movies/_search ({}) [application/json] {
> >   "query": {
> >     "knn": {
> >       "embedding": {
> >         "vector": [
> >           1.4,
> >           2.3,
> >           3.5,
> >           4.1,
> >           9.2
> >         ],
> >         "k": 1,
> >         "filter": {
> >           "term": {
> >             "year": 2012
> >           }
> >         }
> >       }
> >     }
> >   }
> > }
> > [INFO] <= 400 (application/json) | {
> >   "root_cause": [
> >     {
> >       "type": "parsing_exception",
> >       "reason": "[knn] unknown token [START_OBJECT] after [filter]",
> >       "line": 1,
> >       "col": 77
> >     }
> >   ],
> >   "type": "parsing_exception",
> >   "reason": "[knn] unknown token [START_OBJECT] after [filter]",
> >   "line": 1,
> >   "col": 77
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > You can reproduce these failures locally pretty easily btw.
> > ```
> > OPENSEARCH_VERSION=1.3.7 docker compose up
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ```
> > npm run test:spec--insecure -- --tests tests/default/_core/search/knn.yaml --verbose
> > ```
> 
> The filter query specifically provided here doesn't work for old version because filters works for 2.4 and above.

Thank you for the info. I ended up restraining the filter test to >= 2.9, as it appears the faiss engine didn't get filter support until then. I tried switching the embedding to lucene, but found out that lucene knn support wasn't added until version 2.2, so I figured it was better to keep things as is. For extra safety, I also constrained the entire test to >= 1.2, which is when faiss k-NN support was added.

dblock (approved):

dblock:
Thanks so much for your help @alex-keeler. Please don't hesitate to pickup other missing/potentially incorrect parts of the search DSL, that would be much appreciated!

Issue #423, opened 2024-07-16
[BUG] Remove non-humanly-readable fields.


dblock:
### What is the bug?

Coming from https://github.com/opensearch-project/opensearch-go/issues/284#issuecomment-2230838482, the spec contains both humanly readable and non-humanly-readable versions of many fields. Fields of type `_common:BysteSize` are no longer returned by the API as they got replaced by the same property suffixed with `_in_bytes` pointing to type number.

The core engine seems to do something clever about it. [For example](https://github.com/opensearch-project/OpenSearch/blob/6ad57bf393bda06da1c463d7c276d4a61d3cb299/server/src/main/java/org/opensearch/index/merge/MergeStats.java#L247). These are added with [`humanlyReadableField`](https://github.com/opensearch-project/OpenSearch/blob/main/libs/core/src/main/java/org/opensearch/core/xcontent/XContentBuilder.java#L972). It looks like there still may be cases where you have a non-humanly-readable field, or are there?

### What is the expected behavior?

Figure out whether non-humanly-readable fields are still possible, delete them.

Jakob3xD:
Did some research and found out, that there is a `human` parameter. When this parameter is set it also returns the the memory_size field. The question now is, if the string can also be something else then a string if there is some sort of unit parameter. Therefore the fields are valid but we should look if they type can be more specific then oneOf string, number. 

```
GET <some-index>/_stats?human
      "fielddata": {
        "memory_size": "0b",
        "memory_size_in_bytes": 0,
        "evictions": 0
      }
```
```
GET <some-index>/_stats
      "fielddata": {
        "memory_size_in_bytes": 0,
        "evictions": 0
      }
```

Jakob3xD:
If I understand the java code correctly than the human fields should be of type string all the time.
At least for fielddata.
The value and both fields are passed to the `builder.humanReadableField()` function.
https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/fielddata/FieldDataStats.java#L121
The `builder.humanReadableField()` function checks if human parameter is true and converts the value into a string.
https://github.com/opensearch-project/OpenSearch/blob/main/libs/core/src/main/java/org/opensearch/core/xcontent/XContentBuilder.java#L976

What I don't understand is what is happening with the rawValue as I don't understand the java code.
[...]

PR #552, opened 2024-09-04
Adjust component for Human readable Bytes

Jakob3xD:
### Description
- Add human parameter to some test
- Add new component type for human readable bytes
- Replace Component referenced to new types
- Add missing fields

### Issues Resolved
Related to #423 
Closes https://github.com/opensearch-project/opensearch-api-specification/issues/554

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -89,6 +89,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Added `terminated_early` to `count@200` ([#547](https://github.com/opensearch-project/opensearch-api-specification/pull/547))
 - Added request and response schemas for `/_cluster/routing/awareness/{attribute}/weights` ([#524](https://github.com/opensearch-project/opensearch-api-specification/pull/524))
 - Added request and response schemas for `/_cluster/decommission/awareness` ([#524](https://github.com/opensearch-project/opensearch-api-specification/pull/524))
+- Added `Bytes` component of type number ([#552](https://github.com/opensearch-project/opensearch-api-specification/pull/552))
 
 ### Changed
 
@@ -104,6 +105,8 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Split test suite ([#472])(https://github.com/opensearch-project/opensearch-api-specification/pull/472)
 - Changed `WriteResponseBase`'s `_primary_term`, `_seq_no` & `_version` to have `int64` format ([#530](https://github.com/opensearch-project/opensearch-api-specification/pull/530))
 - Adjust indices, shards cat API to test against unassigned indices ([#551](https://github.com/opensearch-project/opensearch-api-specification/pull/551))
+- Rename `Bytes` component to `StorageType` ([#552](https://github.com/opensearch-project/opensearch-api-specification/pull/552))
+- Rename `ByteSize` to `StorageSize` ([#552](https://github.com/opensearch-project/opensearch-api-specification/pull/552))
 
 ### Deprecated
 
--- a/spec/schemas/_common.yaml
+++ b/spec/schemas/_common.yaml
@@ -686,7 +686,10 @@ components:
             $ref: '#/components/schemas/NodeId'
     NodeId:
       type: string
-    Bytes:
+    StorageSize:
+      type: string
+      pattern: '\d+(\.\d+)?(b|kb|k|mb|m|gb|g|tb|t|pb|p)'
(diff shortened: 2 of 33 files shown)


CHANGELOG.md
```diff
@@ -104,6 +105,8 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Split test suite ([#472])(https://github.com/opensearch-project/opensearch-api-specification/pull/472)
 - Changed `WriteResponseBase`'s `_primary_term`, `_seq_no` & `_version` to have `int64` format ([#530](https://github.com/opensearch-project/opensearch-api-specification/pull/530))
 - Adjust indices, shards cat API to test against unassigned indices ([#551](https://github.com/opensearch-project/opensearch-api-specification/pull/551))
+- Rename Bytes component to StorageType ([#552](https://github.com/opensearch-project/opensearch-api-specification/pull/552))
+- Rename ByteSize to StorageSize ([#552](https://github.com/opensearch-project/opensearch-api-specification/pull/552))
```
dblock, line 109:
```suggestion
- Rename `ByteSize` to `StorageSize` ([#552](https://github.com/opensearch-project/opensearch-api-specification/pull/552))
```

dblock (changes requested):
Typo in CHANGELOG, thanks!

dblock (approved):

dblock:
I have a fix for the flaky routing test in #556

spec/schemas/_common.yaml
```diff
@@ -686,7 +686,10 @@ components:
             $ref: '#/components/schemas/NodeId'
     NodeId:
       type: string
-    Bytes:
+    StorageSize:
+      type: string
+      pattern: '\d+(\.\d+)?(b|kb|k|mb|m|gb|g|tb|t|pb|p)'
+    StorageType:
```
Xtansia, line 692: This naming does not make sense to me. The name would imply this is types of storage (cold, warm, hdd, etc). When in actuality this is byte units.

spec/schemas/_common.yaml
```diff
@@ -686,7 +686,10 @@ components:
             $ref: '#/components/schemas/NodeId'
     NodeId:
       type: string
-    Bytes:
+    StorageSize:
```
Xtansia, line 689: Related to my comment on `StorageType`, this doesn't seem accurate to me, not all things measured in bytes are storage.

spec/schemas/_common.yaml
```diff
@@ -700,10 +703,8 @@ components:
         - pb
         - t
         - tb
-    ByteSize:
-      oneOf:
-        - type: number
-        - type: string
+    Bytes:
+      type: integer
```
Xtansia, line 707: This probably should have had a `format: int64` as well

spec/schemas/_common.yaml
Jakob3xD: What would you prefer? My suggestion would be `BytesHuman` or `BytesString` ?

spec/schemas/_common.yaml
Jakob3xD: What would you prefer? My suggestion would be `BytesUnit` ?

Xtansia: I think that's reasonable as we do have other `...Unit` enums.

spec/schemas/_common.yaml
Xtansia: I feel `BytesString` could be mis-interpreted as some mechanism for encoding raw-bytes as a string (ie. a hex/base64 string). Which also kind of applies to `Bytes` on it's own too.

Maybe a more verbose `ByteCount` and `HumanReadableByteCount` would make sense?

Xtansia: cc: @dblock @nhtruong ^^

dblock: The original Elastic docs call these "human-formatted", so I think we could use that term, `HumanFormattedBytes`, or since these appear a lot in the CAT API, which is "compact and aligned text", we could use `BytesText`. WDYT?

Jakob3xD: Just as a side note for CAT API.
Due to the `bytes` parameter the content differ. Without `bytes` parameter it would be printed as `"1gb"` and with it as `"1024"`. So it would not match the `StorageSize`. We would need an extra Type where the unit suffix is optional.

dblock: The type can have a `anyOf:` with different configurations, one of which is a string with all numbers?

Issue #555, opened 2024-09-05
[BUG] Reroute an index shard between nodes test is flaky


dblock:
### What is the bug?

Cluster rerouting test is flaky.

https://github.com/opensearch-project/opensearch-api-specification/actions/runs/10720682021/job/29731120869?pr=552

```
ERROR   cluster/reroute.yaml (/home/runner/work/opensearch-api-specification/opensearch-api-specification/tests/routing/cluster/reroute.yaml)
    ERROR   CHAPTERS
        PASSED  Shows the allocation of shards in a cluster.
            PASSED  REQUEST BODY
            PASSED  RESPONSE STATUS
            PASSED  RESPONSE PAYLOAD BODY
            PASSED  RESPONSE PAYLOAD SCHEMA
            SKIPPED RESPONSE OUTPUT VALUES
        PASSED  Shows the allocation of shards in a cluster with parameters.
            PASSED  PARAMETERS
                PASSED  dry_run
                PASSED  explain
                PASSED  metric
                PASSED  retry_failed
            PASSED  REQUEST BODY
            PASSED  RESPONSE STATUS
            PASSED  RESPONSE PAYLOAD BODY
            PASSED  RESPONSE PAYLOAD SCHEMA
            SKIPPED RESPONSE OUTPUT VALUES
        ERROR   Reroute an index shard between nodes.
            PASSED  REQUEST BODY
            ERROR   RESPONSE STATUS (Expected status 200, ...)
            SKIPPED RESPONSE PAYLOAD BODY
            SKIPPED RESPONSE PAYLOAD SCHEMA
            SKIPPED RESPONSE OUTPUT VALUES
```

### How can one reproduce the bug?

Try the test enough times.

```
cd tests/routing/
docker compose up
```

```
export OPENSEARCH_PASSWORD=myStrongPassword123!
npm run test:spec--insecure -- --tests tests/routing/cluster/reroute.yaml --verbose
```

### What is the expected behavior?

Test to pass consistently.

### What is your host/environment?

GHA

### Do you have any screenshots?

https://github.com/opensearch-project/opensearch-api-specification/actions/runs/10720682021/job/29731120869?pr=552

PR #556, opened 2024-09-05
Truncate errors at 256 characters.

dblock:
### Description

I was trying to debug #555 and couldn't reproduce it locally. I wish I saw the error in CI, but we're being too clever with truncation of the error message. Be less clever in this PR.


tests/routing/cluster/reroute.yaml
```diff
@@ -51,5 +63,8 @@ chapters:
               shard: 0
               from_node: opensearch-node1
               to_node: opensearch-node2
+    retry:
```
dblock, line 66: Shard may be still relocating, retry.

tools/src/tester/ResultLogger.ts
```diff
@@ -152,8 +152,9 @@ export class ConsoleResultLogger implements ResultLogger {
   }
 
   #maybe_shorten_error_message(message: string | undefined): string | undefined {
-    if (message === undefined || message.length <= 128 || this._verbose) return message
-    const part = message.split(',')[0]
+    const cut_at = 256
```
nhtruong, line 155: Nit: This is better off as a configurable number in the constructor, esp when we refactor the test framework to be a standalone product.

nhtruong (approved):

tools/src/tester/ResultLogger.ts
dblock: I'll leave it as is for now, but we can expose it as an option when someone actually wants to change this value programmatically.

PR #575, opened 2024-09-14
Ignore superseded operations in coverage reports.

dblock:
### Description

IMO we don't need to test superseded operations, exclude those from test coverage numbers and reports.

### DIFF
--- a/tools/src/tester/TestResults.ts
+++ b/tools/src/tester/TestResults.ts
@@ -47,11 +47,14 @@ export default class TestResults {
 
   operations(): Operation[] {
     if (this._operations !== undefined) return this._operations
-    this._operations = _.uniqWith(Object.entries(this._spec.paths()).flatMap(([path, path_item]) => {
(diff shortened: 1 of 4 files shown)


tools/src/tester/TestResults.ts
```diff
@@ -45,13 +46,30 @@ export default class TestResults {
...
+    this._ignorable_operations = _.uniqWith(_.compact(Object.entries(this._spec.spec().paths).flatMap(([path, ops]) => {
+      return Object.entries(ops as Record<string, any>).map(([method, spec]) => {
+        if (spec['x-ignorable'] === true) {
+          return { method: method.toUpperCase(), path }
+        }
+      })
+    })), isEqual)
+
+    return this._ignorable_operations
+  }
```
nhtruong, line 61: Where is this method used?

dblock: Was just leftoever from some older version, deleted.

nhtruong (approved):

PR #576, opened 2024-09-14
Added aggregation tests for avg, max, min, range, terms, sum, nested and histogram.

kkewwei:
### Description

### Issues Resolved

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -166,6 +166,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Fixed `/_plugins/_transform/_preview` ([#568](https://github.com/opensearch-project/opensearch-api-specification/pull/568))
 - Fixed create/delete/index operation in `_bulk` ([#582](https://github.com/opensearch-project/opensearch-api-specification/pull/582))
 - Add `mode` and `compression` to k-NN index creation and search, and add `rescore` and `oversample_factor` to k-NN search ([#588](https://github.com/opensearch-project/opensearch-api-specification/pull/588))
+- Fixed `/{index}/_search` with aggregations ([#576](https://github.com/opensearch-project/opensearch-api-specification/pull/576))
 
 ### Security
 
--- a/spec/schemas/_common.aggregations.yaml
+++ b/spec/schemas/_common.aggregations.yaml
@@ -7,7 +7,7 @@ paths: {}
 components:
   schemas:
     Aggregate:
-      oneOf:
+      anyOf:
         - $ref: '#/components/schemas/CardinalityAggregate'
         - $ref: '#/components/schemas/HdrPercentilesAggregate'
         - $ref: '#/components/schemas/HdrPercentileRanksAggregate'
@@ -786,6 +786,9 @@ components:
       allOf:
         - $ref: '#/components/schemas/SingleBucketAggregateBase'
         - type: object
+          properties:
+            doc_count:
+              type: number
     ReverseNestedAggregate:
       allOf:
         - $ref: '#/components/schemas/SingleBucketAggregateBase'
--- a/tests/default/_core/search/aggregations/avg.yaml
+++ b/tests/default/_core/search/aggregations/avg.yaml
@@ -1,0 +1,38 @@
+$schema: ../../../../../json_schemas/test_story.schema.yaml
+
+description: Test avg aggregation.
+prologues:
+  - path: /_bulk
+    method: POST
+    parameters:
+      refresh: true
+    request:
+      content_type: application/x-ndjson
+      payload:
+        - {create: {_index: movies}}
+        - {director: Bennett Miller, title: Moneyball, duration: 126}
+        - {create: {_index: movies}}
+        - {director: Nicolas Winding Refn, title: Drive, duration: 100}
+chapters:
+  - synopsis: Aggregate with avg.
+    path: /{index}/_search
+    parameters:
+      index: movies
+    method: POST
+    request:
+      payload:
+        size: 0
+        aggregations:
+          duration_avg:
+            avg:
+              field: duration
+    response:
+      status: 200
+      payload:
+        aggregations:
+          duration_avg:
+            value: 113
+epilogues:
+  - path: /movies
+    method: DELETE
+    status: [200, 404]
--- a/tests/default/_core/search/aggregations/historgram.yaml
+++ b/tests/default/_core/search/aggregations/historgram.yaml
@@ -1,0 +1,48 @@
+$schema: ../../../../../json_schemas/test_story.schema.yaml
+
+description: Test histogram aggregation.
+prologues:
+  - path: /_bulk
+    method: POST
+    parameters:
+      refresh: true
+    request:
+      content_type: application/x-ndjson
+      payload:
+        - {create: {_index: movies}}
+        - {director: Bennett Miller, title: Moneyball, year: 2011}
+        - {create: {_index: movies}}
+        - {director: Bennett Miller, title: The Cruise, year: 1998}
+        - {create: {_index: movies}}
+        - {director: Nicolas Winding Refn, title: Drive, year: 1960}
+chapters:
+  - synopsis: Aggregate with histogram.
+    path: /{index}/_search
+    parameters:
+      index: movies
+    method: POST
+    request:
+      payload:
+        size: 0
+        aggregations:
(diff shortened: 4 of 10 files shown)


CHANGELOG.md
```diff
@@ -113,6 +113,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Rename `Bytes` component to `StorageType` ([#552](https://github.com/opensearch-project/opensearch-api-specification/pull/552))
 - Rename `ByteSize` to `StorageSize` ([#552](https://github.com/opensearch-project/opensearch-api-specification/pull/552))
 - Split `Percentage` into `PercentageNumber`, `PercentageString` ([#573](https://github.com/opensearch-project/opensearch-api-specification/pull/573))
+- Add test about nested aggregation ([#576](https://github.com/opensearch-project/opensearch-api-specification/issues/576))
```
dblock, line 116: No need to include a CHANGELOG for tests.

tests/default/_core/aggregation/nested.yaml
```diff
@@ -0,0 +1,100 @@
+$schema: ../../../../json_schemas/test_story.schema.yaml
```
dblock, line 1: Let's map the file name to the query, so this should be `search/aggs/nested_agg.yaml`.

tests/default/_core/aggregation/nested.yaml
```diff
@@ -0,0 +1,100 @@
...
+
+description: Test aggregation endpoints with nested.
```
dblock, line 3: I'll nitpick, there's no such a thing as aggregation endpoint. 

```suggestion
description: Test aggregation.
```

tests/default/_core/aggregation/nested.yaml
```diff
@@ -0,0 +1,100 @@
...
+                  type: nested
+                  properties:
+                    c:
+                      type: nested
+                      properties:
+                        d:
+                          type: keyword
+
+chapters:
+  - synopsis: Bulk document.
```
dblock, line 26: Should go into prologue.

tests/default/_core/aggregation/nested.yaml
```diff
@@ -0,0 +1,100 @@
...
+  - path: movies
+    method: PUT
+    request_body:
+      payload:
+        mappings:
+          properties:
+            a:
+              type: nested
+              properties:
+                b1:
```
dblock, line 14: We plan to use these examples in the documentation. Is there some realistic agg scenario with movies? Maybe actors or something that would differentiate keywords like `nested` from data like `movie`.

dblock (changes requested):
Thanks!

Looks like tests failed on 2.0, possibly because the feature wasn't there, so the test needs to be annotated accordingly.

Run `npm run lint--fix` for the linter to be happy.

I have some more nitpicks below.

tests/default/_core/search/aggs/nested_agg.yaml
```diff
@@ -0,0 +1,96 @@
...
+                      type: nested
+                      properties:
+                        country:
+                          type: keyword
+  - path: /_bulk
+    method: POST
+    request_body:
+      content_type: application/x-ndjson
+      payload:
+        - {index: {_index: user_infos, _id: '0'}}
```
dblock, line 29: Could we keep with the theme of movies and books? We plan to use these in the documentation, trying to be a bit more consistent.

dblock (changes requested):
Some tests against 2.x failed, probably needs a `version: ` somewhere.

dblock:
```
ERROR   _core/search/aggs/nested_agg.yaml (Invalid Story: data/prologues/0 contains unsupported properties: request_body, request_body, request_body --- data/chapters/0/synopsis must match pattern "^\p{Lu}[\s\S]*\.$")
```

It's `request/payload`.

kkewwei:
@dblock  Thank you for your patient review, I'm not familiar with `opensearch-api-specification`, Is there any command I can use to test the change?  `npm run test` is ok?

dblock:
> @dblock Thank you for your patient review, I'm not familiar with `opensearch-api-specification`, Is there any command I can use to test the change? `npm run test` is ok?

Yes, do take a moment to read https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md and https://github.com/opensearch-project/opensearch-api-specification/blob/main/TESTING_GUIDE.md. I know it's a lot. 

We test an entire matrix of versions of OpenSearch, the 2.0 tests are failing. If you want to run those locally, it's basically starting a container with `OPENSEARCH_VERSION=2.0.0 docker compose up` from `tests/default` and then running tests with `OPENSEARCH_PASSWORD=admin npm run test:spec`.

kkewwei:
@dblock sorry for the latency reply(a bit busy), I will continue to follow up now.

dblock (changes requested):
Thanks! LMK when it's ready to be reviewed.

Some of this should work in 2.17, right?

kkewwei:
> Thanks! LMK when it's ready to be reviewed.
> 
> Some of this should work in 2.17, right?

@dblock thanks for you help, I know how to test.

When I test the aggregation, it throws the exception:
```
            FAILED  RESPONSE PAYLOAD SCHEMA (data/aggregations/title MUST contain the missing properties: value, values, values, values, values, values, value, value, value, value, value, value, value, value, value, value, keys, avg, count, max, min, sum, avg, count, max, min, sum, avg, count, max, m, ...)
```
It seems that I should define the response schema of the aggregation, I try a long time to define user-input `aggs1` but fails, can you help adding a example aggregation test, and let me learn?
image

dblock:
Given the search response:

```
    search@200:
      content:
        application/json:
          schema:
            $ref: '../schemas/_core.search.yaml#/components/schemas/ResponseBody'
```

it's a `ResponseBody` which defines `aggregations:`.

```
        aggregations:
          type: object
          additionalProperties:
            $ref: '_common.aggregations.yaml#/components/schemas/Aggregate'
```

if we replace this with just a vanilla object with any property (`additionalProperties: true`) the test will pass, so we're in the right place.

```
        aggregations:
          type: object
          additionalProperties: true
          # additionalProperties:
          #   $ref: '_common.aggregations.yaml#/components/schemas/Aggregate'
```

An `Aggregate` is one of many things:

```
    Aggregate:
      oneOf:
        - $ref: '#/components/schemas/CardinalityAggregate'
        - $ref: '#/components/schemas/HdrPercentilesAggregate'
        - ...
```

If we make the aggregate match our return value the spec also passes.

```
    Aggregate:
      oneOf:
        - type: object
          properties:
            doc_count:
              type: number
```

But if we add the other possibilities, it fails!

```
      oneOf:
        - type: object
          properties:
            doc_count:
              type: number
        - $ref: '#/components/schemas/CardinalityAggregate'
        - $ref: '#/components/schemas/HdrPercentilesAggregate'
```

This is where I've had a :facepalm: moment multiple times. Turns out that `oneOf` behaves like "only one of", not "any one of", so if more than one aggregate type matches, it will fail. And the schema validator is not smart enough to say "your object matches N items, but expected to only match 1".

The fix is to change `oneOf` by `anyOf`.

```
    Aggregate:
      anyOf:
        - $ref: '#/components/schemas/CardinalityAggregate'
        - $ref: '#/components/schemas/HdrPercentilesAggregate'
```

Note that each aggregation correctly uses `allOf`, combining `aggregateBase` and other properties of a specific aggregation. For example:

```
    PercentilesAggregateBase:
      allOf:
        - $ref: '#/components/schemas/AggregateBase'
        - type: object
          properties:
            values:
              $ref: '#/components/schemas/Percentiles'
          required:
            - values
```

https://github.com/dblock/opensearch-api-specification/commit/e497deadb4d4e13e6c0aa951382b4364698f7216

Rebase, then merge https://github.com/dblock/opensearch-api-specification/tree/kkewwei-add_nested_aggs onto your branch? I was on a plane and added some more agg tests on top.

dblock (approved):

dblock (changes requested):
Let's add a 1-liner to CHANGELOG, maybe "Fix `/{index}/_search` with `aggregations`"?

Link checker is failing not sure how to fix that ... unrelated but if you have an idea include that.

kkewwei:
"Link checker is failing" seems to be a probabilistic problem, I have no idea how to fix it.

dblock (approved):

Issue #224, opened 2024-04-11
[FEATURE] Add specs for `ism` namespace


nhtruong:
Part of [Adding missing API specs issue](https://github.com/opensearch-project/opensearch-api-specification/issues/168)

- [x] /_opendistro/_ism/add: ["POST"]
- [x] /_opendistro/_ism/add/{index}: ["POST"]
- [x] /_opendistro/_ism/change_policy: ["POST"]
- [x] /_opendistro/_ism/change_policy/{index}: ["POST"]
- [x] /_opendistro/_ism/explain: ["GET"]
- [x] /_opendistro/_ism/explain/{index}: ["GET"]
- [x] /_opendistro/_ism/policies: ["GET","PUT"]
- [x] /_opendistro/_ism/policies/{policyID}: ["GET","PUT","DELETE","HEAD"]
- [x] /_opendistro/_ism/remove: ["POST"]
- [x] /_opendistro/_ism/remove/{index}: ["POST"]
- [x] /_opendistro/_ism/retry: ["POST"]
- [x] /_opendistro/_ism/retry/{index}: ["POST"]
---
- [x] /_plugins/_ism/add: ["POST"]
- [x] /_plugins/_ism/add/{index}: ["POST"]
- [x] /_plugins/_ism/change_policy: ["POST"]
- [x] /_plugins/_ism/change_policy/{index}: ["POST"]
- [x] /_plugins/_ism/explain: ["GET"]
- [x] /_plugins/_ism/explain/{index}: ["GET"]
- [x] /_plugins/_ism/policies: ["GET","PUT"]
- [x] /_plugins/_ism/policies/{policyID}: ["GET","PUT","DELETE","HEAD"]
- [x] /_plugins/_ism/remove: ["POST"]
- [x] /_plugins/_ism/remove/{index}: ["POST"]
- [x] /_plugins/_ism/retry: ["POST"]
- [x] /_plugins/_ism/retry/{index}: ["POST"]
---
- [x] /_opendistro/_refresh_search_analyzers: ["POST"]
- [x] /_opendistro/_refresh_search_analyzers/{index}: ["POST"]
---
- [x] /_plugins/_refresh_search_analyzers: ["POST"]
- [x] /_plugins/_refresh_search_analyzers/{index}: ["POST"]

dblock:
I started working on this in https://github.com/opensearch-project/opensearch-api-specification/pull/568.

dblock:
[...]

PR #578, opened 2024-09-15
Completed ISM specs and tests.

dblock:
### Description

Completed ISM specs and tests.

### Issues Resolved

Closes https://github.com/opensearch-project/opensearch-api-specification/issues/224.

### DIFF
--- a/spec/namespaces/ism.yaml
+++ b/spec/namespaces/ism.yaml
@@ -33,7 +33,7 @@ paths:
     get:
       operationId: ism.get_policy.0
       x-operation-group: ism.get_policy
-      description: Creates or updates a policy.
+      description: Gets a policy.
       externalDocs:
         url: https://opensearch.org/docs/latest/im-plugin/ism/api/#put-policy
       parameters:
@@ -169,6 +169,8 @@ paths:
       description: Gets the currently applied policy on all indices.
       externalDocs:
         url: https://opensearch.org/docs/latest/im-plugin/ism/api/#explain-index
+      requestBody:
+        $ref: '#/components/requestBodies/ism.explain'
       responses:
         '200':
           $ref: '#/components/responses/ism.explain@200'
@@ -178,6 +180,8 @@ paths:
       description: Gets the currently applied policy on all indices.
       externalDocs:
         url: https://opensearch.org/docs/latest/im-plugin/ism/api/#explain-index
+      requestBody:
+        $ref: '#/components/requestBodies/ism.explain'
       responses:
         '200':
           $ref: '#/components/responses/ism.explain@200'
@@ -190,6 +194,21 @@ paths:
         url: https://opensearch.org/docs/latest/im-plugin/ism/api/#explain-index
       parameters:
         - $ref: '#/components/parameters/ism.explain_policy::path.index'
+      requestBody:
+        $ref: '#/components/requestBodies/ism.explain_policy'
+      responses:
+        '200':
+          $ref: '#/components/responses/ism.explain_policy@200'
+    post:
+      operationId: ism.explain_policy.1
+      x-operation-group: ism.explain_policy
+      description: Gets the currently applied policy on an index.
+      externalDocs:
+        url: https://opensearch.org/docs/latest/im-plugin/ism/api/#explain-index
+      parameters:
+        - $ref: '#/components/parameters/ism.explain_policy::path.index'
+      requestBody:
+        $ref: '#/components/requestBodies/ism.explain_policy'
       responses:
         '200':
           $ref: '#/components/responses/ism.explain_policy@200'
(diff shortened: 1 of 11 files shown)


spec/namespaces/ism.yaml
```diff
@@ -4,18 +4,56 @@ info:
...
+        - $ref: '#/components/parameters/ism.put_policies::query.policyID'
+      responses:
+        '200':
+          $ref: '#/components/responses/ism.put_policies@200'
   /_plugins/_ism/policies/{policy_id}:
     get:
       operationId: ism.get_policy.0
       x-operation-group: ism.get_policy
-      description: Gets the policy.
+      description: Creates or updates a policy.
```
nhtruong, line 36: the group says `ism.get_policy` but the description is Create or Update?

spec/namespaces/ism.yaml
```diff
@@ -4,18 +4,56 @@ info:
...
-        url: https://opensearch.org/docs/latest/im-plugin/ism/api/#get-policy
+        url: https://opensearch.org/docs/latest/im-plugin/ism/api/#put-policy
       parameters:
         - $ref: '#/components/parameters/ism.get_policy::path.policy_id'
       responses:
         '200':
           $ref: '#/components/responses/ism.get_policy@200'
+    head:
+      operationId: ism.exists_policy.0
+      x-operation-group: ism.exists_policy
```
nhtruong, line 46: maybe `policy_exists`?

spec/namespaces/ism.yaml
```diff
@@ -58,6 +110,18 @@ paths:
       responses:
         '200':
           $ref: '#/components/responses/ism.add_policy@200'
+  /_plugins/_ism/remove:
+    post:
+      operationId: ism.remove.0
+      x-operation-group: ism.remove
```
nhtruong, line 116: This should be in the same group as `ism.remove_policy`

spec/namespaces/ism.yaml
```diff
@@ -70,6 +134,20 @@ paths:
       responses:
         '200':
           $ref: '#/components/responses/ism.remove_policy@200'
+  /_plugins/_ism/change_policy:
+    post:
+      operationId: ism.change.0
+      x-operation-group: ism.change
```
nhtruong, line 140: This should be in the same group as `ism_change_policy`

spec/namespaces/ism.yaml
```diff
@@ -84,36 +162,160 @@ paths:
...
+        $ref: '#/components/requestBodies/ism.explain'
+      responses:
+        '200':
+          $ref: '#/components/responses/ism.explain@200'
   /_plugins/_ism/explain/{index}:
     get:
-      operationId: ism.explain_index.0
-      x-operation-group: ism.explain_index
+      operationId: ism.explain_policy.0
+      x-operation-group: ism.explain_policy
```
nhtruong, line 191: Renaming groups will result in breaking changes in the generated clients. Are you sure you wanna do this?

dblock: Yes, it's all new code. I added ISM support that has yet to be merged into any client AFAIK.

spec/namespaces/ism.yaml
dblock: But they have different parameters, one specifies the index in the query, the other in the path. So if you try to make them the same group you get this:

```
{
  file: 'namespaces/ism.yaml',
  location: 'Operation Group: ism.change_policy',
  message: "2 'ism.change_policy' operations must have an identical set of query parameters."
}
```

spec/namespaces/ism.yaml
dblock: Same as for `change_policy` they have different parameters, one specifies the index in the query, the other in the path.

spec/namespaces/ism.yaml
```diff
@@ -44,6 +82,20 @@ paths:
       responses:
         '200':
           $ref: '#/components/responses/ism.delete_policy@200'
+  /_plugins/_ism/add:
+    post:
+      operationId: ism.add.0
+      x-operation-group: ism.add
```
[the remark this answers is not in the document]
dblock: Same as for `change_policy` they have different parameters, one specifies the index in the query, the other in the path.

spec/namespaces/ism.yaml
dblock: We have `exists_component_template` and similar, but none of `_exists`.

spec/namespaces/ism.yaml
nhtruong: That's correct. But as far as the clients users are concerned, which is what the operation group is for, where the `index` param is put in the URL request is irrelevant. We have many API actions, like search, that also allows the index to be specified as either a path param or a query param, the generated clients will favor use it as a path param regardless.

dblock: Ok, I had to add `query` to the one that doesn't really take it, but it won't matter from the client POV I think.

spec/namespaces/ism.yaml
dblock: Updated.

Issue #537, opened 2024-08-28
[FEATURE] Document new experimental ingestion streaming APIs


reta:
### Is your feature request related to a problem?
The new experimental ingestion streaming APIs are feature complete and need API specs.

### What solution would you like?
Document new experimental ingestion streaming APIs. 

The APIs use NDJSON (Newline delimited JSON, https://github.com/ndjson/ndjson-spec) which fits perfectly to the streaming use cases: each chunk sent along the request represents distinct bulk operation/action (index, create, delete, and update). 

### What alternatives have you considered?
N/A

### Do you have any additional context?
- https://github.com/opensearch-project/OpenSearch/issues/9070

reta:
@dblock could you please assign it to me? it is scheduled to 2.17.0, thank you!

PR #584, opened 2024-09-18
Document new experimental ingestion streaming APIs

reta:
### Description
Document new experimental ingestion streaming APIs

### Issues Resolved
Closes https://github.com/opensearch-project/opensearch-api-specification/issues/537


spec/namespaces/_core.yaml
```diff
@@ -76,6 +76,57 @@ paths:
       responses:
         '200':
           $ref: '#/components/responses/bulk@200'
+  /_bulk/stream:
+    post:
+      operationId: bulk.stream.0
+      x-operation-group: bulk
```
nhtruong, line 82: The running linter will throw an error on the `operationId` and the `x-operation-group`.
I think you want to do `bulk_stream` for `x-operation-group` and `bulk_stream.0` for `operationId`. Then on the generated client, there will be a `client.bulk_stream()` method, which is different from `client.bulk()`

nhtruong:
Please add tests to these operations as well.

spec/namespaces/_core.yaml
reta: Thanks a lot for the hints, @nhtruong !

spec/namespaces/_core.yaml
```diff
@@ -76,6 +76,57 @@ paths:
       responses:
         '200':
           $ref: '#/components/responses/bulk@200'
+  /_bulk/stream:
+    post:
+      operationId: bulk_stream.0
+      x-operation-group: bulk_stream
+      x-version-added: '2.17'
+      description: Allows to perform multiple index/update/delete operations using request / response streaming
```
dblock, line 84:
```suggestion
      description: Allows to perform multiple index/update/delete operations using request/response streaming.
```

spec/namespaces/_core.yaml
```diff
@@ -76,6 +76,57 @@ paths:
...
+      requestBody:
+        $ref: '#/components/requestBodies/bulk'
+      responses:
+        '200':
+          $ref: '#/components/responses/bulk@200'
+    put:
+      operationId: bulk_stream.1
+      x-operation-group: bulk_stream
+      x-version-added: '2.17'
+      description: Allows to perform multiple index/update/delete operations using request / response streaming
```
dblock, line 109:
```suggestion
      description: Allows to perform multiple index/update/delete operations using request/response streaming.
```

spec/namespaces/_core.yaml
```diff
@@ -3217,6 +3268,23 @@ components:
...
       x-default: '1'
+    bulk::query.batch_size:
+      in: query
+      name: batch_size
+      description: |-
+        Specifies how many bulk operations should be accumulated into a batch before sending the batch to data nodes.
+      schema:
+        $ref: '../schemas/_common.yaml#/components/schemas/BatchSize'
+      style: form
+      x-default: '1'
```
dblock, line 3279:
```suggestion
      x-default: 1
```

dblock (changes requested):
Good start. Needs tests, please!

tests/default/_core/bulk_stream.yaml
```diff
@@ -0,0 +1,54 @@
+$schema: ../../../json_schemas/test_story.schema.yaml
```
dblock, line 1: File should be `_core/bulk/stream.yaml`.

tests/default/indices/bulk_stream.yaml
```diff
@@ -0,0 +1,60 @@
+$schema: ../../../json_schemas/test_story.schema.yaml
```
dblock, line 1: Same, `bulk/stream.yaml`.

dblock (changes requested):
You'll need to update the 2.17 version used to the released one, and switch the next version to 2.18.

reta:
> So does it actually stream data? 🤔

It does, I am not sure how we could express that in OpenAPI

dblock:
> > So does it actually stream data? 🤔
> 
> It does, I am not sure how we could express that in OpenAPI

I don't think we need to, but do clients automatically do it? AFAIK we don't have code in our tooling that does anything special.

reta:
```
ERROR   _core/bulk/stream.yaml (tests/default/_core/bulk/stream.yaml)
    ERROR   CHAPTERS
        ERROR   Create an index.
            PASSED  REQUEST BODY
            ERROR   RESPONSE STATUS (Expected status 200, but received 500: application/json. The engine does not support HTTP streaming, unable to serve uri [/_bulk/stream] and method [POST])
            SKIPPED RESPONSE PAYLOAD BODY
        SKIPPED Delete document in an index.
        SKIPPED Bulk document CRUD.
```

@dblock need your advice here: streaming (as of today) require dedicated transport plugin. Here are 2 options we could make it work:
 - ignore the test if particular plugin is not installed, or
 - install (and make default) the streaming transport

reta:
> I don't think we need to, but do clients automatically do it? AFAIK we don't have code in our tooling that does anything special.

Oh, we don't need to actually, no special treatment required

dblock:
> * install (and make default) the streaming transport

I think this is the right answer. The tooling in this repo should support streaming just like we support various content types.

tests/default/indices/bulk/stream.yaml
```diff
@@ -0,0 +1,60 @@
+$schema: ../../../../json_schemas/test_story.schema.yaml
```
dblock, line 1: You'll have to move tests in the plugins folder as well from tests/default.

dblock (approved):

Issue #583, opened 2024-09-17
Update k-NN API specifications with 2.17 changes


ryanbogan:
In OpenSearch 2.17, k-NN made changes to various APIs, which have not yet been reflected in the specifications.  These changes include adding rescore and oversampling factor to a k-NN query, and adding mode and compression options during index creation and training models.

The API specifications need to be updated to be consistent with the current code in the k-NN repository.

PR #588, opened 2024-09-24
Update API specifications for k-NN 2.17 changes

ryanbogan:
### Description
This PR updates the API specifications to be consistent with changes released for k-NN in OpenSearch 2.17.
These changes include adding rescore and oversampling factor to a k-NN query, and adding mode and compression options during index creation and training models.

### Issues Resolved
https://github.com/opensearch-project/opensearch-api-specification/issues/583


CHANGELOG.md
```diff
@@ -158,6 +158,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Fixed tasks namespace schemas ([#520](https://github.com/opensearch-project/opensearch-api-specification/pull/520))
 - Fixed `/_plugins/_transform/_preview` ([#568](https://github.com/opensearch-project/opensearch-api-specification/pull/568))
 - Fixed create/delete/index operation in `_bulk` ([#582](https://github.com/opensearch-project/opensearch-api-specification/pull/582))
+- Update API specifications for k-NN 2.17 changes ([#588](https://github.com/opensearch-project/opensearch-api-specification/pull/588))
```
dblock, line 161: Can we get a little more specific of what the changes are? New APIs? Fixed responses?

e.g. "Added `mode` and `compression_level` to ..."

tests/default/_core/search/knn_disk_based.yaml
```diff
@@ -0,0 +1,71 @@
...
+        -  { "index": { "_index": "my-vector-index", "_id": "5" } }
+        -  { "my_vector_field": [ 5.5, 5.5, 5.5, 5.5, 5.5, 5.5, 5.5, 5.5 ], "price": 3.7 }
+        -  { "index": { "_index": "my-vector-index", "_id": "6" } }
+        -  { "my_vector_field": [ 6.5, 6.5, 6.5, 6.5, 6.5, 6.5, 6.5, 6.5 ], "price": 10.3 }
+        -  { "index": { "_index": "my-vector-index", "_id": "7" } }
+        -  { "my_vector_field": [ 7.5, 7.5, 7.5, 7.5, 7.5, 7.5, 7.5, 7.5 ], "price": 5.5 }
+        -  { "index": { "_index": "my-vector-index", "_id": "8" } }
+        -  { "my_vector_field": [ 8.5, 8.5, 8.5, 8.5, 8.5, 8.5, 8.5, 8.5 ], "price": 4.4 }
+        -  { "index": { "_index": "my-vector-index", "_id": "9" } }
+        -  { "my_vector_field": [ 9.5, 9.5, 9.5, 9.5, 9.5, 9.5, 9.5, 9.5 ], "price": 8.9 }
```
dblock, line 46: We plan to use these examples in documentation, so maybe rename the fields to be movie/books based. For example instead of `my_vector_field` it could just be called `recommendation_vector`, and `price` could remain `price` or be `duration`.

tests/default/_core/search/knn_disk_based.yaml
```diff
@@ -0,0 +1,71 @@
...
+    status: [200]
+epilogues:
+  - path: /my-vector-index
```
dblock, line 49: Use `movies` or `books`.

tests/default/_core/search/knn_disk_based.yaml
```diff
@@ -0,0 +1,71 @@
+$schema: ../../../../json_schemas/test_story.schema.yaml
```
dblock, line 1: Let's make a folder for these and match the relevant option name, so `knn/on_disc.yaml`.

dblock (changes requested):
Looking good.

1. Adopt the movies/books theme in the examples since we plan to use those in the documentation.
2. Add 2.17 to CI/CD.

dblock:
> @dblock Do I need to change the sha ref in the workflow file to get 3.0 to pass?

Yes, because that's `main` and your additions exist in a newer build.

If you have a moment, also check whether the build you're getting fixes https://github.com/opensearch-project/opensearch-api-specification/blob/4f8be9eb9a62ea86a3a864f732c6fc126e4554a6/tests/default/asynchronous_search/stats.yaml#L9 or similar checks that were disabled because main wasn't including a plugin.

.github/workflows/test-spec.yml
```diff
@@ -37,8 +37,6 @@ jobs:
           - version: 2.16.0
             tests: snapshot
           - version: 2.17.0
-            hub: opensearchstaging
-            ref: '@sha256:1273489ebbedcb470ea13563dae4c6dc6b2ed431e87e686ed931ae0733034b25'
```
dblock, line 41: Add the latest 2.18 from `opensearchstaging` (2.x) please.

ryanbogan:
Sure, I'll take a look

CHANGELOG.md
```diff
@@ -158,6 +158,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Fixed tasks namespace schemas ([#520](https://github.com/opensearch-project/opensearch-api-specification/pull/520))
 - Fixed `/_plugins/_transform/_preview` ([#568](https://github.com/opensearch-project/opensearch-api-specification/pull/568))
 - Fixed create/delete/index operation in `_bulk` ([#582](https://github.com/opensearch-project/opensearch-api-specification/pull/582))
+- Add mode and compression to k-NN index creation and search, and add rescore and oversample_factor to k-NN search ([#588](https://github.com/opensearch-project/opensearch-api-specification/pull/588))
```
dblock, line 161:
```suggestion
- Add `mode` and `compression` to k-NN index creation and search, and add `rescore` and `oversample_factor` to k-NN search ([#588](https://github.com/opensearch-project/opensearch-api-specification/pull/588))
```

spec/schemas/_common.yaml
```diff
@@ -656,6 +656,16 @@ components:
...
+        method_parameters:
+          type: object
+          x-version-added: '2.16'
+          additionalProperties:
+            type: number
+        rescore:
+          type: object
+          x-version-added: '2.17'
+          additionalProperties:
+            type: number
```
navneet1v, line 668: if I understand this correctly, the additionalProperties are more of a map whose values will always be of type number? correct?

if the above understanding is correct, then my worry is going forward we can add more parameters and those parameters might not be of number type. Can use some type like object or anything that can allow us to specify number, strings as additional parameter values in future.

@dblock thoughts?

ryanbogan: If we add them in the future, we can update this with:
```
anyOf:
  type:
    number
    string
    etc.
```

ryanbogan:
Looks like the asynchronous search tests are still failing on 3.0, will remove the version checks again

spec/schemas/_common.yaml
dblock: @navneet1v you're correct, and @ryanbogan is also correct, if the schema _today_ is a number, leave it as such, and tomorrow it can be extended when it can actually be something else.

dblock (approved):

spec/schemas/_common.yaml
dblock: Yes.

PR #612, opened 2024-10-13
update with accurate numeric type for common, core

### LINKED ISSUE #596: [BUG] inaccurate numeric type
### What is the bug?
there are many places using generic "number" as numeric type. 

example: https://github.com/opensearch-project/opensearch-api-specification/blob/939a292753eeb1b3bffdf5ed4f3039ab5edd2f04/spec/schemas/_core.bulk.yaml#L56

should use more accurate definition for numeric types. 
https://swagger.io/docs/specification/v3_0/data-models/data-types/

amberzsy:
### Description
improve accuracy of numeric type for document, search api. 

### Issues Resolved
#596 

### DIFF
--- a/spec/schemas/_common.query_dsl.yaml
+++ b/spec/schemas/_common.query_dsl.yaml
@@ -1824,7 +1824,8 @@ components:
       description: Can only be used as a clause in a span_near query.
       type: object
       additionalProperties:
-        type: number
+        type: integer
+        format: int32
       minProperties: 1
       maxProperties: 1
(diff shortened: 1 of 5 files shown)


dblock (approved):

amberzsy:
> I wonder whether it's a good idea to add a linter rule that warns us that a type is too generic? And then eventually turn it into an error? If you agree do open an issue please!

agree.  opened issue [here](https://github.com/opensearch-project/opensearch-api-specification/issues/617)

PR #613, opened 2024-10-15
Adding specs for new _list/indices and _list/shards APIs

gargharsh3134:
### Description
Adding specs for ```_list/indices``` and ```_list/shards``` APIs, new paginated alternates of ```_cat/indices``` and ```_cat/shards```. Both the new list APIs will continue to support all the params supported by existing cat API counterparts in addition to three new parameters which are specific to pagination (namely ```next_token```, ```size``` and ```order```).

Related Opensearch changes: https://github.com/opensearch-project/OpenSearch/issues/14258 & https://github.com/opensearch-project/OpenSearch/issues/14257

### Issues Resolved

### DIFF
--- a/spec/namespaces/list.yaml
+++ b/spec/namespaces/list.yaml
@@ -1,0 +1,480 @@
+openapi: 3.1.0
+info:
+  title: OpenSearch List API
+  description: OpenSearch List API.
+  version: 1.0.0
+paths:
+  /_list:
+    get:
+      operationId: list.help.0
+      x-operation-group: list.help
+      x-version-added: '2.18'
+      description: Returns help for the List APIs.
+      externalDocs:
+        url: https://opensearch.org/docs/latest/api-reference/list/index/
+      responses:
+        '200':
+          $ref: '#/components/responses/list.help@200'
+  /_list/indices:
+    get:
+      operationId: list.indices.0
+      x-operation-group: list.indices
+      x-version-added: '2.18'
+      description: 'Returns information about indices in a paginated fashion: number of primaries and replicas, document counts, disk size, ...'
+      externalDocs:
+        url: https://opensearch.org/docs/latest/api-reference/list/list-indices/
+      parameters:
+        - $ref: '#/components/parameters/list.indices::query.bytes'
+        - $ref: '#/components/parameters/list.indices::query.cluster_manager_timeout'
+        - $ref: '#/components/parameters/list.indices::query.expand_wildcards'
+        - $ref: '#/components/parameters/list.indices::query.format'
+        - $ref: '#/components/parameters/list.indices::query.h'
+        - $ref: '#/components/parameters/list.indices::query.health'
+        - $ref: '#/components/parameters/list.indices::query.help'
+        - $ref: '#/components/parameters/list.indices::query.include_unloaded_segments'
+        - $ref: '#/components/parameters/list.indices::query.local'
+        - $ref: '#/components/parameters/list.indices::query.master_timeout'
+        - $ref: '#/components/parameters/list.indices::query.pri'
+        - $ref: '#/components/parameters/list.indices::query.s'
+        - $ref: '#/components/parameters/list.indices::query.time'
+        - $ref: '#/components/parameters/list.indices::query.v'
+        - $ref: '#/components/parameters/list.indices::query.next_token'
+        - $ref: '#/components/parameters/list.indices::query.size'
+        - $ref: '#/components/parameters/list.indices::query.sort'
+      responses:
+        '200':
+          $ref: '#/components/responses/list.indices@200'
+  /_list/indices/{index}:
+    get:
+      operationId: list.indices.1
+      x-operation-group: list.indices
+      x-version-added: '2.18'
+      description: 'Returns information about indices in a paginated fashion: number of primaries and replicas, document counts, disk size, ...'
+      externalDocs:
+        url: https://opensearch.org/docs/latest/api-reference/list/list-indices/
+      parameters:
+        - $ref: '#/components/parameters/list.indices::path.index'
+        - $ref: '#/components/parameters/list.indices::query.bytes'
+        - $ref: '#/components/parameters/list.indices::query.cluster_manager_timeout'
+        - $ref: '#/components/parameters/list.indices::query.expand_wildcards'
+        - $ref: '#/components/parameters/list.indices::query.format'
+        - $ref: '#/components/parameters/list.indices::query.h'
+        - $ref: '#/components/parameters/list.indices::query.health'
+        - $ref: '#/components/parameters/list.indices::query.help'
+        - $ref: '#/components/parameters/list.indices::query.include_unloaded_segments'
+        - $ref: '#/components/parameters/list.indices::query.local'
+        - $ref: '#/components/parameters/list.indices::query.master_timeout'
+        - $ref: '#/components/parameters/list.indices::query.pri'
+        - $ref: '#/components/parameters/list.indices::query.s'
+        - $ref: '#/components/parameters/list.indices::query.time'
+        - $ref: '#/components/parameters/list.indices::query.v'
+        - $ref: '#/components/parameters/list.indices::query.next_token'
+        - $ref: '#/components/parameters/list.indices::query.size'
+        - $ref: '#/components/parameters/list.indices::query.sort'
+      responses:
+        '200':
+          $ref: '#/components/responses/list.indices@200'
+  /_list/shards:
+    get:
+      operationId: list.shards.0
+      x-operation-group: list.shards
+      x-version-added: '2.18'
+      description: Provides a detailed view of shard allocation on nodes in a paginated fashion.
+      externalDocs:
+        url: https://opensearch.org/docs/latest/api-reference/list/list-shards/
+      parameters:
+        - $ref: '#/components/parameters/list.shards::query.bytes'
+        - $ref: '#/components/parameters/list.shards::query.cluster_manager_timeout'
+        - $ref: '#/components/parameters/list.shards::query.format'
+        - $ref: '#/components/parameters/list.shards::query.h'
+        - $ref: '#/components/parameters/list.shards::query.help'
+        - $ref: '#/components/parameters/list.shards::query.local'
+        - $ref: '#/components/parameters/list.shards::query.master_timeout'
+        - $ref: '#/components/parameters/list.shards::query.s'
+        - $ref: '#/components/parameters/list.shards::query.time'
+        - $ref: '#/components/parameters/list.shards::query.v'
+        - $ref: '#/components/parameters/list.shards::query.next_token'
+        - $ref: '#/components/parameters/list.shards::query.size'
+        - $ref: '#/components/parameters/list.shards::query.sort'
+      responses:
+        '200':
+          $ref: '#/components/responses/list.shards@200'
+  /_list/shards/{index}:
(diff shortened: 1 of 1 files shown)


gargharsh3134:
@dblock Can you please take a look again? Added tests taking inspiration from existing _cat APIs. Verified the tests by running them against Amazon Opensearch domain.

gargharsh3134:
The staging 2.18 distribution seems to be missing the _list APIs, and as a result Spec test in CI are failing.

spec/namespaces/list.yaml
```diff
@@ -0,0 +1,485 @@
...
+        url: https://opensearch.org/docs/latest/api-reference/list/index/
+      responses:
+        '200':
+          $ref: '#/components/responses/list.help@200'
+  /_list/indices:
+    get:
+      operationId: list.indices.0
+      x-operation-group: list.indices
+      x-version-added: '2.18'
+      description: 'Returns information about indices in a paginated fashion: number of primaries and replicas, document counts, disk size, ...'
```
dblock, line 23: Let's make it a full sentence.

```suggestion
      description: 'Returns paginated information about indices including number of primaries and replicas, document counts, disk size, etc.'
```

spec/namespaces/list.yaml
```diff
@@ -0,0 +1,485 @@
...
+        - $ref: '#/components/parameters/list.indices::query.v'
+      responses:
+        '200':
+          $ref: '#/components/responses/list.indices@200'
+  /_list/indices/{index}:
+    get:
+      operationId: list.indices.1
+      x-operation-group: list.indices
+      x-version-added: '2.18'
+      description: 'Returns information about indices in a paginated fashion: number of primaries and replicas, document counts, disk size, ...'
```
dblock, line 52: Same as above.

spec/namespaces/list.yaml
```diff
@@ -0,0 +1,485 @@
...
+        - $ref: '#/components/parameters/list.indices::query.v'
+      responses:
+        '200':
+          $ref: '#/components/responses/list.indices@200'
+  /_list/shards:
+    get:
+      operationId: list.shards.0
+      x-operation-group: list.shards
+      x-version-added: '2.18'
+      description: Provides a detailed view of shard allocation on nodes in a paginated fashion.
```
dblock, line 82:
```suggestion
      description: Returns paginated details of shard allocation on nodes.
```

spec/namespaces/list.yaml
```diff
@@ -0,0 +1,485 @@
...
+        - $ref: '#/components/parameters/list.shards::query.v'
+      responses:
+        '200':
+          $ref: '#/components/responses/list.shards@200'
+  /_list/shards/{index}:
+    get:
+      operationId: list.shards.1
+      x-operation-group: list.shards
+      x-version-added: '2.18'
+      description: Provides a detailed view of shard allocation on nodes in a paginated fashion.
```
dblock, line 107: Same as above.

spec/namespaces/list.yaml
```diff
@@ -0,0 +1,485 @@
...
+      name: size
+      in: query
+      description: Maximum number of indices to be displayed in a page.
+      schema:
+        type: integer
+        format: int32 
+    list.indices::query.sort:
+      name: sort
+      in: query
+      description: Defines order in which indices will be displayed. If "desc", most recently created indices would be displayed first.
```
dblock, line 364: I think we quote these as `desc`, but double check.

spec/namespaces/list.yaml
```diff
@@ -0,0 +1,485 @@
...
+      name: size
+      in: query
+      description: Maximum number of shards to be displayed in a page.
+      schema:
+        type: integer
+        format: int32 
+    list.shards::query.sort:
+      name: sort
+      in: query
+      description: Defines order in which shards will be displayed. If "desc", most recently created shards would be displayed first.
```
dblock, line 480: Same as above.

tests/default/list/indices.yaml
```diff
@@ -0,0 +1,151 @@
...
+        - amazon-serverless
+    method: GET
+    path: /_list/indices
+    parameters:
+      format: cbor
+    response:
+      status: 200
+      content_type: application/cbor
+  - synopsis: List information related to indexes in different formats (format=smile).
+    version: '>= 2.18'
```
dblock, line 140: Since this whole thing is `>= 2.18`, move that to the top of this file and remove from individual tests.

tests/default/list/shards.yaml
```diff
@@ -0,0 +1,51 @@
...
+  - synopsis: List the state of all primary and replica shards and how they are distributed.
+    version: '>= 2.18'
+    path: /_list/shards
+    method: GET
+    parameters:
+      format: json
+      h:
+        - '*'
+  - synopsis: List the state of all primary and replica shards and how they are distributed for an index.
+    version: '>= 2.18'
```
dblock, line 38: Move up.

dblock (changes requested):
This is looking good!

The next version (2.18 and 3.0) distribution uses a specific SHA from https://hub.docker.com/r/opensearchstaging/opensearch/tags to remain stable. Update both to the latest ones that have the `_list` API, and make sure tests pass.

spec/namespaces/list.yaml
gargharsh3134: Updated in the revision.

spec/namespaces/list.yaml
gargharsh3134: Updated in the revision.

spec/namespaces/list.yaml
gargharsh3134: Updated in the revision.

spec/namespaces/list.yaml
gargharsh3134: Addressed in the revision.

spec/namespaces/list.yaml
gargharsh3134: Thanks for pointing out. Replaced "desc" with `desc` as is being used at other places.

spec/namespaces/list.yaml
gargharsh3134: Addressed in the revision.

tests/default/list/indices.yaml
gargharsh3134: Addressed in the revision.

tests/default/list/shards.yaml
gargharsh3134: Addressed in the revision.

gargharsh3134:
@dblock Thanks for reviewing. I have addressed previous set of comments, please do take a look again whenever you get a chance. Also, sorry for the delay in between the revisions, got occupied with other urgent items.

On the tests, i have updated the staging distributions for 2.18 and 3.0 and the newly added list tests are passing. Not sure why some of the existing ones have started to fail. Trying to root cause.

spec/namespaces/list.yaml
```diff
@@ -0,0 +1,485 @@
...
+        url: https://opensearch.org/docs/latest/api-reference/list/index/
+      responses:
+        '200':
+          $ref: '#/components/responses/list.help@200'
+  /_list/indices:
+    get:
+      operationId: list.indices.0
+      x-operation-group: list.indices
+      x-version-added: '2.18'
+      description: 'Returns paginated information about indices including number of primaries and replicas, document counts, disk size, etc.'
```
dblock, line 23:
```suggestion
      description: Returns paginated information about indices including number of primaries and replicas, document counts, disk size, etc.
```

spec/namespaces/list.yaml
```diff
@@ -0,0 +1,485 @@
...
+        - $ref: '#/components/parameters/list.indices::query.v'
+      responses:
+        '200':
+          $ref: '#/components/responses/list.indices@200'
+  /_list/indices/{index}:
+    get:
+      operationId: list.indices.1
+      x-operation-group: list.indices
+      x-version-added: '2.18'
+      description: 'Returns paginated information about indices including number of primaries and replicas, document counts, disk size, etc.'
```
dblock, line 52:
```suggestion
      description: Returns paginated information about indices including number of primaries and replicas, document counts, disk size, etc.
```

dblock (changes requested):
The 3.0 build gets generated even if not all plugins build with it, so looks like the update SHA doesn't have all the plugins? Try another more recent SHA until the tests all pass, possibly hunt down why the build is failing for the plugins that are failing (geospatial looks like).

spec/namespaces/list.yaml
gargharsh3134: Updated in the revision.

spec/namespaces/list.yaml
gargharsh3134: Updated in the revision.

gargharsh3134:
> The 3.0 build gets generated even if not all plugins build with it, so looks like the update SHA doesn't have all the plugins? Try another more recent SHA until the tests all pass, possibly hunt down why the build is failing for the plugins that are failing (geospatial looks like).

Got it, I had sorted the distributions via the newest and used the latest one already. Can this staging build be triggered manually to generate a latest distribution?

dblock:
> > The 3.0 build gets generated even if not all plugins build with it, so looks like the update SHA doesn't have all the plugins? Try another more recent SHA until the tests all pass, possibly hunt down why the build is failing for the plugins that are failing (geospatial looks like).
> 
> Got it, I had sorted the distributions via the newest and used the latest one already. Can this staging build be triggered manually to generate a latest distribution?

You should try with a newer SHA. The builds happen every day.

If the build is broken (other tests failing, please try another SHA or chase that down).

dblock:
Looks like the style checker is failing for good reason (use of indices instead of indexes, etc.).

nhtruong:
> Looks like the style checker is failing for good reason (use of indices instead of indexes, etc.).

We have a namespace called `indices` that has been around since the beginning of time. Shouldn't it be the other way around where the use indices over indexes?

dblock:
> > Looks like the style checker is failing for good reason (use of indices instead of indexes, etc.).
> 
> We have a namespace called `indices` that has been around since the beginning of time. Shouldn't it be the other way around where the use indices over indexes?

The descriptive text was standardized as "indexes", but technically it's "indices" in APIs and such I believe. Let's open an issue in docs about potentially changing that, but for now the style checker is correct.

gargharsh3134:
Test run against 3.0.0 build is the only failing task. Newly added spec tests though are passing, the existing plugin related tests are failing.

gargharsh3134:
@dblock @nhtruong Can you please help with suggestions on 3.0.0 build? I have picked up the latest SHA as part of this PR.
Thanks!

dblock:
Looks like we don't have a good 3.0 build since with all the plugins, hence the failures :(

Undo the 3.0 hash change and add `version: '< 3.0'` in the new list tests, maybe with a `# TODO: re-enable in 3.0, see  https://github.com/opensearch-project/opensearch-api-specification/pull/613`.

gargharsh3134:
> Undo the 3.0 hash change and add version: '< 3.0' in the new list tests, maybe with a # TODO: re-enable in 3.0, see  https://github.com/opensearch-project/opensearch-api-specification/pull/613.

@dblock Thanks for the suggestions. The tasks are green now, please take a look whenever you get a chance.
Thanks!

dblock (approved):

dblock:
Good work @gargharsh3134, thanks for hanging in here!

PR #615, opened 2024-10-16
Add spec for new rename_alias_pattern and rename_alias_replacement parameters to snapshot restore

mispencer:
### Description
Add spec for new rename_alias_pattern and rename_alias_replacement parameters to snapshot restore implemented in [Opensearch PR 16292](https://github.com/opensearch-project/OpenSearch/pull/16292)

### Issues Resolved
No related issues 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Added `PhoneAnalyzer` from `analysis-phonenumber` plugin ([#609](https://github.com/opensearch-project/opensearch-api-specification/pull/609))
 - Added `/_list/indices` & `/_list/shards` api specs ([#613](https://github.com/opensearch-project/opensearch-api-specification/pull/613))
 - Added `GET` and `HEAD /{index}/_source/{id}` returning `404` ([#673](https://github.com/opensearch-project/opensearch-api-specification/pull/673))
+- Added `rename_alias_pattern` and `rename_alias_replacment` to `/_snapshot/{repository}/{snapshot}/_restore` body parameters ([#615](https://github.com/opensearch-project/opensearch-api-specification/pull/615))
 
 ### Removed
 - Removed unsupported `_common.mapping:SourceField`'s `mode` field and associated `_common.mapping:SourceFieldMode` enum ([#652](https://github.com/opensearch-project/opensearch-api-specification/pull/652))
@@ -148,7 +149,6 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Added API spec for `adjust_pure_negative` for bool queries ([#641](https://github.com/opensearch-project/opensearch-api-specification/pull/641))
 - Added a spec style checker [#620](https://github.com/opensearch-project/opensearch-api-specification/pull/620).
 - Added `remote_store` to node `Stats` ([#643](https://github.com/opensearch-project/opensearch-api-specification/pull/643))
-- Added `rename_alias_pattern` and `rename_alias_replacment` to `/_snapshot/{repository}/{snapshot}/_restore` body parameters ([#615](https://github.com/opensearch-project/opensearch-api-specification/pull/615))
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -148,6 +148,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Added API spec for `adjust_pure_negative` for bool queries ([#641](https://github.com/opensearch-project/opensearch-api-specification/pull/641))
 - Added a spec style checker [#620](https://github.com/opensearch-project/opensearch-api-specification/pull/620).
 - Added `remote_store` to node `Stats` ([#643](https://github.com/opensearch-project/opensearch-api-specification/pull/643))
+- Added `rename_alias_pattern` and `rename_alias_replacment` to `/_snapshot/{repository}/{snapshot}/_restore` body parameters ([#615](https://github.com/opensearch-project/opensearch-api-specification/pull/615))
 
 ### Changed
 
--- a/spec/namespaces/snapshot.yaml
+++ b/spec/namespaces/snapshot.yaml
@@ -360,8 +360,10 @@ components:
               rename_replacement:
                 type: string
               rename_alias_pattern:
+                x-version-added: '2.18'
                 type: string
               rename_alias_replacement:
+                x-version-added: '2.18'
                 type: string
             description: Details of what to restore
   responses:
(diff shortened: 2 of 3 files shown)


mispencer:
The implementing PR for this spec change is not yet merged.

tests/snapshot/snapshot/restore.yaml
```diff
@@ -19,6 +19,12 @@ epilogues:
   - path: /books
     method: DELETE
     status: [200, 404]
+  - path: /stories
+    method: DELETE
+    status: [200, 404]
+  - path: /new_stories
```
dblock, line 25: We should call it something meaningful, maybe `tales`?

dblock (changes requested):
Looking good, needs a CHANGELOG, passing tests, etc.

tests/snapshot/snapshot/restore.yaml
mispencer: Do you mean `new_stories`? This is the restored-with-new-name index. I changed it to `renamed_stories`.

dblock: I just don't love names like `new_xyz` or `xyz_1` because we plan to use these samples to generate documentation. It's better with names like "movies" or "films", so instead of "new_stories" maybe we can use something semantically interesting like "tales"?

mispencer: This name is for the rename parameters for this API. An entirely new name like that would not fit the most common use case of those parameters, which is to restore a copy of existing index with a new, non-conflicting name.  So, even for documentation, I think a prefixed name like currently is more suitable. The [current documentation of restore](https://opensearch.org/docs/latest/api-reference/snapshots/restore-snapshot/) supports this - it has an example of a prefixed name, `recovered-logs`, and a suffixed name, `opendistro-reports-definitions_restored`, but no example of a entirely new name for renamed indexes.

dblock: Understood, thanks.

mispencer:
This feature has been merged.

tests/snapshot/snapshot/restore.yaml
mispencer: I have renamed this index and rewrote the rename parameters to match realistic use cases.

spec/namespaces/snapshot.yaml
```diff
@@ -359,6 +359,12 @@ components:
                 type: string
               rename_replacement:
                 type: string
+              rename_alias_pattern:
```
dblock, line 362: Take the time to add a `description:` field to these. Eventually they will appear in the documentation.

dblock:
Now that 2.18 has been released this should be easier. Want to finish this @mispencer?

spec/namespaces/snapshot.yaml
mispencer: I synced this with the current documentation description and parameters.

mispencer:
> Now that 2.18 has been released this should be easier. Want to finish this @mispencer?

Looks like this has already been updated in the main branch.

spec/namespaces/snapshot.yaml
```diff
@@ -335,30 +335,51 @@ components:
...
+                type: string
+                description: The rename replacement string for aliases.
+              source_remote_store_repository:
+                x-version-added: '2.10'
+                type: string
+                description: The name of the remote store repository of the source index being restored. If not provided, the Snapshot Restore API will use the repository that was registered when the snapshot was created.
+              storage_type:
+                x-version-added: '2.7'
+                type: string
+                description: '`local` indicates that all snapshot metadata and index data will be downloaded to local storage.  `remote_snapshot` indicates that snapshot metadata will be downloaded to the cluster, but the remote repository will remain the authoritative store of the index data. Data will be downloaded and cached as necessary to service queries. At least one node in the cluster must be configured with the search role in order to restore a snapshot using the type `remote_snapshot`.  Defaults to `local`.'
```
dblock, line 382:
```suggestion
                description: |-
                     Storage type.
                     A value of `local` indicates that all snapshot metadata and index data will be downloaded to local storage.
                     A value of `remote_snapshot` indicates that snapshot metadata will be downloaded to the cluster, but the remote repository will remain the authoritative store of the index data.
                     Data will be downloaded and cached as necessary to service queries. At least one node in the cluster must be configured with the search role in order to restore a snapshot using the type `remote_snapshot`.
                     Defaults to `local`.
```

mispencer: I added line breaks to match the [source documentation](https://github.com/opensearch-project/documentation-website/blob/main/_api-reference/snapshots/restore-snapshot.md).

tests/snapshot/snapshot/restore.yaml
```diff
@@ -98,4 +118,40 @@ chapters:
             - stage: DONE
               type: SNAPSHOT
     retry:
-      count: 3
\ No newline at end of file
+      count: 3
+  - synopsis: Restore snapshot with rename_pattern and rename_replacement.
```
dblock, line 122: For a future PR (unless this one needs more iterating) you should break up this test into `snapshot/restore/index.yaml` for basic scenarios `restore/rename.yaml`, etc., this way each stands alone as an example that can be reused in the appropriate documentation.

CHANGELOG.md
```diff
@@ -139,6 +139,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Added API spec for `adjust_pure_negative` for bool queries ([#641](https://github.com/opensearch-project/opensearch-api-specification/pull/641))
 - Added a spec style checker [#620](https://github.com/opensearch-project/opensearch-api-specification/pull/620).
 - Added `remote_store` to node `Stats` ([#643](https://github.com/opensearch-project/opensearch-api-specification/pull/643))
+- Added `rename_alias_pattern` and `rename_alias_replacment` to `/_snapshot/{repository}/{snapshot}/_restore` body parameters ([#615](https://github.com/opensearch-project/opensearch-api-specification/pull/615))
```
dblock, line 142: This should go into the unreleased section above, and maybe the one above was also put in the wrong section?

dblock (changes requested):
Sorry to be a pest about CHANGELOG.

Style checker failed too. Doc uses indexes :(

CHANGELOG.md
mispencer: This has been moved.

dblock (approved):

dblock:
Merged.

Would you have a bit of time to break up the long test into snapshot/restore/index.yaml for basic scenarios restore/rename.yaml, etc.? This way each stands alone as an example that can be reused in the appropriate documentation.

spec/namespaces/snapshot.yaml
```diff
@@ -294,7 +294,7 @@ components:
             type: object
             properties:
               ignore_unavailable:
-                description: If `true`, the request ignores data streams and indexes in `indices` that are missing or closed. If `false`, the request returns an error for any data stream or index that is missing or closed.
+                description: If `true`, the request ignores data streams and indexes in `indexes` that are missing or closed. If `false`, the request returns an error for any data stream or index that is missing or closed.
```
dblock, line 297: If this refers to a variable (indices) then quoting it will work, so I think this means `indices`.

PR #616, opened 2024-10-16
Add basic GET tests for core and indices endpoints

brunoLloret:
### Description
Set of basic GET tests for various core and indices endpoints:

Added tests for core endpoints: _refresh, _remote/info, _script_context, and _script_language
Added tests for indices endpoints: _shard_stores and _template
Updated docker-compose.yml to include OpenSearch and OpenSearch Dashboards services
Minor updates to package.json and .gitignore

### Issues Resolved


.gitignore
```diff
@@ -19,3 +19,4 @@ build/
 # coverage output
 /coverage/
 
+.env
```
dblock, line 22: Do you use a `.env` file with a tool? Maybe something we can recommend in dev guide for env variables?

dblock (changes requested):
This is great, thanks for filling some of these out! 

- Get rid of the extra docker compose, is Dashboards needed for these tests? We should probably think about introducing that separately, and otherwise it should go into `tests/default`.
- Fix DCO, `git commit --amend -s`.
- Rebase/squash your commits, will be easier to fix 1 commit for DCO.
- Fix linter, `npm run lint--fix`, possibly more.
- Mark this as ready, let's merge it.

dblock:
@brunoLloret lmk if you need help, I'm dblock on the OpenSearch public slack (https://opensearch.org/slack.html)

brunoLloret:
> @brunoLloret lmk if you need help, I'm dblock on the OpenSearch public slack (https://opensearch.org/slack.html)

Hey! I tried to join the OpenSearch public Slack using the link you provided (https://opensearch.org/slack.html), but it says the link is no longer active. Do you know if there's an updated link or another way to join the Slack workspace?

dblock:
> > @brunoLloret lmk if you need help, I'm dblock on the OpenSearch public slack (https://opensearch.org/slack.html)
> 
> Hey! I tried to join the OpenSearch public Slack using the link you provided (https://opensearch.org/slack.html), but it says the link is no longer active. Do you know if there's an updated link or another way to join the Slack workspace?

Uh oh, @krisfreedain?

.gitignore
brunoLloret: I used it for setting my local environment and perhaps it does not make sense to add it for the remote repo's codebase. I think it can be removed for the remote repo although if you think it's good to document for the guidelines I can do it. Lmk!

Issue #624, opened 2024-10-17
[FEATURE] Query Insights API spec


ansjcy:
### Is your feature request related to a problem?
We have introduced query insights plugin with top n queries API but we don't have API spec defined for the new API.

### What solution would you like?
Add API spec for top queries https://opensearch.org/docs/latest/observing-your-data/query-insights/top-n-queries/

### What alternatives have you considered?
N/A

### Do you have any additional context?
Please refer to the below doc for more details
- https://opensearch.org/docs/latest/observing-your-data/query-insights/top-n-queries/
- https://github.com/opensearch-project/OpenSearch/issues/11522

dblock:
Closed via https://github.com/opensearch-project/opensearch-api-specification/pull/625.

PR #625, opened 2024-10-17
Query Insights API spec

ansjcy:
### Description
API spec for query insights, more specifically, for the top n queries API.
Related API: https://opensearch.org/docs/latest/observing-your-data/query-insights/top-n-queries/

### Issues Resolved
https://github.com/opensearch-project/opensearch-api-specification/issues/624

### DIFF
--- a/spec/namespaces/query_insights.yaml
+++ b/spec/namespaces/query_insights.yaml
@@ -1,0 +1,29 @@
+openapi: 3.1.0
+info:
+  title: Query Insights API
+  description: API to retrieve top queries based on latency, CPU, or memory usage.
+  version: 1.0.0
+paths:
+  /_insights/top_queries:
+    get:
+      operationId: insights.getTopQueries
+      x-operation-group: insights
+      description: Retrieves the top queries based on the given metric type (latency, CPU, or memory).
+      parameters:
+        - $ref: '#/components/parameters/type'
+      responses:
+        '200':
+          content:
+            application/json:
+              schema:
+                $ref: '../schemas/query_insights._common.yaml#/components/schemas/TopQueriesResponse'
+components:
+  parameters:
+    type:
+      name: type
+      in: query
+      required: true
+      description: Get top n queries by a specific metric.
+      schema:
+        type: string
+        enum: [latency, cpu, memory]
--- a/tests/plugins/query_insights/docker-compose.yml
+++ b/tests/plugins/query_insights/docker-compose.yml
@@ -1,0 +1,15 @@
+version: '3'
+
+services:
+  opensearch-cluster:
+    image: ${OPENSEARCH_DOCKER_HUB_PROJECT:-opensearchproject}/opensearch:${OPENSEARCH_VERSION:-latest}${OPENSEARCH_DOCKER_REF}
+    ports:
+      - 9200:9200
+      - 9600:9600
+    environment:
+      - OPENSEARCH_INITIAL_ADMIN_PASSWORD=${OPENSEARCH_PASSWORD:-myStrongPassword123!}
+      - OPENSEARCH_JAVA_OPTS=${OPENSEARCH_JAVA_OPTS}
+      - discovery.type=single-node
+      - search.insights.top_queries.latency.enabled=true
+      - search.insights.top_queries.cpu.enabled=true
+      - search.insights.top_queries.memory.enabled=true
--- a/tests/plugins/query_insights/top_queries/top_queries.yaml
+++ b/tests/plugins/query_insights/top_queries/top_queries.yaml
@@ -1,0 +1,88 @@
+$schema: ../../../../json_schemas/test_story.schema.yaml
+
+description: Test top n queries API in the Query Insights plugin.
+version: '>= 2.15'
+prologues:
+  - path: /movies
+    method: PUT
+    request:
+      payload:
+        mappings:
+          properties:
+            director:
+              type: text
+              fielddata: true
+              fields:
+                raw:
+                  type: keyword
+  - path: /_bulk
+    method: POST
+    parameters:
+      refresh: true
+    request:
+      content_type: application/x-ndjson
+      payload:
+        - {create: {_index: movies}}
+        - {director: Bennett Miller, title: Moneyball}
+        - {create: {_index: movies}}
+        - {director: Bennett Miller, title: The Cruise}
+        - {create: {_index: movies}}
+        - {director: Nicolas Winding Refn, title: Drive}
+
+  - path: /{index}/_search
+    parameters:
+      index: movies
+    method: GET
+    request:
+      payload:
+        size: 0
+        aggregations:
+          directors:
+            terms:
+              field: director.raw
+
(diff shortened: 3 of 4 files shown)


spec/namespaces/query_insights.yaml
```diff
@@ -0,0 +1,29 @@
+openapi: 3.1.0
```
dblock, line 1: Should we call this `insights.yaml` to match the API?

tests/plugins/query_insights/docker-compose.yml
```diff
@@ -0,0 +1,15 @@
...
+    ports:
+      - 9200:9200
+      - 9600:9600
+    environment:
+      - OPENSEARCH_INITIAL_ADMIN_PASSWORD=${OPENSEARCH_PASSWORD:-myStrongPassword123!}
+      - OPENSEARCH_JAVA_OPTS=${OPENSEARCH_JAVA_OPTS}
+      - discovery.type=single-node
+      - search.insights.top_queries.latency.enabled=true
+      - search.insights.top_queries.cpu.enabled=true
+      - search.insights.top_queries.memory.enabled=true
```
dblock, line 15: Are these custom parameters required, meaning is query insights not working in the default configuration? If that's the case then this is right, otherwise add tests to `default`.

tests/plugins/query_insights/top_queries/top_queries.yaml
```diff
@@ -0,0 +1,88 @@
+$schema: ../../../../json_schemas/test_story.schema.yaml
```
dblock, line 1: Since this is for the `_insights` API, make the path `insights`.

tests/plugins/query_insights/top_queries/top_queries.yaml
tests/plugins/query_insights/**insights**/top_queries.yaml

dblock (changes requested):
Looks good! Some asks below, iterate till tests/lints pass, add to CHANGELOG.

tests/plugins/query_insights/docker-compose.yml
ansjcy: added tests for default

dblock: I meant removing the `tests/plugins/query_insights` folder and moving the tests to the `tests/default` folder.

ansjcy:
Hi @dblock! Thanks a lot for your review. I just pushed a commit based on your comments, let me know if it looks good to you!

tests/plugins/query_insights/docker-compose.yml
ansjcy: query insights features are all in a plugin that needs to be installed first. In this case shouldn't we write the tests in the plugin folder?

dblock (changes requested):
Good. You do have to add `query_insights` to the CI/CD matrix.

https://github.com/opensearch-project/opensearch-api-specification/blob/main/.github/workflows/test-spec.yml#L25

tests/plugins/query_insights/docker-compose.yml
dblock: Cool. Add this to the CI/CD matrix in https://github.com/opensearch-project/opensearch-api-specification/blob/main/.github/workflows/test-spec.yml#L25 and it should be it.

dblock: And since everything is called "insights", should the folder be called "tests/plugins/insights" too? Or is the plugin actually called query_insights?

ansjcy: Yes. The plugin is actually called `query_insights`. I added it to the CI/CD matrix as well.

.github/workflows/test-spec.yml
```diff
@@ -40,6 +40,8 @@ jobs:
             tests: plugins/streaming
           - version: 2.17.0
             tests: plugins/notifications
+          - version: 2.18.0
```
dblock, line 43: 2.18 hasn't been released so you need to have a `ref` and `hub` like below for a 2.18 build if you want to use that, but if query insights was added in 2.15 you should use 2.17, the latest stable release here.

tests/plugins/query_insights/docker-compose.yml
dblock: Cool. Iterate to green.

.github/workflows/test-spec.yml
ansjcy: Got it. Just updated to 2.17. Thank you!

dblock (approved):
👏 Good work, thanks for hanging in here with me!

spec/namespaces/insights.yaml
```diff
@@ -0,0 +1,34 @@
...
+info:
+  title: Query Insights API
+  description: API to retrieve top queries based on latency, CPU, or memory usage.
+  version: 1.0.0
+paths:
+  /_insights/top_queries:
+    get:
+      operationId: insights.top_queries.0
+      x-operation-group: insights.top_queries
+      x-version-added: '1.0'
```
dblock, line 11: I think this should be 2.15?

ansjcy: yes. Let me send out a follow up on this.

dblock:
@ansjcy I also see an intermittent CI failure, e.g. in https://github.com/opensearch-project/opensearch-api-specification/actions/runs/11487125212/job/31970965603?pr=637

I can reproduce it on a new docker start from tests/plugins/query_insights.

```
tests/plugins/query_insights $ docker compose up
```
```
npm run test:spec--insecure -- --tests tests/plugins/query_insights/insights/top_queries.yaml --verbose
```

```
        FAILED  Retrieve top queries by memory usage.
            PASSED  PARAMETERS
                PASSED  type
            PASSED  REQUEST BODY
            PASSED  RESPONSE STATUS
            PASSED  RESPONSE PAYLOAD BODY
            FAILED  RESPONSE PAYLOAD SCHEMA (data/top_queries/2/source/query/bool/adjust_pure_negative property is not defined in the spec)
```

Here's the response:

```
"top_queries": [
    {
      "timestamp": 1729713532744,
      "source": {
        "size": 10000,
        "query": {
          "bool": {
            "must_not": [
              {
                "exists": {
                  "field": "source",
                  "boost": 1
                }
              }
            ],
            "adjust_pure_negative": true,
            "boost": 1
          }
        }
      },
      "phase_latency_map": {
        "expand": 0,
        "query": 53,
        "fetch": 2
      },
      "search_type": "query_then_fetch",
      "indices": [
        ".opensearch-sap-log-types-config"
      ],
      "node_id": "G9AF1X_gQtWHg5VZjUVgtw",
      "total_shards": 1,
      "labels": {},
      "task_resource_usages": [
        {
          "action": "indices:data/read/search",
          "taskId": 43,
          "parentTaskId": -1,
          "nodeId": "G9AF1X_gQtWHg5VZjUVgtw",
          "taskResourceUsage": {
            "cpu_time_in_nanos": 1731501,
            "memory_in_bytes": 97912
          }
        }
      ],
```

PR #630, opened 2024-10-18
Tests for _core refresh.yaml remote_info.yaml script_context.yaml 

brunoLloret:
### Description
Tests for _core
refresh.yaml
remote_info.yaml
script_context.yaml

### Issues Resolved

### DIFF
--- a/tests/default/_core/refresh.yaml
+++ b/tests/default/_core/refresh.yaml
@@ -1,0 +1,10 @@
+$schema: ../../../json_schemas/test_story.schema.yaml
+
+description: Test _refresh.
+
+chapters:
+  - synopsis: Refresh all indices in the cluster.
+    path: /_refresh
+    method: GET
+    response:
+      status: 200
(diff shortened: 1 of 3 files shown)


tests/default/_core/refresh.yaml
```diff
@@ -0,0 +1,10 @@
+$schema: ../../../json_schemas/test_story.schema.yaml
+
+description: Test _refresh.
+
+chapters:
+  - synopsis: Refresh all indices in the cluster.
+    path: /_refresh
```
dblock, line 7: We should remove it from [here](https://github.com/opensearch-project/opensearch-api-specification/blob/main/tests/default/indices/refresh.yaml#L17).

tests/default/_core/remote_info.yaml
```diff
@@ -0,0 +1,10 @@
+$schema: ../../../json_schemas/test_story.schema.yaml
```
dblock, line 1: The endpoint is remote, so rename this to `remote/info.yaml`.

dblock (changes requested):
Thanks, see below.

See if you can add the parameters these support in the next iteration

dblock:
I finished this in https://github.com/opensearch-project/opensearch-api-specification/pull/637.

PR #639, opened 2024-10-24
Add spec for cluster stats metric and index_metric filter paths.

SwethaGuptha:
### Description
Adding specs for path` /_cluster/stats/{metric}/nodes/{nodeId}` and `/_cluster/stats/{metric}/{index_metric}/nodes/{nodeId}` in cluster/stats API. With https://github.com/opensearch-project/OpenSearch/pull/15938, the cluster/stats API will have filtering support that will allow the clients to fetch only specific metrics from cluster/stats instead of all metrics.

### Issues Resolved

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -129,6 +129,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Added API spec for `adjust_pure_negative` for bool queries ([#641](https://github.com/opensearch-project/opensearch-api-specification/pull/641))
 - Added a spec style checker [#620](https://github.com/opensearch-project/opensearch-api-specification/pull/620).
 - Added `remote_store` to node `Stats` ([#643](https://github.com/opensearch-project/opensearch-api-specification/pull/643))
+- Added `/_cluster/stats/{metric}/nodes/{node_id}` and `/_cluster/stats/{metric}/{index_metric}/nodes/{node_id}` ([#639](https://github.com/opensearch-project/opensearch-api-specification/pull/639))
 
 ### Changed
 
--- a/spec/namespaces/cluster.yaml
+++ b/spec/namespaces/cluster.yaml
@@ -331,6 +331,39 @@ paths:
       responses:
         '200':
           $ref: '#/components/responses/cluster.stats@200'
+  /_cluster/stats/{metric}/nodes/{node_id}:
+    get:
+      operationId: cluster.stats.2
+      x-operation-group: cluster.stats
+      x-version-added: '2.18'
+      description: Returns high-level overview of cluster statistics.
+      externalDocs:
+        url: https://opensearch.org/docs/latest/api-reference/cluster-api/cluster-stats/
+      parameters:
+        - $ref: '#/components/parameters/cluster.stats::path.metric'
+        - $ref: '#/components/parameters/cluster.stats::path.node_id'
+        - $ref: '#/components/parameters/cluster.stats::query.flat_settings'
+        - $ref: '#/components/parameters/cluster.stats::query.timeout'
+      responses:
+        '200':
+          $ref: '#/components/responses/cluster.stats@200'
+  /_cluster/stats/{metric}/{index_metric}/nodes/{node_id}:
+    get:
+      operationId: cluster.stats.3
+      x-operation-group: cluster.stats
+      x-version-added: '2.18'
+      description: Returns high-level overview of cluster statistics.
+      externalDocs:
+        url: https://opensearch.org/docs/latest/api-reference/cluster-api/cluster-stats/
+      parameters:
+        - $ref: '#/components/parameters/cluster.stats::path.index_metric'
+        - $ref: '#/components/parameters/cluster.stats::path.metric'
+        - $ref: '#/components/parameters/cluster.stats::path.node_id'
+        - $ref: '#/components/parameters/cluster.stats::query.flat_settings'
+        - $ref: '#/components/parameters/cluster.stats::query.timeout'
+      responses:
+        '200':
+          $ref: '#/components/responses/cluster.stats@200'
   /_cluster/voting_config_exclusions:
     post:
       operationId: cluster.post_voting_config_exclusions.0
@@ -1380,6 +1413,27 @@ components:
       schema:
         $ref: '../schemas/_common.yaml#/components/schemas/Duration'
       style: form
+    cluster.stats::path.index_metric:
+      in: path
+      name: index_metric
+      description: Limit the information returned for indexes metric to the specific index metrics. It can be used only if indexes (or all) metric is specified.
+      required: true
+      schema:
+        type: array
+        items:
+          $ref: '../schemas/cluster.stats.yaml#/components/schemas/IndexMetric'
+      style: simple
+    cluster.stats::path.metric:
+      in: path
+      name: metric
+      description: Limit the information returned to the specified metrics
+      required: true
+      schema:
+        type: array
+        items:
+          $ref: '../schemas/cluster.stats.yaml#/components/schemas/Metric'
+      style:
+        simple
     cluster.stats::path.node_id:
       in: path
       name: node_id
--- a/spec/schemas/cluster.stats.yaml
+++ b/spec/schemas/cluster.stats.yaml
@@ -27,8 +27,6 @@ components:
           required:
             - cluster_name
             - cluster_uuid
-            - indices
-            - nodes
             - status
             - timestamp
     ClusterIndices:
@@ -60,17 +58,6 @@ components:
           type: array
           items:
             $ref: '#/components/schemas/IndicesVersions'
-      required:
-        - analysis
-        - completion
-        - count
-        - docs
-        - fielddata
-        - mappings
-        - query_cache
-        - segments
-        - shards
-        - store
     CharFilterTypes:
       type: object
       properties:
@@ -271,18 +258,6 @@ components:
           type: array
           items:
             $ref: '_common.yaml#/components/schemas/VersionString'
-      required:
-        - count
-        - discovery_types
-        - fs
-        - ingest
-        - jvm
-        - network_types
-        - os
-        - packaging_types
-        - plugins
-        - process
-        - versions
     ClusterNodeCount:
       type: object
       properties:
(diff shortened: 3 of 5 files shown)


dblock (changes requested):
Looking good.

1. Bump the SHA for a more recent build of 2.18 and 3.0 in https://github.com/opensearch-project/opensearch-api-specification/blob/main/.github/workflows/test-spec.yml#L45 until tests pass.
2. Add to CHANGELOG.

dblock:
@SwethaGuptha Looks like the build has other problems (missing other plugins like SQL and ISM). This is a known and rather annoying problem, 3.x builds don't always have everything. You should chase what's broken in OpenSearch main distribution.

I opened https://github.com/opensearch-project/opensearch-build/issues/5130 to see what the infra team says about this (cc: @gaiksaya).


@gaiksaya

SwethaGuptha:
> @SwethaGuptha Looks like the build has other problems (missing other plugins like SQL and ISM). This is a known and rather annoying problem, 3.x builds don't always have everything. You should chase what's broken in OpenSearch main distribution.
> 
> I opened [opensearch-project/opensearch-build#5130](https://github.com/opensearch-project/opensearch-build/issues/5130) to see what the infra team says about this (cc: @gaiksaya).
> 
> @gaiksaya

Thanks @dblock for creating the issue. Did the tests work for v#3.0.0 previously? I see there are some failures from core OS as well. Also,I see the latest image available for v#3.0 is from eight days ago, do you know what cadence is followed for updating the version image?

gaiksaya:
> I opened [opensearch-project/opensearch-build#5130](https://github.com/opensearch-project/opensearch-build/issues/5130) to see what the infra team says about this (cc: @gaiksaya).
> 
> @gaiksaya

Will add more details to the issue but docker image is build using tarball. And all the distributions contains [distribution manifest](https://github.com/opensearch-project/opensearch-build/wiki/Building-an-OpenSearch-and-OpenSearch-Dashboards-Distribution#what-are-manifests) with all the information about a distribution and the components bundled in that distribution.

dblock:
> > I opened [opensearch-project/opensearch-build#5130](https://github.com/opensearch-project/opensearch-build/issues/5130) to see what the infra team says about this (cc: @gaiksaya).
> > @gaiksaya
> 
> Will add more details to the issue but docker image is build using tarball. And all the distributions contains [distribution manifest](https://github.com/opensearch-project/opensearch-build/wiki/Building-an-OpenSearch-and-OpenSearch-Dashboards-Distribution#what-are-manifests) with all the information about a distribution and the components bundled in that distribution.

It's just hard to tell from a distribution what the docker SHA is.

dblock:
@SwethaGuptha yes, update the SHAs to the latest available for 2.18 and 3.0 and see if all tests pass - it has to pass 2.18, but you can mark tests as < 3.0 if that one doesn't pass, for now

dblock:
Now that 2.18 has been released this should be easier.

SwethaGuptha:
@dblock Can you please help with the review?

CHANGELOG.md
```diff
@@ -129,6 +129,8 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Added API spec for `adjust_pure_negative` for bool queries ([#641](https://github.com/opensearch-project/opensearch-api-specification/pull/641))
 - Added a spec style checker [#620](https://github.com/opensearch-project/opensearch-api-specification/pull/620).
 - Added `remote_store` to node `Stats` ([#643](https://github.com/opensearch-project/opensearch-api-specification/pull/643))
+- Added spec for _cluster/stats metric and index_metric filter paths([#639](https://github.com/opensearch-project/opensearch-api-specification/pull/639))
```
dblock, line 132:
```suggestion
- Added `/_cluster/stats/{metric}/nodes/{node_id}` and `/_cluster/stats/{metric}/{index_metric}/nodes/{node_id}` ([#639](https://github.com/opensearch-project/opensearch-api-specification/pull/639))
```

spec/schemas/cluster.stats.yaml
```diff
@@ -60,17 +58,7 @@ components:
...
-        - completion
-        - count
-        - docs
-        - fielddata
-        - mappings
-        - query_cache
-        - segments
-        - shards
-        - store
+      required: []
```
dblock, line 61: If none of these are required just remove the `required` entirely.

spec/schemas/cluster.stats.yaml
```diff
@@ -271,18 +259,7 @@ components:
...
-        - fs
-        - ingest
-        - jvm
-        - network_types
-        - os
-        - packaging_types
-        - plugins
-        - process
-        - versions
+      required: []
```
dblock, line 262:
```suggestion

```

dblock (changes requested):
Looks good, some cosmetic asks, please.

CHANGELOG.md
SwethaGuptha: Updated.

CHANGELOG.md
```diff
@@ -129,6 +129,8 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Added API spec for `adjust_pure_negative` for bool queries ([#641](https://github.com/opensearch-project/opensearch-api-specification/pull/641))
 - Added a spec style checker [#620](https://github.com/opensearch-project/opensearch-api-specification/pull/620).
 - Added `remote_store` to node `Stats` ([#643](https://github.com/opensearch-project/opensearch-api-specification/pull/643))
+- Added `/_cluster/stats/{metric}/nodes/{node_id}` and `/_cluster/stats/{metric}/{index_metric}/nodes/{node_id}`([#639](https://github.com/opensearch-project/opensearch-api-specification/pull/639))
```
dblock, line 132: Sorry to be annoying, a space is missing and the line below has an extra `-`.

```suggestion
- Added `/_cluster/stats/{metric}/nodes/{node_id}` and `/_cluster/stats/{metric}/{index_metric}/nodes/{node_id}` ([#639](https://github.com/opensearch-project/opensearch-api-specification/pull/639))
```

CHANGELOG.md
```diff
@@ -129,6 +129,8 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
...
+-
```
dblock, line 133:
```suggestion

```

spec/namespaces/cluster.yaml
```diff
@@ -1380,6 +1413,27 @@ components:
       schema:
         $ref: '../schemas/_common.yaml#/components/schemas/Duration'
       style: form
+    cluster.stats::path.index_metric:
+      in: path
+      name: index_metric
+      description: Limit the information returned for indices metric to the specific index metrics. It can be used only if indices (or all) metric is specified.
```
dblock, line 1419: Vale is correct here, please fix, indexes is the standard in the documentation.

dblock (changes requested):
Sorry to be annoying, some more small stuff flagged here.

spec/namespaces/cluster.yaml
SwethaGuptha: Fixed.

dblock (approved):

dblock:
Thanks for taking this on, merged!

There's also node stats, do those have to mirror cluster stats?

```
 GET /_nodes/{node_id}/stats/{metric}/{index_metric}
 GET /_nodes/{node_id}/usage
 GET /_nodes/{node_id}/usage/{metric}
 GET /_nodes/{node_id}/{metric}
```

nhtruong:
@dblock 
So `cluster.stats` will have the following paths:
- /_cluster/stats
- /_cluster/stats/nodes/{node_id}
- /_cluster/stats/{metric}/nodes/{node_id}
- /_cluster/stats/{metric}/{index_metric}/nodes/{node_id}

current `path-finding` strat for the JS CodeGenw will create:
```ts
  const path = ['/_cluster/stats/', metric, '/', index_metric, '/nodes/', node_id].filter(c => c).join('').replace('//', '/');
```
which works fine for most situations EXCEPT for when no path params are provided since it will yield `/_cluster/stats/nodes`, which is not a valid path. I've gotta update the CodeGen for JS (I really like the above solution if not for this edge case 😢 )
cc: @Xtansia in case this affects Java/.NET CodeGen too.

dblock:
@nhtruong I think these are 2 separate APIs, one is `cluster` stats (`/_cluster/stats`), and the other is `cluster/nodes` stats (the other 3), would that help?

nhtruong:
I'm referring to these 4 operations: https://github.com/opensearch-project/opensearch-api-specification/pull/639/files#diff-c68dc8d4bd22aabd667fe0978fba9a121988d77ac766b1e2a5e94091603c0734R305-R366

It's not big deal. We just can't have that one-line solution 🥲

dblock:
@nhtruong Yes, I think they should be totally different operations, no? would that not work?

nhtruong:
They are in the same `operation-group` which means they will be piped into the same API function on the client. That function has to pick a path depending on the path params provided by the user. NodeJS CodeGen is using that one-line solution which works for the vast majority of the operation groups, even the notorious `nodes.info`:
- /_nodes
- /_nodes/{node_id}
- /_nodes/{metric}
- /_nodes/{node_id}/{metric}

` const path = ['/_nodes/', node_id, '/', metric].filter(c => c).join('').replace('//', '/');`

It's not hard update this.

nhtruong:
Yes, the 4 operations I stated in https://github.com/opensearch-project/opensearch-api-specification/pull/639#issuecomment-2477036191 are in `cluster.stats`

SwethaGuptha:
>current path-finding strat for the JS CodeGen will create:
 const path = ['/_cluster/stats/', metric, '/', index_metric, '/nodes/', node_id].filter(c => c).join('').replace('//', '/');
which works fine for most situations EXCEPT for when no path params are provided since it will yield /_cluster/stats/nodes, which is not a valid path

@nhtruong Trying to understand the problem here, was never a valid path even earlier.  How is handling of two new paths causing a challenge in the Code Gen?

nhtruong:
Sorry I missed this earlier.
> Trying to understand the problem here, was never a valid path even earlier. How is handling of two new paths causing a challenge in the Code Gen?
This PR wasn't the cause. Reading through this made me realize that the paths for this operation group will cause issue for the JS CodeGen specifically because of its peculiar strategy to build the path. And this is the only group with paths like these. We also have other edge cases where the server accepts both `_alias` and `_aliases` so we endup with paths that are aliases of each other (pun intended 😄):
- /{index}/_aliases/{name}
- /{index}/_alias/{name}

Issue #614, opened 2024-10-16
[BUG]


dblock:
### What is the bug?

Flaky test.

```
ERROR   knn/train_model.yaml (tests/default/knn/train_model.yaml)
    ERROR   EPILOGUES
        PASSED  DELETE /movies
        ERROR   DELETE /_plugins/_knn/models/{model_id} (Cannot delete model [PeuoTEF5Sq-cOLVW6hivgg]. Model is still in training)
```

### How can one reproduce the bug?

https://github.com/opensearch-project/opensearch-api-specification/actions/runs/11346241610/job/31563950086?pr=611

### What is the expected behavior?

Test to always pass.

### What is your host/environment?

GHA

### Do you have any screenshots?

![Screenshot 2024-10-16 at 6 18 56 AM](https://github.com/user-attachments/assets/1bbb3e17-c1bd-4bac-af3d-6e1ade38885d)

PR #644, opened 2024-10-24
Fix: model training intermittent test failure.

dblock:
### Description

The reason this was a flake was that half the time the model would successfully begin training, and half the time would not.

The time it would not it would fail with the following error:

```
Error: 'nx >= k' failed: Number of training points (0) should be at least as large as number of clusters (4)
```

This was caused by the fact that a refresh cycle had yet to run, so the model had 0 documents to use. However when a refresh cycle would happen in time, the model would begin training. In that case it could not be deleted until training was complete, causing the CI failure.

### Issues Resolved

Closes https://github.com/opensearch-project/opensearch-api-specification/issues/614.

### DIFF
--- a/spec/namespaces/knn.yaml
+++ b/spec/namespaces/knn.yaml
@@ -283,7 +283,11 @@ components:
       required: true
   responses:
     knn.delete_model@200: {}
-    knn.get_model@200: {}
+    knn.get_model@200:
+      content:
+        application/json:
+          schema:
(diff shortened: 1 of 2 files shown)


nhtruong (approved):

nhtruong:
My motto when dealing with OpenSearch errors: "When in doubt: REFRESH!"

Issue #633, opened 2024-10-21
[BUG] OpenAPI spec validation failed on missing descriptions and name pattern


zelinh:
### What is the bug?
I tried to load the opensearch API spec YAML into a python client [openapi-spec-validator](https://github.com/python-openapi/openapi-spec-validator).
However, when running the spec validation with python client, it fails on different aspects. 
- The response component doesn't have `description` as required. https://spec.openapis.org/oas/v3.1.0#fixed-fields-14
- The component responses name doesn't match pattern `'^[a-zA-Z0-9._-]+$'`

### How can one reproduce the bug?
Download the opensearch-api-spec YAML. https://github.com/opensearch-project/opensearch-api-specification/releases/download/main-latest/opensearch-openapi.yaml
```
pip install openapi-spec-validator
```
Run this python script
```
from openapi_spec_validator import validate
from openapi_spec_validator.readers import read_from_filename

spec_file = "/.../Downloads/opensearch-openapi.yaml"
# Load your OpenAPI spec (YAML format)
spec_dict, base_uri = read_from_filename(spec_file)
# If no exception is raised by validate(), the spec is valid.
validate(spec_dict)
```
Will see error 
```
openapi_spec_validator.validation.exceptions.OpenAPIValidationError: 'description' is a required property

Failed validating 'required' in schema['properties']['components']['properties']['responses']['additionalProperties']['else']:
    {'$comment': 'https://spec.openapis.org/oas/v3.1.0#response-object',
     'type': 'object',
     'properties': {'description': {'type': 'string'},
                    'headers': {'type': 'object',
                                'additionalProperties': {'$ref': '#/$defs/header-or-reference'}},
                    'content': {'$ref': '#/$defs/content'},
                    'links': {'type': 'object',
                              'additionalProperties': {'$ref': '#/$defs/link-or-reference'}}},
     'required': ['description'],
     '$ref': '#/$defs/specification-extensions',
     'unevaluatedProperties': False}

On instance['components']['responses']['indices.create_data_stream@200']:
    {'content': {'application/json': {'schema': {'$ref': '#/components/schemas/_common:AcknowledgedResponseBase'}}}}
```
After manually updating the component description, we would see another error regarding of the component schema name issue such as `info@200/_common:Name` don't match the pattern. 

### What is the expected behavior?
The API spec yaml should be validated by spec validator for OpenAPI 3.1. 

### What is your host/environment?
MacOS, python 3.9.13

[...]

PR #646, opened 2024-10-25
Ensure correct OpenAPI 3.1.0 spec.

dblock:
### Description

Use a Python and a Ruby OpenAPI spec validator and ensure correct OpenAPI 3.1.0 spec.

### Issues Resolved

Closes https://github.com/opensearch-project/opensearch-api-specification/issues/633.


dblock:
With https://github.com/opensearch-project/opensearch-api-specification/pull/646/commits/b9b4f4ec0f8d16db066a04a684de21d6bb878c58 we go down from 4593 to 3713.

dblock:
@nhtruong @Xtansia this is green and ready to review, thx

nhtruong:
Are there discrepancies between Ruby and Python's validators where they can throw different validation errors? Just wondering why we need both.

spec/_global_parameters.yaml
```diff
@@ -10,21 +10,18 @@ components:
...
       schema:
         type: boolean
-      default: true
     error_trace:
       name: error_trace
       in: query
       description: Whether to include the stack trace of returned errors.
       schema:
         type: boolean
-      default: false
```
nhtruong, line 27: We should move the default values into their respective schemas in this case.
We should also add `x-default` extension for parameters when the default for the param differs from the default of the schema OR when a shared schema itself doesn't have a default but certain param using said schema does.

tests/default/security/api/nodesdn.yaml
```diff
@@ -1,65 +0,0 @@
-$schema: ../../../../json_schemas/test_story.schema.yaml
```
nhtruong, line 1: Why are we removing this story?

dblock:
> Are there discrepancies between Ruby and Python's validators where they can throw different validation errors? Just wondering why we need both.

Yes, they are very different codebases and throw some different errors. The Ruby one spits out all errors and was useful for the large categories like `#/paths/%2F_cat%2Fsegment_replication/get/parameters/17/%24ref: #/components/parameters/_global%3A%3Aquery.pretty does not resolve to a valid object`, while the python one picked up unresolved parameters in superseded operations that the Ruby one didn't (`Path parameter 'policyID' for 'get' operation in '/_opendistro/_ism/policies/{policyID}' was not resolved` and `Required list has not defined properties: ['order0']`).

tools/tests/merger/fixtures/extractor/opensearch/expected_1.3.yaml
```diff
@@ -12,33 +12,33 @@ paths:
...
+          $ref: '#/components/responses/info.503'
         removed-2.0:
-          $ref: '#/components/responses/info@removed-2.0'
+          $ref: '#/components/responses/info.removed-2.0'
           x-version-removed: '2.0'
         removed-2.0-refs:
-          $ref: '#/components/responses/info@removed-2.0-refs'
+          $ref: '#/components/responses/info.removed-2.0-refs'
         added-1.3-removed-2.0:
-          $ref: '#/components/responses/info@added-1.3-removed-2.0'
```
nhtruong, line 39: This gets tricky with operation-group in non-root namespaces, say `indices.create`. We will end up with `indices.create.removed-2.0`. Visually it's hard to tell which is which. Is it a response for `indices` that's named `create.remove-2.0` or, a response for `indices.create` that's named `removed-2.0`.

This is not an issue on the technical term because the CodeGen does not parse the response names (at least the ones I work don't. @Xtansia can chime in), but we loses the visual cue with this change. I'm just stating a cost of this change. It's not a blocker.

We can also replace `@` with `___` --> `indices.create___removed-2.0`

tools/tests/merger/fixtures/extractor/opensearch/expected_1.3.yaml
```diff
@@ -95,42 +97,45 @@ components:
...
-    info@removed-2.0:
+    info.removed-2.0:
       description: Removed in 2.0 via attribute next to ref.
-    info@removed-2.0-refs:
+    info.removed-2.0-refs:
       description: One of the ref values removed in 2.0.
       schema:
         oneOf:
-          - $ref: '#/components/schemas/_common:Type'
-          - $ref: '#/components/schemas/_common:OldId'
```
nhtruong, line 124: Can we also replace `:` with `___` in this case to separate the schema category from the type name? Similar to the previous comment regarding the response naming, this can get confusing with more complex schema categories say `ai_dsl.aggregation_common_Type`. It's hard to visually tell where the category ends and where the schema name starts. `ai_dsl.aggreation_common___Type` is slightly better and easier for the CodeGen to parse.

nhtruong (changes requested):

tests/default/security/api/nodesdn.yaml
dblock: See https://github.com/opensearch-project/opensearch-api-specification/issues/654, the story doesn't test anything.

tools/tests/merger/fixtures/extractor/opensearch/expected_1.3.yaml
dblock: I went with this map.

```
      .replaceAll('::', '___')
      .replaceAll('@', '__')
      .replaceAll(':', '_')
```

tools/tests/merger/fixtures/extractor/opensearch/expected_1.3.yaml
dblock: Take a look at the update, we have to replace a few things, I went with the following map.

```
      .replaceAll('::', '___')
      .replaceAll('@', '__')
      .replaceAll(':', '_')
```

spec/_global_parameters.yaml
dblock: I did the first part, didn't realize that default should/can be nested under `schema` when I was removing this.

I don't think we have a case where the default differs, do you know of one?

tools/src/merger/OpenApiMerger.ts
```diff
@@ -134,6 +137,58 @@ export default class OpenApiMerger {
...
+      }
+    }
+  }
+
+  #normalize_key(key: string): string {
+    return key
+      .replaceAll('::', '___')
+      .replaceAll('@', '__')
+      .replaceAll(':', '_')
+  }
```
nhtruong, line 160: I think it's time for a quick trip down the memory lane to learn why `::`, `@`, and `:` were chosen as separators for parameter, response, and schema names. There were 2 criteria:
 1.  The character(s) used as a separator cannot appear anywhere else in the $ref. (This makes it easier to tell different component of a name apart, and easier for CodeGen to parse)
 2. The separators for parameters, responses, and schemas must be easily distinguishable from one another. (This encoding of component type into the separators helps the reader quickly identify where they are in the giant YAML file, and easier to grep for a component definition.

We can ignore criterion No.2 but I think No.1 is worth upholding. So, `.replaceAll(':', '_')` is not good. We can use `___` as the separator for all 3 types of `#components` if we let go of criteria No.2 OR come up with a different set of separators that still conform to strict OpenAPI 3.1's naming rules. Say `___`, `---`, and `...`.

spec/_global_parameters.yaml
nhtruong: Haven't come across on where the schema itself has a default that is different from the param that uses it. But I've come across schemas that are used in different params of different operations but have different default values depending on the operation. But That was a long time ago, before we moved onto OpenAPI. But that's still a real possibility.

tools/src/merger/OpenApiMerger.ts
dblock: I believe we're only parsing `::` and `@` and `:` is pure decoration (otherwise `::` vs. `:` would be a problem already). So, with this approach you can split by 3 `___` to get the equivalent of `::` and by 2 `__` to get the equivalent of `@`, as long as you parse in that order. 

I could definitely use things like `___` vs. `---`, but I find them rather ugly :) WDYT?

dblock: I like using `___` as the separator better, even though technically someone could create a key like `___` to mess with the merger, I don't think it's a real scenario. 

I don't know whether we want to just replace all keys in the spec, it's nice to use a `::` and `@` convention for readability/developer productivity.

Updated PR with `___`.

nhtruong (approved):

Xtansia (approved):

Xtansia:
Seems good to me, nothing that'll be a major issue for .NET or Java codegen

rursprung:
FYI: this broke a link in the README which will now block all PRs based on this until resolved. i've created a fix for it: #657

PR #649, opened 2024-10-29
add workload management querygroup api specs

ruai0511:
### Description
API spec for workload management, more specifically, for the query group lifecycle APIs.
Related PR: https://github.com/opensearch-project/OpenSearch/pull/14680, https://github.com/opensearch-project/OpenSearch/pull/14775/, https://github.com/opensearch-project/OpenSearch/pull/14735, https://github.com/opensearch-project/OpenSearch/pull/14709


ruai0511:
Workload Management plugin is only available in 2.17.0, but when I use the default ${OPENSEARCH_VERSION:-latest} in docker-compose file, it only gives 2.14.0. So I hardcoded 2.17.0 in it. I'm wondering if there's any cleaner ways of doing it?

spec/namespaces/query_group.yaml
```diff
@@ -0,0 +1,127 @@
+openapi: 3.1.0
```
dblock, line 1: I think this should be `wlm.yaml` to group all WLM APIs together.

tests/plugins/wlm/query_group/query_group.yaml
```diff
@@ -0,0 +1,67 @@
+$schema: ../../../../json_schemas/test_story.schema.yaml
```
dblock, line 1: Since the API is `/_wlm/query_group` this should be in `tests/default/wlm/query_group.yaml`.

dblock (changes requested):
Thanks!

The docker compose in `tests/plugins/wlm` is identical to the default one. We do custom `plugins/xyz` for when additional setup is required, so move this into `tests/default` and you don't need a special entry in the CI matrix.

ruai0511:
> Thanks!
> 
> The docker compose in `tests/plugins/wlm` is identical to the default one. We do custom `plugins/xyz` for when additional setup is required, so move this into `tests/default` and you don't need a special entry in the CI matrix.

The default docker compose uses ${OPENSEARCH_VERSION:-latest} which is 2.14.0, but we need 2.17.0 here

dblock:
> > Thanks!
> > The docker compose in `tests/plugins/wlm` is identical to the default one. We do custom `plugins/xyz` for when additional setup is required, so move this into `tests/default` and you don't need a special entry in the CI matrix.
> 
> The default docker compose uses ${OPENSEARCH_VERSION:-latest} which is 2.14.0, but we need 2.17.0 here. That's why I put the tests in plugin since we need Dockerfile (to install the workload-management plugin) and docker compose with version 2.17.0

You should upgrade that number, and it's 2.17 in CI: https://github.com/opensearch-project/opensearch-api-specification/blob/main/.github/workflows/test-spec.yml

ruai0511:
> > > Thanks!
> > > The docker compose in `tests/plugins/wlm` is identical to the default one. We do custom `plugins/xyz` for when additional setup is required, so move this into `tests/default` and you don't need a special entry in the CI matrix.
> > 
> > 
> > The default docker compose uses ${OPENSEARCH_VERSION:-latest} which is 2.14.0, but we need 2.17.0 here. That's why I put the tests in plugin since we need Dockerfile (to install the workload-management plugin) and docker compose with version 2.17.0
> 
> You should upgrade that number, and it's 2.17 in CI: https://github.com/opensearch-project/opensearch-api-specification/blob/main/.github/workflows/test-spec.yml

wlm is not installed by default. So we still need a separate dockerfile to install it.

dblock:
> > > > Thanks!
> > > > The docker compose in `tests/plugins/wlm` is identical to the default one. We do custom `plugins/xyz` for when additional setup is required, so move this into `tests/default` and you don't need a special entry in the CI matrix.
> > > 
> > > 
> > > The default docker compose uses ${OPENSEARCH_VERSION:-latest} which is 2.14.0, but we need 2.17.0 here. That's why I put the tests in plugin since we need Dockerfile (to install the workload-management plugin) and docker compose with version 2.17.0
> > 
> > 
> > You should upgrade that number, and it's 2.17 in CI: https://github.com/opensearch-project/opensearch-api-specification/blob/main/.github/workflows/test-spec.yml
> 
> wlm is not installed by default. So we still need a separate dockerfile to install it and it should be in the plugins directory

Sorry to be a pest, the plugin is called `workload-management` not `wlm` so let's make the folder `tests/plugins/workload-management`? The namespace is correct, since the route is `wlm`.

dblock (approved):
Good work! Thanks for hanging in here with me.

CHANGELOG.md
```diff
@@ -106,6 +106,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Added `config_id` and `config_id_list` to `/_plugins/_notifications/configs` query parameters ([#594](https://github.com/opensearch-project/opensearch-api-specification/pull/594))
 - Added a release workflow triggered on a tag ([#635](https://github.com/opensearch-project/opensearch-api-specification/pull/635))
 - Added API spec for query insights plugin ([#625](https://github.com/opensearch-project/opensearch-api-specification/pull/625))
+- Added API specs for query groups lifecycle APIs ([#649](https://github.com/opensearch-project/opensearch-api-specification/pull/649))
```
dblock, line 109: @ruai0511 I missed this, but we released 0.1.0, could you please move this in another PR to the unreleased section? thanks

PR #666, opened 2024-11-12
Correct index mapping types

Xtansia:
### Description
Correct index mapping types


spec/schemas/_common.mapping.yaml
```diff
@@ -109,6 +109,7 @@ components:
       type: object
       discriminator:
         propertyName: type
+        x-default: 'object'
```
dblock, line 112: Remove quotes, linter is right.

spec/schemas/_common.mapping.yaml
```diff
@@ -843,20 +743,22 @@ components:
             - type
     GeoOrientation:
       oneOf:
+        # eslint-disable yml/sort-sequence-values
```
dblock, line 746: Does the order here matter, otherwise remove `eslint-disable`.

dblock (changes requested):
Fix linter.

Add a test, at least one that exercises the added `IcuCollationKeywordProperty`?

dblock:
Rebase?

Also in https://github.com/opensearch-project/opensearch-api-specification/pull/609 @rursprung is adding an analysis plugin test like this one. Should we combine them into one analysis folder or not?

spec/schemas/_common.mapping.yaml
Xtansia: They're in preference order as left & right are the true values and cw/ccw are the aliases

dblock (approved):

Issue #653, opened 2024-11-04
[PROPOSAL] Protobuf Automation 


amberzsy:
### What are you proposing?
There are bunch gaps for having E2E protobuf and API spec release: 
1. Open api spec not in sync with core API and lack of automation and enforcement 
2. No automation and tooling for protobuf generation from api spec. 
3. Lack of validation on protobuf and api spec 

#### Development flow

- There will be a new repo called opensearch-api-protobufs, it can be nested as a submodule into opensearch core repo or separate repo similar to opensearch-api-specification 

- As part of grpc new api development,  developer manually update the protobuf repo, and update their local submodule configs to point to their protobuf changes. If it’s for adding corresponding grpc endpoint per http api, use https://github.com/dblock/opensearch-api to produce API spec at runtime and trigger spec<>protobuf conversion to generate protobuf definition. 

- During PR on opensearch core repo, there would be multiple workflow/job triggered on merge action, 

Workflow 1: Trigger spec generation job with versioning annotations, it would generate a PR in opensearch-api-spec repo. 

- Tooling: https://github.com/dblock/opensearch-api

Workflow 2: trigger ‘merge to protobuf’ which generate protobuf from the spec. 
Tooling: spec <> proto generator (WIP)

#### Linter support 

- Check the protobuf which is manually written is matched to what it will be generated by the spec. It would help with

1. Developer needs to manually write proto for grpc endpoint, linter would make sure it’s aligned with spec<>proto conversion rules.
2. Automatically generate Spec -> protos for the PR
3. Block the PR if manual proto definition does not align with Spec. 
4. Corner cases: api only support in grpc or http endpoint.

dblock:
Thanks for opening this. 

I think we should resolve what the source of truth for OpenSearch would be with this proposal. My understanding was (still is :)) that the spec in this repo is the OpenSearch API truth. While today it's reverse engineering the existing OpenSearch API because of the legacy, we see a future where the OpenSearch server-side API is generated from spec as well. If this is true, then protobufs are just another output from the source of truth, aka a build of API spec in this repo that outputs .proto files.

WDYT?

amberzsy:
> Thanks for opening this.
> 
[...]

PR #676, opened 2024-11-15
532 publish proto files

### LINKED ISSUE #532: [PROPOSAL] Publish protobuf files.
### What are you proposing?

Coming from https://github.com/opensearch-project/OpenSearch/pull/15411 and https://github.com/opensearch-project/OpenSearch/issues/15190 I propose we generate and publish protobuf files as part of the [build/release workflow](https://github.com/opensearch-project/opensearch-api-specification/blob/main/.github/workflows/build.yml).

### What users have asked for this feature?

OpenSearch is trying to use the OpenAPI spec to generate .proto files then use those on both client and server sides. 

### What problems are you trying to solve?

OpenSearch server wants .proto files.
OpenSearch clients that will implement gRPC/protobuf want .proto files.

### What is the developer experience going to be?

A release of opensearch-api-specification will contain a .zip with .proto files.

### Why should it be built? Any reason not to?

The right place for .proto files is this repo because multiple consumers want .proto files.

amberzsy:
### Description
publish example protobufs generated with semi-automation tooling

TODO: 
1. add support for compile protos. (make sure no syntax error etc)
2. generate protobuf language clients in spec repo. 

after second thought on [proposal](https://github.com/opensearch-project/opensearch-api-specification/issues/653), submodule not optimal ideas. 
probably, we should do 
1. Raw proto (which generated from Spec with certain tooling) defined in /protobuf. it should be machine generated files. But for now, i have it as handwritten ones to unblock grpc poc effort. 
2. protobuf compile should compile the raw proto into *.proto (or some jars) in the `release`. 
3. similar to opensearch-java which download https://github.com/opensearch-project/opensearch-java/blob/c7103d87423e44aa5ef57bb940496a6e3b74ff78/java-codegen/build.gradle.kts#L75 the output yaml file from release, the Core and OpenSearch-Java or other language client can download/pull from spec repo.    

cc: @karenyrx @dblock for review and better idea. 

### Issues Resolved
#532 


dblock:
Were these authored by hand or a result of a tool? 

I'd like to merge a little bit more than this, some kind of useful MVP, that for example contains these files and just enough README information of what they are, maybe some CI.

Issue #663, opened 2024-11-11
[BUG] Add missing API Spec Tests


dblock:
### What is the bug?

API specs test coverage using output added in https://github.com/opensearch-project/opensearch-api-specification/pull/697.

- [x] GET /
- [x] HEAD /
- [x] GET /_alias
- [x] PUT /_alias
- [x] GET /_alias/{name}
- [x] HEAD /_alias/{name}
- [x] POST /_alias/{name}
- [x] PUT /_alias/{name}
- [x] POST /_aliases
- [x] POST /_aliases/{name}
- [x] PUT /_aliases/{name}
- [x] GET /_analyze
- [x] POST /_analyze
- [x] POST /_bulk
- [x] PUT /_bulk
- [x] POST /_bulk/stream
- [x] PUT /_bulk/stream
- [x] POST /_cache/clear
- [x] GET /_cat
- [x] GET /_cat/aliases
- [x] GET /_cat/aliases/{name}
- [x] GET /_cat/allocation
- [x] GET /_cat/allocation/{node_id}
- [x] GET /_cat/cluster_manager
- [x] GET /_cat/count
- [x] GET /_cat/count/{index}
- [x] GET /_cat/fielddata
- [x] GET /_cat/fielddata/{fields}
- [x] GET /_cat/health
- [x] GET /_cat/indices
- [x] GET /_cat/indices/{index}
- [x] GET /_cat/master
- [x] GET /_cat/nodeattrs
- [x] GET /_cat/nodes
- [x] GET /_cat/pending_tasks
- [x] GET /_cat/pit_segments
- [x] GET /_cat/pit_segments/_all
- [x] GET /_cat/plugins
- [x] GET /_cat/recovery
- [x] GET /_cat/recovery/{index}
- [x] GET /_cat/repositories
- [x] GET /_cat/segment_replication
- [x] GET /_cat/segment_replication/{index}
- [x] GET /_cat/segments
- [x] GET /_cat/segments/{index}
- [x] GET /_cat/shards
- [x] GET /_cat/shards/{index}
- [x] GET /_cat/snapshots
- [x] GET /_cat/snapshots/{repository}
- [x] GET /_cat/tasks
- [x] GET /_cat/templates
- [x] GET /_cat/templates/{name}
- [x] GET /_cat/thread_pool
- [x] GET /_cat/thread_pool/{thread_pool_patterns}
- [x] GET /_cluster/allocation/explain
[...]

PR #678, opened 2024-11-17
Add missing Index API Spec Tests, Part 1

Tokesh:
### Description
Adding tests for this API list:
  POST /{index}/_mapping
  GET /{index}/_mapping/field/{fields}
  POST /{index}/_mget
  GET /{index}/_msearch
  POST /{index}/_msearch
  GET /{index}/_validate/query
  POST /{index}/_validate/query
### Issues Resolved
[#663 ]


Tokesh:
As soon as my PR with replication is merged, I will rebase it so that it doesn't interfere with checking
In the meantime, I made it a draft because I am still in the process of adding tests. I will try to do them as quickly as possible)

tests/default/indices/mapping.yaml
```diff
@@ -146,3 +146,40 @@ chapters:
...
+        properties:
+          genre:
+            type: text
+          rating:
+            type: float
+    response:
+      status: 200
+      payload:
+        acknowledged: true
+  - synopsis: Get specific field mappings for an index.
```
dblock, line 165: Consider splitting tests for different paths into their own scenarios if there's no duplication, e.g. this one would be `tests/default/indices/mapping/field.yaml`.

tests/default/indices/msearch.yaml
```diff
@@ -0,0 +1,124 @@
...
+    method: POST
+    parameters:
+      index: books
+      id: '3'
+    request:
+      payload:
+        name: Book 3
+        age: 30
+    response:
+      status: 201
```
dblock, line 74: These should be in a prologue. You can also use `_bulk` from other examples which makes it much shorter.

tests/default/indices/msearch.yaml
```diff
@@ -0,0 +1,124 @@
+$schema: ../../../json_schemas/test_story.schema.yaml
+
+description: Test endpoints relevant to the lifecycle of an index, including multi-get and multi-search operations.
+epilogues:
+  - path: /books
+    method: DELETE
+    status: [200, 404]
+  - path: /games
```
dblock, line 8: No games index in this test, remove.

tests/default/indices/msearch.yaml
```diff
@@ -0,0 +1,124 @@
...
+      payload:
+        mappings:
+          properties:
+            name:
+              type: keyword
+            age:
+              type: integer
+        settings:
+          number_of_shards: 5
+          number_of_replicas: 2
```
dblock, line 27: I don't think number of shards and replicas is material for this test, remove. If msearch works the same without a mapping, as well.

tests/default/indices/msearch.yaml
```diff
@@ -0,0 +1,124 @@
...
+      index: books
+      id: '3'
+    request:
+      payload:
+        name: Book 3
+        age: 30
+    response:
+      status: 201
+
+  - synopsis: Perform a `_mget` request to retrieve documents by IDs.
```
dblock, line 76: This belongs in mget.yaml.

tests/default/indices/refresh.yaml
```diff
@@ -37,3 +37,27 @@ chapters:
       payload:
         _shards:
           failed: 0
+  - synopsis: Adding one more document.
```
dblock, line 40: We're not trying to test the server, but the API schema. I don't think this adds anything as we already call `_refresh` above with the same results.

dblock (changes requested):
Thanks, some asks below.

Tokesh:
> Thanks, some asks below.

Thank you for review, I have resolved the conversations and added ```/{index}/_validate/query``` API specs

tests/default/indices/validate_query.yaml
```diff
@@ -0,0 +1,49 @@
+$schema: ../../../json_schemas/test_story.schema.yaml
```
dblock, line 1: This file should be `validate/query.yaml`.

dblock (changes requested):
One file name change, sorry to be annoying.

tests/default/indices/validate_query.yaml
Tokesh: no worries! i'm really interested in working with OpenSearch in the form of tests or writing specs to understand it more deeply.

dblock (approved):

PR #682, opened 2024-11-19
Misc Vale fixes.

dblock:
### Description

Getting Vale violations to a manageable number (60 or so).

### DIFF
--- a/spec/schemas/_common.aggregations.yaml
+++ b/spec/schemas/_common.aggregations.yaml
@@ -954,7 +954,7 @@ components:
             to_as_string:
               type: string
             key:
-              description: The bucket key. Present if the aggregation is _not_ keyed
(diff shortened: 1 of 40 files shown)


spec/schemas/_common.aggregations.yaml
```diff
@@ -954,7 +954,7 @@ components:
             to_as_string:
               type: string
             key:
-              description: The bucket key. Present if the aggregation is _not_ keyed
+              description: The bucket key. Present if the aggregation is not keyed.
```
Xtansia, line 957: This does lose the emphasis on "not", is there someway to get Vale to accept markdownisms or is there an explicit reason to deny this?

dblock: This is the only place we do this AFAIK, I decided to remove it.

Xtansia (approved):

PR #690, opened 2024-11-24
Index `Upgrade` API specs and tests for them

Tokesh:
### Description
API specs and tests for upgrade api
And marking them deprecated
GET /_upgrade
POST /_upgrade
GET /{index}/_upgrade
POST /{index}/_upgrade

### Issues Resolved
[#663]


dblock (commented):
Let's find out when this was deprecated and why and put a real version number here. What message do you get (aka how do you know it's deprecated)? Look for it in https://github.com/opensearch-project/opensearch and narrow down the commit / PR that did it.

tests/default/indices/upgrade.yaml
```diff
@@ -0,0 +1,40 @@
...
+      refresh: true
+    request:
+      content_type: application/x-ndjson
+      payload:
+        - {create: {_index: movies, _id: movie1}}
+        - {director: Bennett Miller, title: The Cruise, year: 1998}
+        - {create: {_index: movies, _id: movie2}}
+        - {director: Nicolas Winding Refn, title: Drive, year: 1960}
+chapters:
+  - synopsis: Check index upgrade status (GET).
```
dblock, line 21: Probably good to flip this one with the triggering of an upgrade so we can get actual results and check schema.

Tokesh:
> Let's find out when this was deprecated and why and put a real version number here. What message do you get (aka how do you know it's deprecated)? Look for it in https://github.com/opensearch-project/opensearch and narrow down the commit / PR that did it.

Hello! I found that, the _upgrade API was deprecated in [Elasticsearch 7.10](https://www.elastic.co/guide/en/elasticsearch/reference/7.10/indices-upgrade.html), which is precisely the version OpenSearch [was forked from](https://opensearch.org/faq/#q1.8).

tests/default/indices/upgrade.yaml
Tokesh: If I understood you correctly, then I have corrected

tests/default/indices/upgrade.yaml
```diff
@@ -0,0 +1,57 @@
...
+
+  - synopsis: Check index upgrade status (GET).
+    path: /{index}/_upgrade
+    method: GET
+    parameters:
+      index: [movies]
+    response:
+      status: 200
+
+  - synopsis: Trigger upgrade (POST).
```
dblock, line 42: Note that this is not an `/{index}` API and should go in its own `tests/default/_core/upgrade.yaml`. I'll merge now, but appreciate it being split in a future PR.

dblock (approved):

Issue #234, opened 2024-04-11
[FEATURE] Add specs for `sm` namespace


nhtruong:
Part of [Adding missing API specs issue](https://github.com/opensearch-project/opensearch-api-specification/issues/168)

- [x] /_plugins/_sm/policies/: ["GET"]
- [x] /_plugins/_sm/policies/{policyName}: ["GET","PUT","POST","DELETE"]
- [x] /_plugins/_sm/policies/{policyName}/_explain: ["GET"]
- [x] /_plugins/_sm/policies/{policyName}/_start: ["POST"]
- [x] /_plugins/_sm/policies/{policyName}/_stop: ["POST"]

aabeshov:
Hi @nhtruong , I'd like to work on this issue if it's still relevant.

dblock:
It is @aabeshov! Thanks.

aabeshov:
Can it be closed ? @nhtruong

PR #701, opened 2024-11-30
SnapshotManagement namespace api

aabeshov:
### Description
Adding snapshot management(sm) namespace specs.

### Issues Resolved
_List any issues this PR will resolve, e.g. Closes [#234 ]._ 


aabeshov:
Not ready yet

aabeshov:
Now, it is ready for review!

tests/plugins/snapshot_management/snapshot_management.yaml
```diff
@@ -0,0 +1,131 @@
+$schema: ../../../json_schemas/test_story.schema.yaml
```
dblock, line 1: This doesn't run unless you add it explicitly into CI.

dblock (changes requested):
Thanks.

1. Tests aren't running, the `plugins` directory is misleading and is there for custom docker setups and is explicitly added to CI with different configuration. We already have https://github.com/opensearch-project/opensearch-api-specification/tree/main/tests/snapshot, so move the tests in there.
2. Keep naming to the route naming, just like we don't name "ml" as "machine_learning", keep "sm" as is, so "sm._common.yaml", etc.
3. Rename the test files to match the routes, so `_sm/policies` becomes `tests/snapshot/sm/policies.yaml`.
4. Split tests and group by API not by theme. So `explain` tests should go into their own file with a prologue that creates a policy, then the chapter only calls explain. The file would be `sm/policies/explain.yaml`, `start.yaml`, etc.

tests/plugins/snapshot_management/snapshot_management.yaml
nhtruong: We should be able to detect this issue automatically and fail CI when some tests are never run. I totally didn't know about explicitly registering a test folder until now. I always assumed that if a docker file is not detected, it will look for the closest docker file in the parent folders.

aabeshov: @dblock aah, sorry I did it again
Last time I was doing observability api and I saw 'plugin' in URI ,I also put It in the different directory
Getting confused all the time
I'll try to finish it by the end of the week

dblock: I added #705 yesterday :)

aabeshov:
@dblock I am not sure what's wrong
Should I ignore version <= 2.0.0

dblock (changes requested):
Looking good!

What's the failure in 2.0? I see some `x-version-added: 2.1` so this means the feature was added in 2.1? Change all tests to do `>= 2.1`, but do dig through the history/code to verify which version the feature was added in.

Iterate to green. Spec validation needs to pass, there are a bunch of language things like quoting variables that do matter since we're using these in the docs.

spec/namespaces/sm.yaml
```diff
@@ -0,0 +1,325 @@
...
+      responses:
+        '201':
+          $ref: '#/components/responses/sm.create_policy@201'
+        '400':
+          $ref: '#/components/responses/sm.create_policy@400'
+    put:
+      operationId: sm.update_policy.0
+      x-operation-group: sm.update_policy
+      x-version-added: '2.1'
+      description: Updates an existing snapshot management policy. Requires 'if_seq_no' and 'if_primary_term'.
```
dblock:
```suggestion
      description: Updates an existing snapshot management policy. Requires `if_seq_no` and `if_primary_term`.
```

dblock (approved):

PR #723, opened 2024-12-10
Add support for multiple test verbs and fix /_render/template.

dblock:
### Description

You can now do this:

```
- synopsis: Use POST and PUT interchangeably.
  path: /{index}
  method:
    - POST
    - PUT
```

Avoids duplicating the test.

Added tests and fixed `/_render/template`. 

### Issues Resolved

Part of #663.

### DIFF
--- a/TESTING_GUIDE.md
+++ b/TESTING_GUIDE.md
@@ -12,6 +12,7 @@
       - [FAILED  Cat with a json response (from security-analytics).](#failed--cat-with-a-json-response-from-security-analytics)
   - [Writing Spec Tests](#writing-spec-tests)
     - [Simple Test Story](#simple-test-story)
+    - [Testing Multiple Verbs](#testing-multiple-verbs)
     - [Using Output from Previous Chapters](#using-output-from-previous-chapters)
     - [Managing Versions](#managing-versions)
     - [Managing Distributions](#managing-distributions)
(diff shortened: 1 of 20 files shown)


TESTING_GUIDE.md
```diff
@@ -182,6 +183,18 @@ chapters:
...
+Some APIs allow multiple verbs for the same effect. Specify multiple verbs as follows and the test tool will execute both.
+
+```yaml
+- synopsis: Use POST and PUT interchangeably.
+  path: /{index}
+  method:
+    - POST
+    - PUT
+```
+
```
nhtruong, line 197: This is a great shorthand for writing the same chapter multiple times with different methods. I think we should treat it as such: Multiple chapters that virtually identical but the method. And then when the story is read, we will turn that 1 chapter of `Create index "books"` into two chapter:
- `Create Index "book" [With POST]`
- `Create Index "book" [With PUT]`
Then you will only have to modify the story parser to turn that 1 chapter into 2 and the rest of the framework can stay intact. We can add a line in this guide saying that the chapter will be split into 2 so they know what to expect in the Result page.

Arguably, a case like this should be treat as 2 different stories because rarely are our non-PUT requests idempotent (even when you delete a resource on OS twice, you get 404 the second time). So you could not actually test this on `PUT/POST  /books` because executing `POST /books` after `POST /books` will result in `index named books already exists` error. This will open a can of worm of matrix argument when we have multiple of such chapters in an original story.

dblock: Okay let me see if I can do a parser-based implementation instead ...

dblock:
Closing in favor of #724

PR #732, opened 2024-12-13
Fixed /_search/scroll/{scroll_id}, missing search tests.

dblock:
### Description

- The `/_search/scroll/{scroll_id}` deprecated variant does not require `scroll_id` in the body because it's already in the path.
- Tests for `GET` and `POST /_search`.
- Tests for `GET /_search/pipeline` and `DELETE /_search/pipeline/{id}`.
- Tests for remainder of `_search/point_in_time`.

### Issues Resolved

Part of #663.

### DIFF
--- a/tests/default/_core/point_in_time/all.yaml
+++ b/tests/default/_core/point_in_time/all.yaml
@@ -1,0 +1,33 @@
+$schema: ../../../../json_schemas/test_story.schema.yaml
+
+description: Test point-in-time.
+version: '>= 2.4'
+epilogues:
+  - path: /_search/point_in_time/_all
+    method: DELETE
+  - path: /movies
+    method: DELETE
+    status: [200, 404]
+prologues:
+  - path: /_bulk
+    method: POST
+    parameters:
+      refresh: true
+    request:
+      content_type: application/x-ndjson
+      payload:
+        - {create: {_index: movies}}
+        - {title: The Cruise, year: 1998}
+        - {create: {_index: movies}}
+        - {title: Drive, year: 1960}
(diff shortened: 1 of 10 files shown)


tests/default/_core/search/pipeline.yaml
```diff
@@ -0,0 +1,33 @@
+$schema: ../../../../json_schemas/test_story.schema.yaml
+
+description: Test the creation of a search pipeline.
+version: '>= 2.8'
```
Tokesh, line 4: probably here it should be > 2.8 ?
https://github.com/opensearch-project/opensearch-api-specification/blob/ebe0f8a885f7db7e882d160c101055a5aa70a707/spec/namespaces/search_pipeline.yaml#L11

Tokesh (commented):
Looking perfect as usual. It looks like the OS version might need to be updated in one of the tests.

tests/default/_core/point_in_time/all.yaml
```diff
@@ -0,0 +1,33 @@
+$schema: ../../../../json_schemas/test_story.schema.yaml
+
+description: Test point-in-time.
+version: '>= 2.4'
+epilogues:
+  - path: /_search/point_in_time/_all
```
Tokesh, line 6: I have a little question. Is it okay to duplicate the use of the API
`DELETE /_search/point_in_time/_all?`
Does this ensure that we will definitely delete it?

dblock: This is just cleanup/safeguard in case the test fails.

Tokesh (approved):
Thank you, LGTM! Great job

PR #737, opened 2024-12-16
Support combining multiple output variables.

dblock:
### Description

- Support combining multiple output variables, e.g. `task_id: ${task.node}:${task.id}`. I ended up not using this and getting the task ID from `delete_by_query`, but it's a useful feature IMO.
 
### Issues Resolved

Part of #663.


dblock:
@nhtruong care to take a look pls?

tools/src/tester/types/eval.types.ts
```diff
@@ -87,16 +87,42 @@ export enum Result {
 export class OutputReference {
```
nhtruong, line 87: I missed this when this was first added but this file is meant to hold type definitions only. This class should be in its own file.

tools/src/tester/types/eval.types.ts
```diff
@@ -87,16 +87,42 @@ export enum Result {
...
+    // preserve type if 1 value is returned
+    let full_match = str.match(/^\$\{([^}]+)\}$/)
+    if (full_match) {
+      const spl = this.#split(full_match[1], '.', 2)
+      return callback(spl[0], spl[1])
+    } else return str.replace(/\$\{([^}]+)\}/g, (_, k) => {
+      const spl = this.#split(k, '.', 2)
+      return callback(spl[0], spl[1])
+    });
+  }
```
nhtruong, line 117: Should add a comment explaining what this function does because it's not clear what it's trying to achieve.

tools/src/tester/types/eval.types.ts
```diff
@@ -87,16 +87,42 @@ export enum Result {
...
+
+  static #split(str: any, delim: string, count: number): string[] {
+    if (str === undefined) return [str]
+    if (count <= 0) return [str]
+    const parts = str.split(delim)
+    if (parts.length <= count) return parts
+    const result = parts.slice(0, count - 1)
+    result.push(parts.slice(count - 1).join(delim))
+    return result
```
nhtruong, line 126: This needs an explanation as well.

tools/src/tester/StoryOutputs.ts
```diff
@@ -48,12 +48,13 @@ export class StoryOutputs {
   }
 
   resolve_string (str: string): any {
-    const ref = OutputReference.parse(str)
-    if (ref) {
-      return this.get_output_value(ref.chapter_id, ref.output_name)
-    } else {
-      return str
-    }
+    return OutputReference.replace(str, (chapter_id, output_name) => {
```
nhtruong, line 51: If this is the only place the `replace` function is used, then having a generic replace function that can handle any callback is a bit over-engineered, IMO.

tools/tests/types/eval.types.test.ts
```diff
@@ -0,0 +1,23 @@
...
+
+describe('OutputReference', () => {
+  let f = (id: any, k: any): string => `[${id}:${k}]`
+
+  describe('replace', () => {
+    it('replaces', () => {
+      expect(OutputReference.replace('string', f)).toEqual('string')
+      expect(OutputReference.replace('${k.v}', f)).toEqual('[k:v]')
+      expect(OutputReference.replace('${k.value}', f)).toEqual('[k:value]')
+      expect(OutputReference.replace('${k.v.m}', f)).toEqual('[k:v.m]')
```
nhtruong, line 20: I was on board until the last example. It's not clearly why `k` is treated differently from `v` and `m` and how the callback `f` cause such an output.

dblock:
Converted this to draft, will iterate. Extracted unrelated tests to https://github.com/opensearch-project/opensearch-api-specification/pull/746.

tools/src/tester/StoryOutputs.ts
dblock: I think it's best we keep the parsing regex values/logic in one place so it doesn't leak into the usage of it.

tools/src/tester/types/eval.types.ts
dblock: Moved to helpers.

dblock:
Ready to be reviewed again, thanks @nhtruong!

Issue #752, opened 2024-12-21
[BUG]Adjust Vale rules to ignore links


Naarcha-AWS:
### What is the bug?

Vale currently interprets words with in hyperlinks and creates errors based on the strings inside.  

### How can one reproduce the bug?

See the Style check in #751 

### What is the expected behavior?

Vale does not throw errors for strings in links.

dblock:
If anyone wants to write a little bit of code this one is easy: we have a piece of code that strips everything from the spec except words in https://github.com/opensearch-project/opensearch-api-specification/tree/main/tools/src/prepare-for-vale, it also needs to remove markdown links and only leave their text.

Tokesh:
Closed via #753

PR #753, opened 2024-12-22
Adjust Vale rules to ignore links

Tokesh:
### Description
Updated the `process_file` method to remove markdown links, leaving only their text content. This ensures the tool processes `description` and `x-deprecation-message` fields cleanly, aligning with Vale style-checking requirements.

### Issues Resolved
[#752]

### DIFF
--- a/tools/src/prepare-for-vale/KeepDescriptions.ts
+++ b/tools/src/prepare-for-vale/KeepDescriptions.ts
@@ -45,8 +45,8 @@ export default class KeepDescriptions {
       } else if (inside_text && line.match(/^[\s]*[\w\\$]*:/)) {
         inside_text = false
       } else if (inside_text) {
-        const cleanedLine = line.replace(/\[([^\]]+)\]\([^\)]+\)/g, '$1');
-        fs.writeSync(writer, this.prune_vars(cleanedLine));
+        const cleaned_line = line.replace(/\[([^\]]+)\]\([^\)]+\)/g, '$1');
+        fs.writeSync(writer, this.prune_vars(cleaned_line));
       }
       if (line.length > 0) {
         fs.writeSync(writer, "\n")
--- a/tools/src/prepare-for-vale/KeepDescriptions.ts
+++ b/tools/src/prepare-for-vale/KeepDescriptions.ts
@@ -41,9 +41,9 @@ export default class KeepDescriptions {
       if (line.match(/^[\s]+((description|x-deprecation-message): \|)/)) {
         inside_text = true
       } else if (line.match(/^[\s]+((description|x-deprecation-message):)[\s]+/)) {
-        var cleaned_line = this.prune(line, /(description|x-deprecation-message):/, ' ')
-        cleaned_line = this.remove_links(line)
+        let cleaned_line = this.prune(line, /(description|x-deprecation-message):/, ' ')
         cleaned_line = this.prune_vars(cleaned_line)
+        cleaned_line = this.remove_links(cleaned_line)
         fs.writeSync(writer, cleaned_line)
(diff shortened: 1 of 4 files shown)


dblock (changes requested):
1. Add tests, make sure there's one with 1 link and 2 links.
2. Extract `line.replace(/\[([^\]]+)\]\([^)]+\)/g, '$1')` into a method, similar to `prune_vars`, maybe something descriptive like `remove_links`.

tools/src/prepare-for-vale/KeepDescriptions.ts
```diff
@@ -45,7 +45,7 @@ export default class KeepDescriptions {
       } else if (inside_text && line.match(/^[\s]*[\w\\$]*:/)) {
         inside_text = false
       } else if (inside_text) {
-        const cleaned_line = line.replace(/\[([^\]]+)\]\([^)]+\)/g, '$1')
+        const cleaned_line = this.remove_links(line)
```
dblock, line 48: You can also have links in descriptions on line 44, so it will need to be handled there as well (and possibly in other cases). Add tests for those of course. 

We have `remove_links`, `prune_vars` and occasionally `prune`. Either extract the common parts into a method, or at least make it visible/more readable that we're doing a series of transformations.

```typescript
var cleaned_line = this.prune(line, /(description|x-deprecation-message):/, ' '))
cleaned_line = this.remove_links(line)
cleaned_line = this.prune_vars(cleaned_line)
fs.writeSync(writer, cleaned_line)
```

```typescript 
let cleaned_line = this.remove_links(line)
cleaned_line = this.prune_vars(cleaned_line)
fs.writeSync(writer, cleaned_line)
```

dblock (changes requested):
One more place where the transformation needs to be applied, see below.

Iterate till unit tests and linter are green (this repo uses a slightly different convention that the default TS with `variable_names`).

Add an integration test to https://github.com/opensearch-project/opensearch-api-specification/blob/main/tools/tests/prepare-for-vale/fixtures.

dblock (changes requested):
The way this works is that we remove all text that doesn’t need to be checked. 

So, you have to replace the text with an equal number of spaces otherwise the comments from the tool appear in the wrong place on GitHub.

tools/src/prepare-for-vale/KeepDescriptions.ts
```diff
@@ -62,4 +67,11 @@ export default class KeepDescriptions {
       return Array(match.length + 1).join(char)
     })
   }
+
+  remove_links(line: string): string {
+    return line.replace(/\[([^\]]+)\]\([^)]+\)/g, (match, p1) => {
+      const spaces = ' '.repeat(match.length - p1.length-1);
```
dblock, line 73:
```suggestion
      const spaces = ' '.repeat(match.length - p1.length - 1)
```

Suprised `npm run lint--fix` didn't pick on this. Remove `;` and space out `-`.

tools/src/prepare-for-vale/KeepDescriptions.ts
```diff
@@ -62,4 +67,11 @@ export default class KeepDescriptions {
...
+      return ' ' + p1 + spaces;
```
dblock, line 74:
```suggestion
      return ' ' + p1 + spaces
```

tools/src/prepare-for-vale/KeepDescriptions.ts
```diff
@@ -62,4 +67,11 @@ export default class KeepDescriptions {
...
+    });
```
dblock, line 75:
```suggestion
    })
```

dblock (changes requested):
Sorry to be annoying.

dblock (approved):

PR #754, opened 2024-12-23
Fix cluster schemas

Xtansia:
### Description
Fix cluster schemas


dblock (changes requested):
Looks good, fix vale complaints pls?

Xtansia:
> Looks good, fix vale complaints pls? See the diff. We should enable that to fail CI as well.

Have fixed the vale complaints

PR #760, opened 2025-01-02
Add ML Model Group APIs

nathaliellenaa:
### Description
Added missing ML Model Group APIs.

### Issues Resolved
Part of https://github.com/opensearch-project/opensearch-py/issues/867.

### ML Model Group APIs to add
- [x] PUT /_plugins/_ml/model_groups/{model_group_id}
- [x] GET /_plugins/_ml/model_groups/_search
- [x] POST /_plugins/_ml/model_groups/_search

### DIFF
--- a/tests/plugins/ml/ml/model_groups/search.yaml
+++ b/tests/plugins/ml/ml/model_groups/search.yaml
@@ -1,0 +1,52 @@
+$schema: ../../../../../json_schemas/test_story.schema.yaml
+
+description: Test the search of model groups.
+version: '>= 2.11'
+prologues:
+  - path: /_plugins/_ml/model_groups/_register
+    id: create_model_group
+    method: POST
+    request:
+      payload:
+        name: NLP_Group
+        description: Model group for NLP models.
+    output:
+      test_model_group_id: payload.model_group_id
+epilogues:
+  - path: /_plugins/_ml/model_groups/{model_group_id}
+    method: DELETE
+    status: [200, 404]
+    parameters:
+      model_group_id: ${create_model_group.test_model_group_id}
+chapters:
+  - synopsis: Search model group.
+    path: /_plugins/_ml/model_groups/_search
+    method: GET
+    request:
+      payload:
+        query:
+          match_all: {}
+        size: 1000
+    response:
+      status: 200
+      payload:
+        hits:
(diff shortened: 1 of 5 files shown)


tests/plugins/ml/ml/model_groups/search.yaml
```diff
@@ -0,0 +1,52 @@
+$schema: ../../../../../json_schemas/test_story.schema.yaml
+
+description: Test the search of model groups.
+version: '>= 2.11'
```
Tokesh, line 4: You mentioned that the tests should run on this version, but the spec itself does not include x-version-added

tests/plugins/ml/ml/model_groups/update.yaml
```diff
@@ -0,0 +1,36 @@
+$schema: ../../../../../json_schemas/test_story.schema.yaml
+
+description: Test the creation of model groups.
+version: '>= 2.11'
```
Tokesh, line 4: and here also

Tokesh (changes requested):
Great job!
Could you please check not only the specs from this PR but also the ML specs you added? We likely missed this, and it could cause issues in the future if someone tries to use these APIs from earlier versions.

nathaliellenaa:
> Great job! Could you please check not only the specs from this PR but also the ML specs you added? We likely missed this, and it could cause issues in the future if someone tries to use these APIs from earlier versions.

Got it, I'll check about this. Since the PR for the train and predict API has been merged, should I create a new PR to update the version or should I make the changes in this PR?

Tokesh:
> should I create a new PR to update the version or should I make the changes in this PR?

It depends on you; whichever is more comfortable for you :)

nathaliellenaa:
> It depends on you; whichever is more comfortable for you :)

I'll create a new PR to avoid confusion, thank you!

Tokesh (approved):

Tokesh:
LGTM, Thank you very much! :)
But just in case, let someone else check, I don't have much experience with the ML plugin

dblock:
> But just in case, let someone else check, I don't have much experience with the ML plugin

I think @nathaliellenaa has more experience with the ML plugin by now than all of us :) Merging.

PR #759, opened 2025-01-02
Merge coverage CI showing right percentage

Tokesh:
### Description
Ignore x-ignorable specs in evaluated operations

### Issues Resolved
[#748]

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

Issue #748, opened 2024-12-18
[BUG] Merge coverage CI showing 100%+


Tokesh:
### What is the bug?
We have a CI in our GitHub Actions that was supposed to calculate how many specs were covered by tests, but it shows over 100%.
https://github.com/opensearch-project/opensearch-api-specification/actions/runs/12397662650/job/34608665915?pr=745
`
{"total_operations_count":536,"evaluated_operations_count":537,"evaluated_paths_pct":100.19}
`
### How can one reproduce the bug?
Create a PR with a passing CI

### What is the expected behavior?
Not more than 100%

### What is your host/environment?
GHA

### Do you have any screenshots?
![image](https://github.com/user-attachments/assets/348dace0-7239-421d-b8af-a1350fd43de7)

Tokesh:
Closed via [#759, #762]

PR #762, opened 2025-01-03
Added spec with `x-ignorable:true` to check Evaluation operations

Tokesh:
### Description
I'm adding a spec with the parameter `x-ignorable:true` to see if it will appear in `operations` and `evaluated_operations` on the existing tests.
debt from #759 

### Issues Resolved
[#748]

### DIFF
--- a/tools/tests/tester/MergedOpenApiSpec.test.ts
+++ b/tools/tests/tester/MergedOpenApiSpec.test.ts
@@ -21,6 +21,7 @@ describe('merged API spec', () => {
 
     test('paths', () => {
       expect(spec.paths()).toEqual({
+        '/_cluster/nodes/hot_threads': ['get'],
         '/_nodes/{id}': ['get', 'post'],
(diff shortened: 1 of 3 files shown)


dblock (changes requested):
If I merge this test without your fix in https://github.com/opensearch-project/opensearch-api-specification/pull/759 no tests will fail. So add a JavaScroipt/code test specifically for the "it will appear in operations and evaluated_operations on the existing tests" part.

dblock:
For a unit test you want something a lot lower level and simpler and intentional. The method you changed was `TestResults.evaluatied_operations()`. So modify/add to https://github.com/opensearch-project/opensearch-api-specification/blob/main/tools/tests/tester/TestResults.test.ts a test that looks something like

```typescript
descrive('evaluatied_operations', () => {
  test('does not include x-ignorable APIs', () => {
    expect(test_results.evaluated_operations()).to ...
  })
})
```

dblock (approved):

PR #785, opened 2025-01-13
Fix field name from getTime to time in GetStats schema

hye-on:
### Description
Fixed response schema in GetStats to rename `getTime` to `time` field to maintain consistency across the API schema.

### Issues Resolved
Related to opensearch-project/OpenSearch#16894


dblock:
Thanks! Given that this is a breaking change the way to introduce it is to add `time` in OpenSearch and deprecate `getTime`, and do the same here in the spec after that PR is merged into OpenSearch core.

Issue #707, opened 2024-12-05
[FEATURE] Spec Conversion to Protobufs


amberzsy:
given [[PROPOSAL] Spec to Protobuf Conversion Rules and Tooling](https://github.com/opensearch-project/opensearch-api-specification/issues/677), task breakdown for protobuf conversion
- [ ] Publish initial protobufs 
  - [ ] Search Protos (PR: https://github.com/opensearch-project/opensearch-api-specification/pull/676)
  - [ ] Add basic CI for proto validation
  - [ ] Document Protos 
  - [ ] Index Protos

PR #789, opened 2025-01-14
Add geneate proto workflow

lucy66hw:
### Description
Create a manual workflow to convert OpenSearch specifications into Protobuf files for testing tool development.

Workflow Steps:
- Use merger to combine all specifications into a single YAML file.
- Clone the development fork repository for Protobuf generation.
- Extract and preprocess the target specification paths for conversion.
- Build the JAR file using the forked repository.
- Generate Protobuf files locally based on the specifications.

### Issues Resolved
[FEATURE] Spec Conversion to Protobufs #707


dblock:
Let's talk about tooling a little, I think we may be complicating things. At a high level, I think 75% of this Python code can be removed in favor of much simpler GitHub actions. All it does is drive command line java tools. For the rest (preprocessors), the entire tooling in this repo is TypeScript. Do we really want to introduce tools that are Python? We already have a lot of well tested code that processes and transforms spec files, so I am a little worried that this will become a Frankenstein repo that is very difficult to work in.

lucy66hw:
Thanks for the comment. This is a temporary workflow for protobuf generation. In the next phase, we plan to integrate wf into GitHub actions to automate the process alongside CI and implement preprocessing with typeScript.

dblock (changes requested):
Will wait for the next iteration.

Issue #617, opened 2024-10-16
[FEATURE] add linter for numeric type


amberzsy:
### Is your feature request related to a problem?
there are bunch of field with inaccurate/incorrect numeric type. as mentioned [here](https://github.com/opensearch-project/opensearch-api-specification/issues/596)

### What solution would you like?
we should add linter for this to enforce and stop the gaps. 

### What alternatives have you considered?
N/A

PR #800, opened 2025-01-17
Added a linter for numeric types

Tokesh:
### Description
Checking numeric types

### Issues Resolved
[#617]


dblock:
> I can't understand why the tests run without errors on two local computers, but fail on CI. image

It's certainly not expected! If I were brave and wanted to get to the bottom of this, I would edit the code to return the full error, then run it locally and in CI to find what the actual root cause is with no other changes.

I also ran into it in https://github.com/opensearch-project/opensearch-api-specification/pull/793 (not yet merged) and updated the test for now. I am guessing a library version must be different or the list of errors is sorted differently depending on some transient environment / operating system / node version. The JSON schema validator fails correctly, but the error messages returned vary in order, and I believe we truncate them, so you get a different output in CI vs. locally.

Tokesh:
> It's certainly not expected! If I were brave and wanted to get to the bottom of this, I would edit the code to return the full error, then run it locally and in CI to find what the actual root cause is with no other changes.
> 
> I also ran into it in #793 (not yet merged) and updated the test for now. I am guessing a library version must be different or the list of errors is sorted differently depending on some transient environment / operating system / node version. The JSON schema validator fails correctly, but the error messages returned vary in order, and I believe we truncate them, so you get a different output in CI vs. locally.

Thank you so much! I actually don’t have any idea why it’s happening. It gives me ideas on how to work with it.

dblock:
> Thank you so much! I actually don’t have any idea why it’s happening. It gives me ideas on how to work with it.

I am happy to look into it next week if you can't figure it out.

Xtansia:
@Tokesh I'd recommend syncing the latest main branch into your fork and rebasing your branch onto it. CI builds usually try to run as if your branch has been merged and so can sometimes have different results compared to your local clone.

Tokesh:
Finally, thank you so much for help and suggestions! @dblock @Xtansia 
34 commits and a lot of debugging...
I think i need to write motivation post here for newbies:
"I fear not the man who has practiced 10,000 kicks once, but I fear the man who has practiced one kick 10,000 times" (c) Bruce Lee

spec/schemas/transforms._common.yaml
```diff
@@ -208,23 +215,30 @@ components:
...
+          format: double
         documents_indexed:
           type: number
+          format: double
         index_time_in_millis:
           type: number
+          format: double
         search_time_in_millis:
           type: number
+          format: double
```
Xtansia, line 241: Blindly adding `format: double` to any `type: number` without a format is not ideal as for example none of these should actually be doubles.

Xtansia (changes requested):

spec/schemas/transforms._common.yaml
Tokesh: Honestly, I initially wanted to fix each one individually, but there were around 500 errors in the Validate CI. One idea was to create an issue and fix them gradually over time, or, as a temporary solution, set everything to `double` or `int32` by default for now.

I'm open to any suggestions from you. In any case, I made this change just to ensure that the CI passes successfully.

dblock: @Xtansia @nhtruong what's the default handling of `format` in generators when we don't have it? We should safely default to that here, no?

nhtruong: > One idea was to create an issue and fix them gradually over time, or, as a temporary solution, set everything to `double` or `int32` by default for now.

Setting everything to double would make it hard to go back and figure out which one has been verified to be an actual double, and which should be corrected to float. If you just leave them blank, it's easier to tell which ones need to be looked into.

The task of determining the format of every numerical schema is going to take a lot of time and effort. Is there a need for that right now?

tools/src/linter/SchemaVisitingValidator.ts
```diff
@@ -73,27 +72,23 @@ export default class SchemaVisitingValidator {
...
+        errors.push(ctx.error(`Schema of type 'number' must specify a valid format. Allowed formats: ${SCHEMA_NUMBER_FORMATS.join(', ')}`));
+        return;
       }
-
-      if (SCHEMA_INTEGER_FORMATS.includes(schema.format)) {
-        errors.push(ctx.error(`schema of type 'number' with format '${schema.format}' should instead be of type 'integer'`))
-      } else {
-        errors.push(ctx.error(`schema of type 'number' with format '${schema.format}' is invalid, expected one of: ${SCHEMA_NUMBER_FORMATS.join(', ')}`))
+      if (!SCHEMA_NUMBER_FORMATS.includes(schema.format)) {
+        errors.push(ctx.error(`Schema of type 'number' with format '${schema.format}' is invalid. Expected one of: ${SCHEMA_NUMBER_FORMATS.join(', ')}`));
```
nhtruong, line 80: If we're only allowing non-null valid format then the first `if` on lines 74-77 are redundant because `SCHEMA_NUMBER_FORMATS` doesn't include `null` or `false`, and the error in the first `if` will be triggered in the second `if` anyway.

spec/schemas/transforms._common.yaml
nhtruong: This might not be the ask from the original issue. Waiting on verification.

Xtansia: > @Xtansia @nhtruong what's the default handling of `format` in generators when we don't have it? We should safely default to that here, no?

Java code-gen maps as following:
| OAS `type` | OAS `format` | Java Type |
| - | - | - |
| `number` | `undefined` | `Number` |
| `number` | `float` | `float`/`Float` |
| `number` | `double` | `double`/`Double` |
| `integer` | `undefined` | `int`/`Integer` |
| `integer` | `int32` | `int`/`Integer` |
| `integer` | `int64` | `long`/`Long` |

Which roughly matches onto how [OAS/Swagger defines the types/formats](https://swagger.io/docs/specification/v3_0/data-models/data-types/#numbers):
Screenshot 2025-02-03 at 11 56 38 AM

Tokesh: @dblock @nhtruong @Xtansia 
I think I now understand the entire context, thanks to everyone for participating in my PR discussion. I suggest that I create an Issue where, sooner or later, we will try to specify the exact format for each number and integer. After we ensure that we've done this everywhere, we can merge this PR. (I will leave just the linter check here.) I think we, as maintainers, can at least start asking people to specify more accurate formats in future PRs so that the number of such inaccurate parameters doesn’t increase. At the very least, I’ll try to gradually fill in the formats of the specs during my free time.
WDYT?

dblock: I think this is a good plan. Thanks for working on this.

nhtruong: It's might be a tall order but we can add a new a linter that flags missing `format` for new number/integer schemas.

dblock: This is that PR, but it forces that requirement on all fields.

@Tokesh maybe this PR can emit a warning instead of an error?

nhtruong: I mean only for new schemas, which makes it hard because the linter will have to be able to detect which ones are new from the git diff.

PR #808, opened 2025-01-25
Fixed swapped schema references in nodes info API buffer fields

hye-on:
### Description
Fixed incorrectly referenced schemas in nodes info API buffer fields by swapping the schema types between `total_indexing_buffer` and `total_indexing_buffer_in_bytes` .

### Issues Resolved
Related to https://github.com/opensearch-project/OpenSearch/issues/16910

### Related PR
https://github.com/opensearch-project/OpenSearch/pull/17070

### Comments
This change will be effective from version 3.0.0 onwards


dblock (approved):
@Xtansia I read the related issue, this is correct right?

dblock (changes requested):
Actually we now have a failing test. The spec needs to work correctly for all versions, how do we fix this?

hye-on:
> Actually we now have a failing test. The spec needs to work correctly for all versions, how do we fix this?

@dblock I noticed that PR https://github.com/opensearch-project/OpenSearch/pull/17070 will implement changes from OpenSearch 3.0.0. Can the api-specification repository define different specifications per version? Or should I wait until OpenSearch 3.0.0 is released to submit this PR?

I've reviewed the failing test files and related files (tests/default/nodes/nodes.yaml, tests/default/nodes/usage.yaml, json_schemas/test_story.schema.yaml, and others) but couldn't identify where to make changes. (Does the test payload response come from OpenSearch itself?) If I've missed any files that should be reviewed, I would greatly appreciate your guidance.

Thank you for taking the time to review this and for your help! :)

dblock:
@hye-on In general we would do this:

1. Annotate the current schema for both fields with `x-version-removed: 3.0`.
2. Add the new schema with `x-version-added: 3.0`.

See https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md#openapi-extensions for field details. 

Since the tools don't allow the same field name twice, I think we need to do something like this:

```yaml
        total_indexing_buffer_in_bytes:
          oneOf:
            - $ref: '_common.yaml#/components/schemas/HumanReadableByteCount'
              x-version-removed: 3.0
            - $ref: '_common.yaml#/components/schemas/ByteCount'
              x-version-added: 3.0
```

You can run the merge (see https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md#spec-merger) with an option to target 3.0 and it should remove the `oneOf` that's not matching.

If that doesn't work, then let's do the above minus the `x-version` options with a comment.

hye-on:
@dblock 
Thank you for your kind guidance.
When I added the `x-version-removed` and `x-version-added` tags, the 3.0 version test failed.
image

As you guided, after removing the tags, the 3.0 test is now working.
Thank you.

I notice that the Check Links is failing (502 Network error: Bad Gateway).
If there's anything I need to fix, please let me know.

Thank you again for taking the time to help me ! :)

dblock:
@hye-on you have to update the SHA to OpenSearch 3.0 to a newer one with https://github.com/opensearch-project/OpenSearch/pull/17070 merged in https://github.com/opensearch-project/opensearch-api-specification/blob/main/.github/workflows/test-spec.yml#L65. 

It's not always trivial to find a complete distribution of OpenSearch 3.0 that has all the plugins, so other tests may fail, but worth a try on 1-2 newer revisions available in DockerHub. Otherwise add the `x-version-` headers and we can keep this open until we find one.

PR #825, opened 2025-03-01
Remove empty spaces in links when producing text masks for vale.

nhtruong:
The extra space characters are causing Vale to [fail](https://github.com/opensearch-project/opensearch-api-specification/actions/runs/13550673566/job/37873283914?pr=824). This change makes the text masks for links appear like how the links are displayed.


nhtruong:
@dblock I know the original goal was to produce the text masks where every displayed character would be at the same position as in the original file. But that's now how the end-user would see it. And since Vale is meant to lint the end-user text, I think it makes more sense to remove those spaces.

dblock (approved):

PR #826, opened 2025-03-04
Remove unnecessary allOf 

lucy66hw:
### Description
From [OpenAPI 3.1.0](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md), allOf subschema should be valid object rather than primitive type. 
`allOf takes an array of object definitions that are validated independently but together compose a single object.
`
### Issues Resolved

### DIFF
--- a/spec/schemas/_common.yaml
+++ b/spec/schemas/_common.yaml
@@ -186,8 +186,7 @@ components:
         - x
         - y
     EpochTimeUnitMillis:
-      allOf:
-        - $ref: '#/components/schemas/UnitMillis'
+      $ref: '#/components/schemas/UnitMillis'
     UnitMillis:
       description: The time unit for milliseconds.
       type: integer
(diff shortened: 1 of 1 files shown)


Xtansia:
I would note that the actual JSON Schema specification does not place any restriction on the types of the sub-schemas:
> [10.2.1.1](https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-00#section-10.2.1.1).  allOf
>
>   This keyword's value MUST be a non-empty array.  Each item of the
>   array MUST be a valid JSON Schema.
>
>   An instance validates successfully against this keyword if it
>   validates successfully against all schemas defined by this keyword's
>   value.

lucy66hw:
The OpenAPI specification mentioned that the allOf subschema should be an object. the reason we are modifying this because, in our spec-to-Proto conversio, using allOf with a simple type causes inaccurate conversions. Is there any specific reason to keep allOf in current form?

nhtruong:
@lucy66hw this change makes sense. I think what @Xtansia was saying is that the way they are right now are still valid schemas (Our client generators have no issues with it). So it's not a blocker. Just a footnote.

PR #842, opened 2025-03-14
Fixed REST status codes for RBAC and provisioning

owaiskazi19:
### Description
Depends on https://github.com/opensearch-project/flow-framework/pull/1083
- Added support for an array of status codes
- Fixed REST status codes for RBAC and provisioning

### Issues Resolved

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -118,6 +118,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Fixed Aggregation schemas ([#801](https://github.com/opensearch-project/opensearch-api-specification/pull/801))
 - Fixed FilterQueryRequestProcessor to use correct query type ([#821](https://github.com/opensearch-project/opensearch-api-specification/pull/821))
 - Fixed `knn.train_model`'s request body `method` field to accept an object rather than a string ([#814](https://github.com/opensearch-project/opensearch-api-specification/pull/814))
+- Fixed REST status codes for RBAC and provisioning for Flow Framework plugin
 
 ### Changed
 - Changed `tasks._common:TaskInfo` and `tasks._common:TaskGroup` to be composed of a `tasks._common:TaskInfoBase` ([#683](https://github.com/opensearch-project/opensearch-api-specification/pull/683))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -118,7 +118,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Fixed Aggregation schemas ([#801](https://github.com/opensearch-project/opensearch-api-specification/pull/801))
 - Fixed FilterQueryRequestProcessor to use correct query type ([#821](https://github.com/opensearch-project/opensearch-api-specification/pull/821))
 - Fixed `knn.train_model`'s request body `method` field to accept an object rather than a string ([#814](https://github.com/opensearch-project/opensearch-api-specification/pull/814))
-- Fixed REST status codes for RBAC and provisioning for Flow Framework plugin
+- Fixed REST status codes for RBAC and provisioning for Flow Framework plugin ([#842](https://github.com/opensearch-project/opensearch-api-specification/pull/842))
 
 ### Changed
 - Changed `tasks._common:TaskInfo` and `tasks._common:TaskGroup` to be composed of a `tasks._common:TaskInfoBase` ([#683](https://github.com/opensearch-project/opensearch-api-specification/pull/683))
--- a/json_schemas/test_story.schema.yaml
+++ b/json_schemas/test_story.schema.yaml
@@ -192,12 +192,12 @@ definitions:
       status:
         oneOf:
           - type: integer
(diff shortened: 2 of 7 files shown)


CHANGELOG.md
```diff
@@ -118,6 +118,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Fixed Aggregation schemas ([#801](https://github.com/opensearch-project/opensearch-api-specification/pull/801))
 - Fixed FilterQueryRequestProcessor to use correct query type ([#821](https://github.com/opensearch-project/opensearch-api-specification/pull/821))
 - Fixed `knn.train_model`'s request body `method` field to accept an object rather than a string ([#814](https://github.com/opensearch-project/opensearch-api-specification/pull/814))
+- Fixed REST status codes for RBAC and provisioning"
```
dbwiddis, line 121: Probably want to mention the plugin here.

tests/default/flow_framework/workflow/provision.yaml
```diff
@@ -32,4 +32,4 @@ chapters:
       payload:
         openai_key: '1234556'
     response:
-      status: 200
\ No newline at end of file
+      status: 202
```
dbwiddis, line 35: Line 3 speceifies version >=2.12.  This is a change for 3.0+.

I'm not sure how we are supposed to specify a version change.  Do we make line 3 `>=2.12 && <3` and have a new file for `>=3.0.0`?  

@Xtansia advice here? Developer guide doesn't cover major version API changes.

owaiskazi19: @nhtruong don't we need to have `3.0.0-alpha-1` rather `3.0` here? That's why it's failing for 3.0

nhtruong: It's getting much more complicated to host 2 different major versions of the spec (2.x and 3.x) and test them now. Having a branch for each major version (with `main` hosting the latest unreleased major version) will solve this issue.  What's your thought on this?
cc: @Xtansia @dblock

dblock: I would avoid duplicating/branching the spec because it means backporting hell like we have in so many repos. You should be able to use `x-version-added` and friends to specify that in 2.x the API returns 200 and in 3.x it returns 202. You could also say that the API returns either of 200/202 with a comment, which I think is ok for now.

owaiskazi19: Thanks @dblock! Pushed the change

owaiskazi19: If we change the type to array instead of integer for expected response [here](https://github.com/opensearch-project/opensearch-api-specification/blob/main/json_schemas/test_story.schema.yaml#L193), that would mean changing all the responses in the test files to return an array rather an int?

dblock: You can have `anyOf` in the schema between an integer and an array of integers.

dblock (changes requested):
You'll need to implement support for an array of status codes to get this in. Give it a shot!

json_schemas/test_story.schema.yaml
```diff
@@ -190,15 +190,23 @@ definitions:
     type: object
     properties:
       status:
-        type: integer
-        description: The expected HTTP status code. Default to 200.
-        default: 200
+        anyOf:
```
dblock, line 193: It's one or the other, and only one, so I think `oneOf` is more appropriate.

dblock (commented):
I'm pretty sure there's some code to write to check status, it's not just schema. Plus tests.

json_schemas/test_story.schema.yaml
owaiskazi19: The return doesn't matches with the elements of array. it has content type too but that we have already passed in the schema.
```
ERROR   flow_framework/workflow/provision.yaml (tests/default/flow_framework/workflow/provision.yaml)
    ERROR   CHAPTERS
        ERROR   Provision workflow.
            PASSED  PARAMETERS
                PASSED  workflow_id
            PASSED  REQUEST BODY
            ERROR   RESPONSE STATUS (Expected status 200,202, but received 200: application/json.)
            SKIPPED RESPONSE PAYLOAD BODY

nhtruong (approved):

json_schemas/test_story.schema.yaml
owaiskazi19: Had to add logic to handle array of status code for chapters. Thanks @nhtruong for the pointer here. Tests passed. @dblock @dbwiddis @nhtruong should be good to review now. Thanks

nhtruong:
Thanks @owaiskazi19  LGTM
@dblock feel free to merge if there's no issue on your end.

dblock (changes requested):
There needs to be a test for the new `ChapterEvaluator` functionality. Something in https://github.com/opensearch-project/opensearch-api-specification/tree/main/tools/tests/tester.

dblock (approved):
Would be a good idea to add a test for when the status code is an array and doesn't match.

PR #843, opened 2025-03-16
Fix Bulk API schemas

karenyrx:
### Description
While working on [GRPC Bulk endpoint](https://github.com/opensearch-project/OpenSearch/pull/17727), discovered a few discrepancies in the Bulk protos with the source code

1. WaitForActiveShardOptions does not support `index-settings`. It is `all`, `null`, or an integer greater than 0 according to [ActiveShardCount#parseString](https://github.com/opensearch-project/OpenSearch/blob/1937f5f06271cd70522c4d7391140b01070a16db/server/src/main/java/org/opensearch/action/support/ActiveShardCount.java#L117-L138).
2. [routing](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/action/bulk/BulkRequest.java#L403) field is a String, not array of String
3. [CreateOperation](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/action/index/IndexRequest.java#L201-L222) does not support version_type, and versioning-related fields. makes sense as CreateOperation is for creating a document for the very first time.
4. [UpdateOperation](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/action/update/UpdateRequest.java#L544-L551) does not support version_types and will always use VersionType.INTERNAL
5. `dynamic_templates` is not supported for IndexOperation and CreateOperation 
6. Spec is missing [ShardInfo](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/action/support/replication/ReplicationResponse.java#L96) type to be used in the BulkResponse. The existing ShardStatistics struct is only used for Search responses. 
7. Missing [metaFields](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/get/GetResult.java#L300-L307) in InlineGetDictUserDefined field. 
8. ErrorCause should have an additional field called  [header](https://github.com/opensearch-project/OpenSearch/blob/1937f5f06271cd70522c4d7391140b01070a16db/libs/core/src/main/java/org/opensearch/OpenSearchException.java#L421), which is a `Map<String, List<String> | String>`  
9. ScriptLanguage should be a oneOf not anyOf.

### Issues Resolved
General spec fixes, contributing to accuracy of the Bulk Protobufs:
https://github.com/opensearch-project/OpenSearch/issues/16784
https://github.com/opensearch-project/opensearch-protobufs/issues/28

### DIFF
--- a/spec/namespaces/_core.yaml
+++ b/spec/namespaces/_core.yaml
@@ -3412,7 +3412,7 @@ components:
       name: routing
       description: Custom value used to route operations to a specific shard.
       schema:
-        $ref: '../schemas/_common.yaml#/components/schemas/RoutingInQueryString'
+        $ref: '../schemas/_common.yaml#/components/schemas/Routing'
       style: form
     bulk::query.timeout:
       in: query
--- a/spec/schemas/_common.yaml
+++ b/spec/schemas/_common.yaml
@@ -969,6 +969,9 @@ components:
           type: object
           description: The source of the document.
           additionalProperties: true
+      additionalProperties: 
+        type: object 
+        title: metadata_fields 
       required:
         - found
     Names:
--- a/spec/schemas/_common.yaml
+++ b/spec/schemas/_common.yaml
@@ -938,10 +938,13 @@ components:
         - type: string
           const: all
           description: Wait for all shards to be active.
+        - type: string
+          not: 
+            type: string
+            const: all
+          description: A numeric value greater than or equal to 0
         - type: 'null'
           description: null represents the default value (which defaults to one shard copy).
-        - $ref: '#/components/schemas/uint'
-          description: A numeric value greater than or equal to 0
     InlineGetDictUserDefined:
       type: object
       description: The inline get operation results for user-defined dictionaries.
--- a/spec/schemas/_common.yaml
+++ b/spec/schemas/_common.yaml
@@ -549,18 +549,9 @@ components:
               required:
                 - source
     ScriptLanguage:
-      oneOf:
-        - title: builtin
-          $ref: '#/components/schemas/BuiltinScriptLanguage'
-        - title: custom
-          type: string
-    BuiltinScriptLanguage:
       type: string
-      enum:
-        - expression
-        - java
-        - mustache
-        - painless
+      description: |-
+        Built in scripts include: 'expression', 'java', 'mustache', and 'painless', but custom script languages can be provided as well.
     ScriptBase:
       type: object
       properties:
--- a/spec/schemas/_common.yaml
+++ b/spec/schemas/_common.yaml
@@ -362,6 +362,25 @@ components:
         - failed
         - successful
         - total
+    ShardInfo:
+      type: object
+      properties:
+        failed:
+          $ref: '#/components/schemas/uint'
+        successful:
+          $ref: '#/components/schemas/uint'
+        total:
+          $ref: '#/components/schemas/uint'
+        failures:
+          type: array
+          items:
+            $ref: '#/components/schemas/ShardFailure'
+        primary:
+          $ref: '#/components/schemas/uint'
+      required:
+        - failed
+        - successful
+        - total
     uint:
       type: integer
     ShardFailure:
@@ -402,6 +421,10 @@ components:
           type: array
           items:
             $ref: '#/components/schemas/ErrorCause'
+        headers:
+          type: object
+          additionalProperties:
+            $ref: '#/components/schemas/StringOrStringArray'
       required:
         - type
       additionalProperties:
@@ -526,7 +549,7 @@ components:
               required:
                 - source
     ScriptLanguage:
-      anyOf:
+      oneOf:
         - title: builtin
           $ref: '#/components/schemas/BuiltinScriptLanguage'
         - title: custom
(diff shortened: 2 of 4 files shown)


dblock:
Looks like a good start. Sign DCO (`git commit -s --amend`), get it to green. Thanks!

karenyrx:
Rebased to address merge conflicts. @dblock @nhtruong could you help take another look? Thanks!

dblock:
Looks fine, but I'd expect to see some tests that fail without these changes?

spec/schemas/_common.yaml
```diff
@@ -526,18 +549,9 @@ components:
...
-          type: string
-    BuiltinScriptLanguage:
       type: string
-      enum:
-        - expression
-        - java
-        - mustache
-        - painless
+      description: |-
+        Built in scripts include: 'expression', 'java', 'mustache', and 'painless', but custom script languages can be provided as well.
```
Xtansia, line 554: I'm not sure I agree with removing information from the specification. I feel like if you want these to be collapsed/simplified, that could be handled at a tooling/translation level?

Additionally these have generated counterparts that are used in opensearch-java:
- https://github.com/opensearch-project/opensearch-java/blob/main/java-client/src/generated/java/org/opensearch/client/opensearch/_types/BuiltinScriptLanguage.java
- https://github.com/opensearch-project/opensearch-java/blob/main/java-client/src/generated/java/org/opensearch/client/opensearch/_types/ScriptLanguage.java

spec/schemas/_common.yaml
```diff
@@ -362,6 +362,25 @@ components:
...
+        successful:
+          $ref: '#/components/schemas/uint'
+        total:
+          $ref: '#/components/schemas/uint'
+        failures:
+          type: array
+          items:
+            $ref: '#/components/schemas/ShardFailure'
+        primary:
+          $ref: '#/components/schemas/uint'
```
Xtansia, line 379: I'm not able to find a variant of shard stats/info that returns a `primary` count (at least from a quick look at the `main` branch), are you able to point to where this may be returned? Does this actually need to be a separate type?

spec/namespaces/_core.yaml
```diff
@@ -3412,7 +3412,7 @@ components:
       name: routing
       description: Custom value used to route operations to a specific shard.
       schema:
-        $ref: '../schemas/_common.yaml#/components/schemas/RoutingInQueryString'
+        $ref: '../schemas/_common.yaml#/components/schemas/Routing'
```
Xtansia, line 3415: What's the reason for changing this? The `routing` query param may be a comma delimited array?

karenyrx: hm, per the java code, it seems [routing](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/action/bulk/BulkRequest.java#L403) field is a String, not array of String

also confirmed by the public doc https://docs.opensearch.org/docs/latest/api-reference/document-apis/bulk/

spec/schemas/_common.yaml
karenyrx: yes, per #6 in the diff description, [ShardInfo](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/action/support/replication/ReplicationResponse.java#L96) is used in the BulkResponse. The existing ShardStatistics struct is only used for Search responses.

spec/schemas/_common.yaml
Xtansia: I'm not sure why `anyOf` represents a problem or inaccuracy? It is true that any known/built-in script language is also a valid "custom" language, but not all "custom" languages will be "known" languages. 

If you really wanted this to be a `oneOf`, you could achieve it via:
```yaml
    ScriptLanguage:
      oneOf:
        - title: builtin
          $ref: '#/components/schemas/BuiltinScriptLanguage'
        - title: custom
          type: string
          not:
            $ref: '#/components/schemas/BuiltinScriptLanguage'
```

The primary goal of the specification at present is enabling the generation of the language clients, not necessarily to match the server's internal representation. So being able to provide an enum of core/commonly used values is helpful to end users.

spec/schemas/_common.yaml
karenyrx: Ah, yes misread your comment.  Yes `primary`  doesnt seem to exist. Removed, thanks for catching it

karenyrx:
> Looks fine, but I'd expect to see some tests that fail without these changes?

For all the fields that were removed -- such as: `index-settings` from `WaitForActiveShardOptions`, `version_type` from `CreateOperation`, `version_type` for `UpdateOperation`, `dynamic_templates` for `IndexOperation` and `CreateOperation`, a string array of `routing`, and `skipped` in  `ShardStatistics` --  I don't think there existed tests that used these properties in the first place, so removing them did not cause any tests to fail. For these, I am unsure if there is a way to retroactively add tests to check the response does _not_ include certain fields, but let me know if that is untrue and I can try adding some tests. 

For the fields that were newly added, such as `metadata_fields` for `InlineGetDictUserDefined`, and `header` in `ErrorCause`, there definitely could exist new tests for them, but since these are on the response-side, I have not had the bandwidth to find a way to trigger these responses from OpenSearch yet. If it is not a blocker for this PR, perhaps we can merge these spec fixes in first and I can create an issue to backfill those tests later. 

For all the fields that were modified, such as `ScriptLanguage`, although tests existed, since I am just tightening the schema definition not changing it, these tests continued to pass as well. 

@dblock Let me know what you think! Thanks

dblock:
> For the fields that were newly added, such as `metadata_fields` for `InlineGetDictUserDefined`, and `header` in `ErrorCause`, there definitely could exist new tests for them, but since these are on the response-side, I have not had the bandwidth to find a way to trigger these responses from OpenSearch yet. If it is not a blocker for this PR, perhaps we can merge these spec fixes in first and I can create an issue to backfill those tests later.

Not a blocker, works for me.

dblock (approved):

PR #850, opened 2025-03-24
Add spec for neural search plugin stats API

q-andy:
### Description
Spec for neural stats API added to feature branch in this PR: https://github.com/opensearch-project/neural-search/pull/1256

Example API responses and test run output:
```
[INFO] => GET /_plugins/_neural/stats ({}) [application/json] 
[INFO] <= 200 (application/json; charset=UTF-8) | {
  "_nodes": {
    "total": 1,
    "successful": 1,
    "failed": 0
  },
  "cluster_name": "integTest",
  "info": {
    "cluster_version": "3.0.0",
    "processors": {
      "ingest": {
        "text_embedding_processors_in_pipelines": 0
      }
    }
  },
  "all_nodes": {
    "processors": {
      "ingest": {
        "text_embedding_executions": 0
      }
    }
  },
  "nodes": {
    "u4HvpBqoQPakwxqvrkOOWA": {
      "processors": {
        "ingest": {
          "text_embedding_executions": 0
        }
      }
    }
  }
}
```
```
[INFO] => GET /_plugins/_neural/stats/text_embedding_executions ({
  "flat_stat_paths": true,
  "include_metadata": true
}) [application/json] 
[INFO] <= 200 (application/json; charset=UTF-8) | {
  "_nodes": {
    "total": 1,
    "successful": 1,
    "failed": 0
  },
  "cluster_name": "integTest",
  "info": {},
  "all_nodes": {
    "processors.ingest.text_embedding_executions": {
      "value": 0,
      "stat_type": "timestamped_event_counter",
      "trailing_interval_value": 0,
      "minutes_since_last_event": 29047465
    }
  },
  "nodes": {
    "u4HvpBqoQPakwxqvrkOOWA": {
      "processors.ingest.text_embedding_executions": {
        "value": 0,
        "stat_type": "timestamped_event_counter",
        "trailing_interval_value": 0,
        "minutes_since_last_event": 29047465
      }
    }
  }
}
```

### Issues Resolved


dblock (changes requested):
If the neural plugin is enabled by default place tests under `default`. Adding something to tests/plugins/xyz doesn't automatically execute them, that requires updating CI test.yml.

q-andy:
Thanks for taking a look. This API is available from 3.0 beta onwards so I added a new entry in the test spect for it based on the release tag on [docker hub](https://hub.docker.com/r/opensearchstaging/opensearch/tags) following [this part](https://github.com/opensearch-project/opensearch-api-specification/blob/main/TESTING_GUIDE.md#future-releases) of the testing readme. I'm not super familiar with the CI and how the release process works, let me know if there's any additional changes needed

dblock:
> Thanks for taking a look. This API is available from 3.0 beta onwards so I added a new entry in the test spect for it based on the release tag on [docker hub](https://hub.docker.com/r/opensearchstaging/opensearch/tags) following [this part](https://github.com/opensearch-project/opensearch-api-specification/blob/main/TESTING_GUIDE.md#future-releases) of the testing readme. I'm not super familiar with the CI and how the release process works, let me know if there's any additional changes needed

This CI here may be confusing, so appreciate you hanging here with me. The documentation could use help to make it clearer (wink, wink).

1. We should be testing 3.0-latest whichever one it is, there's no point in testing both alpha and beta. So do that upgrade first (possibly in a separate PR).
2. The neural plugin feature exists in 3.0 by default, which means you do not need to make a custom test suite in tests/plugins, you want to add to the default folder of tests which runs against all versions.
3. Because the API is only available in 3.0, both the API spec and the tests need to be marked `>= 3.0`.

Does this make sense?

q-andy:
Yup, thanks for the swift reply!
1. Sure, I believe currently 3.0.0 beta was just cut, does it make sense to wait until 3.0 GA to make the upgrade, and then merge this PR afterwards? Since if we change all of them to 3.0.0 beta, I'll assume we'll have to do switch them to 3.0 GA afterwards anyways, similar to something like [this PR](https://github.com/opensearch-project/opensearch-api-specification/pull/665/files) I'm assuming
2. Ah, I don't believe neural exists in OpenSearch by default, it depends on kNN and ml-commons plugins which I see also have separate (non-default) specs so I assume it follows a similar pattern. I believe it makes sense to put them in the CI separately?
3. Sure, I'll add a `version: '>= 3.0'` line to the neural test yaml, similar to [this](https://github.com/opensearch-project/opensearch-api-specification/blob/main/tests/plugins/knn/knn/models.yaml#L4). For the API spec itself I've added the `x-version-added: '3.0'` tags, which should be enough?

dblock:
> 1. Sure, I believe currently 3.0.0 beta was just cut, does it make sense to wait until 3.0 GA to make the upgrade, and then merge this PR afterwards? Since if we change all of them to 3.0.0 beta, I'll assume we'll have to do switch them to 3.0 GA afterwards anyways, similar to something like [this PR](https://github.com/opensearch-project/opensearch-api-specification/pull/665/files) I'm assuming

Someone should upgrade to beta ASAP to test things before they get released. Including this plugin.

> 2. Ah, I don't believe neural exists in OpenSearch by default, it depends on kNN and ml-commons plugins which I see also have separate (non-default) specs so I assume it follows a similar pattern. I believe it makes sense to put them in the CI separately?

If the plugin is not installed by default then yes. You will need to customize Dockerfile accordingly.

> 3. Sure, I'll add a `version: '>= 3.0'` line to the neural test yaml, similar to [this](https://github.com/opensearch-project/opensearch-api-specification/blob/main/tests/plugins/knn/knn/models.yaml#L4). For the API spec itself I've added the `x-version-added: '3.0'` tags, which should be enough?

Yep.

q-andy:
@dblock [3.0.0-beta1](https://hub.docker.com/r/opensearchproject/opensearch/tags) released a few days ago, I've included changes for the version matrix bump in this PR. Happy to break it up into a separate PR as well if that makes more sense.

I've also verified neural-search is included in the release artifact by default with the following:
```
$ docker run -it --entrypoint bash opensearchproject/opensearch:3.0.0-beta1 -c "cat /usr/share/opensearch/manifest.yml" | grep neural
```
```
  - name: neural-search
    repository: https://github.com/opensearch-project/neural-search.git
    location: https://ci.opensearch.org/ci/dbc/distribution-build-opensearch/3.0.0-beta1/11026/linux/arm64/tar/builds/opensearch/plugins/opensearch-neural-search-3.0.0.0-beta1.zip
```

Some question about this:
> If the neural plugin is enabled by default place tests under default. Adding something to tests/plugins/xyz doesn't automatically execute them, that requires updating CI test.yml.

- So if neural-search is included as part of the release artifact then should it be under the default? If that's so, why would ml-commons and kNN be under test/plugins/xyz, shouldn't they also be undert default? 
- Should I still include a `- version 2.19.1` entry in the version matrix to ensure the integration tests still run against a 2.19.x version? I still there will be 2.19.2, 2.19.3 release in the future: https://opensearch.org/releases/

spec/namespaces/neural.yaml
```diff
@@ -0,0 +1,107 @@
...
+          schema:
+            $ref: '../schemas/neural._common.yaml#/components/schemas/NeuralStats'
+  parameters:
+    neural.stats::path.node_id:
+      name: node_id
+      in: path
+      description: Comma-separated list of node IDs or names to limit the returned information; leave empty to get information from all nodes.
+      schema:
+        type: string
+      required: true
```
yizheliu-amazon, line 75: is node_id required param?

.github/workflows/test-spec.yml
```diff
@@ -28,45 +28,47 @@ jobs:
             admin_password: admin
           - version: 2.0.0
             admin_password: admin
-          - version: 2.19.1
-          - version: 2.19.1
+          - version: 3.0.0-beta1
+          - version: 3.0.0-beta1
```
dblock, line 32: Do not change these as we need to test the latest released version most of the time.

.github/workflows/test-spec.yml
```diff
@@ -28,45 +28,47 @@ jobs:
...
-          - version: 2.19.1
+          - version: 3.0.0-beta1
             tests: plugins/learning_to_rank
-          - version: 2.19.1
+          - version: 3.0.0-beta1
             tests: plugins/security
             cert: tests/plugins/security/.kirk.pem
             key: tests/plugins/security/.kirk-key.pem
+          - version: 3.0.0-beta1
+            tests: plugins/neural
```
dblock, line 67: There's already tests/plugins/ml that sets up ml-commons, do you need a separate run? Add to those tests.

dblock (changes requested):
This repo tests multiple versions, undo all the changes from 2.19.1 to 3.0.0-beta1. You should do the upgrade from 3.0-alpha-1 to 3.0-beta-1 in a separate PR, then add tests for the neural stats here.

spec/namespaces/neural.yaml
q-andy: It's required in paths where node_id is in the url, otherwise it's not specified as a parameter

q-andy:
@dblock Addressed remaining comments and also the tests to default since neural search is installed by default. I've also added `x-version-added: 3.0` tags to the schema since we plan to add new stats with new versions. Since 3.0.0-beta1 has been added to the test specs in #883, the workflows should run successfully.

Tests pass locally with `npm run test:spec -- --opensearch-insecure --opensearch-url=http://localhost:9200  --tests tests/default/neural --verbose`.

CHANGELOG.md
```diff
@@ -76,6 +76,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Added `store` and `value_type` to TermsQuery  ([#860](https://github.com/opensearch-project/opensearch-api-specification/pull/860))
 - Added `search_only` to `IndexSettingBlocks` ([#883](https://github.com/opensearch-project/opensearch-api-specification/pull/883))
 - Added `searchOnly` to `NodeShard` ([#883](https://github.com/opensearch-project/opensearch-api-specification/pull/883))
+- Added specs for neural search plugin for stats API ([#850](https://github.com/opensearch-project/opensearch-api-specification/pull/850))
```
dblock, line 79:
```suggestion
- Added `GET /_plugins/_neural/stats` ([#850](https://github.com/opensearch-project/opensearch-api-specification/pull/850))
```

To be consistent with the rest of CHANGELOG.

dblock (approved):

dblock:
Changes look good. Fix spec linting / validation and it's good to go.

q-andy:
Thanks, fixed the failing workflows. All of the following commands run successfully:

```
npm run test:spec -- --opensearch-insecure --opensearch-url=http://localhost:9200  --tests tests/default/neural          
npm run lint
npm run lint:spec
```

dblock (approved):

Issue #786, opened 2025-01-13
[FEATURE] Add specifications for the `learning-to-rank`/`ltr` plugin


Xtansia:
### Is your feature request related to a problem?
The specification does not contain definitions for the [Learning to Rank Plugin](https://opensearch.org/docs/latest/search-plugins/ltr/index/).

### What solution would you like?
Specifications to be added for the plugin.

### What alternatives have you considered?
As it's not a default installed plugin we may not want to add it to this repo, but it is owned under the OpenSearch Github Org.

JohannesDaniel:
As this is quite comprehensive, I need someone to agree on how to do it - whether to do it in many steps or all at once.

nhtruong:
@JohannesDaniel We usually break something of this scale into multiple PRs. You can start with a small one with only 1 endpoint to get the feel of this first.

JohannesDaniel:
@nhtruong created first PR, passed all checks:
https://github.com/opensearch-project/opensearch-api-specification/pull/829

nhtruong:
Wonderful. Thank you. Approved and merged!

PR #861, opened 2025-04-08
Created specs and tests for LTR store endpoints

sstults:
### Description
Adds specs and tests for store actions in the LTR plugin.

### Issues Resolved
Partially addresses #786. Separate PRs are being submitted for each major component.


dblock:
Good start. Needs `version` fields in both spec and tests. Is this plugin optional? If so you'll need a custom setup in tests.

dblock (changes requested):
I mean all the versions :)

1. The test in tests/plugins/learning_to_rank/ltr/store.yaml does not run as part of CI without extra work. Either the plugin is available and installed by default in 2.19, in which case the test should go into `default`, or it's not installed by default in which case you need to wire up `tests/plugins/learning_to_rank` into CI - see the test.yaml.
2. In the spec you need to add `version` tags to say since when this is available.

Hope this helps!

sstults:
I think I got it. I was a little unsure about the differing URI roots between the stats endpoints and the rest of the operations post-2.19, but I think I at least got the versions right.

dblock:
Making good progress! See linter/CI failures, otherwise looking good.

PR #876, opened 2025-04-18
[Docs]: Add Index API schema descriptions

Naarcha-AWS:
Adds missing descriptions for all Index schemas.

### DIFF
--- a/spec/schemas/indices._common.yaml
+++ b/spec/schemas/indices._common.yaml
@@ -1303,7 +1303,7 @@ components:
           $ref: '_common.yaml#/components/schemas/ByteCount'
         indices:
           type: object
-          description: The upgrade status for individual indices.
+          description: The upgrade status for individual indexes.
           additionalProperties:
             $ref: '#/components/schemas/UpgradeStatus'
     UpgradeStatus:
--- a/spec/schemas/indices._common.yaml
+++ b/spec/schemas/indices._common.yaml
@@ -953,20 +953,26 @@ components:
           items:
             type: string
     StorageType:
+      anyOf:
+        - title: builtin
+          $ref: '#/components/schemas/BuiltinStorageType'
+        - title: custom
+          type: string
+    BuiltinStorageType:
       description: The type of storage implementation.
       oneOf:
         - description: Uses the default file system implementation.
           const: fs
           type: string
-        - description: Uses a hybrid of memory-mapped and regular file system implementation.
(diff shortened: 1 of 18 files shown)


spec/schemas/indices._common.yaml
```diff
@@ -816,46 +913,63 @@ components:
...
+          type: string
+        - description: Uses a hybrid of memory-mapped and regular file system implementation.
+          const: hybridfs
+          type: string
+        - description: Uses memory-mapped file system implementation.
+          const: mmapfs
+          type: string
+        - description: Uses NIO file system implementation.
+          const: niofs
+          type: string
```
Xtansia, line 969: This change is not equivalent, the original allows arbitrary strings on top of the enum defined ones. Which is resulting in the CI failures.

I would potentially split these into two separate definitions like so:
```yaml
StorageType:
  anyOf:
    - title: builtin
      $ref: '#/components/schemas/BuiltinStorageType'
    - title: custom
      type: string
BuiltinStorageType:
  description: The type of storage implementation.
  oneOf:
    - description: Uses the default file system implementation.
      const: fs
      type: string
    - description: Uses a hybrid of memory-mapped and regular file system implementation.
      const: hybridfs
      type: string
    - description: Uses memory-mapped file system implementation.
      const: mmapfs
      type: string
    - description: Uses NIO file system implementation.
      const: niofs
      type: string
```

We have this same pattern elsewhere: https://github.com/opensearch-project/opensearch-api-specification/blob/5fe396d6ec52d72f00f13448bdada7c8109b6ca8/spec/schemas/_common.yaml#L528

Xtansia (changes requested):

Naarcha-AWS:
@Xtansia: Addressed.

spec/schemas/indices._common.yaml
```diff
@@ -1155,10 +1303,12 @@ components:
           $ref: '_common.yaml#/components/schemas/ByteCount'
         indices:
           type: object
+          description: The upgrade status for individual indices.
```
Xtansia, line 1306: Per Vale lint, this should be:
```suggestion
          description: The upgrade status for individual indexes.
```

Xtansia (changes requested):

Naarcha-AWS:
@Xtansia: Feedback implemented.

Xtansia (approved):

PR #879, opened 2025-04-25
Bump latest default test version to 3.0.0-beta1

q-andy:
### Description
Bumps latest integ test version to latest release 3.0.0-beta1, replacing 2.20 and 3.0.0-alpha1 as discussed in #850 

### Issues Resolved
n/a

### DIFF
--- a/.github/workflows/test-spec.yml
+++ b/.github/workflows/test-spec.yml
@@ -66,7 +66,7 @@ jobs:
           - version: 2.20.0
             hub: opensearchstaging
             ref: '@sha256:037a2eb5a4f48b9a9a29950be5e5fae4ebd82d41de1066e00bb8d8b0ce7871b1'
-          - version: 3.0.0-alpha1
+          - version: 3.0.0-beta1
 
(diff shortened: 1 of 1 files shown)


.github/workflows/test-spec.yml
```diff
@@ -63,10 +63,7 @@ jobs:
             tests: plugins/security
             cert: tests/plugins/security/.kirk.pem
             key: tests/plugins/security/.kirk-key.pem
-          - version: 2.20.0
```
dblock, line 66: On a second thought maybe we should leave this one in here in case there's a 2.20? It also shows how to do a `ref`.

dblock:
Looks like a few tests are failing which isn't expected, likely a regression in the server or a spec change that needs to be fixed.

.github/workflows/test-spec.yml
peterzhuamazon: There is no plan for 2.20 for the time being, 2.19.x will be the last version of 2.x.
I think we are safe to remove it.

Thanks.

prudhvigodithi:
I have a PR https://github.com/opensearch-project/opensearch-api-specification/pull/883 created for API Spec compatible with 3.0.0-beta1.
@dblock @q-andy please take a look. Thanks

q-andy:
Thanks for the update @prudhvigodithi, closing this in favor of #883

PR #892, opened 2025-05-12
Change ScriptedMetricAggregate value from true to {}

lucy66hw:
### Description
Change ScriptedMetricAggregate value from true to {}

### DIFF
--- a/spec/schemas/_common.aggregations.yaml
+++ b/spec/schemas/_common.aggregations.yaml
@@ -1268,7 +1268,7 @@ components:
         - $ref: '#/components/schemas/AggregateBase'
         - type: object
           properties:
-            value: true
+            value: {}
           required:
             - value
     TopHitsAggregate:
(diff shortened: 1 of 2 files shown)


spec/schemas/_common.aggregations.yaml
```diff
@@ -1268,7 +1268,7 @@ components:
         - $ref: '#/components/schemas/AggregateBase'
         - type: object
           properties:
-            value: true
+            value: {}
```
karenyrx, line 1271: just curious, is this the same as
```
value
    type: object
```
, which was what it was before? 
image (62)

do we have a preference/standard for which one to use for generic object types?

Xtansia: @karenyrx `value` isn't necessarily an object, it can also be a float, string etc. 
It's of arbitrary type which is why it's unspecified

Xtansia (approved):

Issue #238, opened 2024-04-11
[FEATURE] Add specs for `geospatial` namespace


nhtruong:
Part of [Adding missing API specs issue](https://github.com/opensearch-project/opensearch-api-specification/issues/168)

- [ ] _plugins/geospatial/_upload/stats: ["GET"]
- [ ] _plugins/geospatial/geojson/_upload: ["PUT","POST"]
- [ ] _plugins/geospatial/ip2geo/datasource: ["GET"]
- [ ] _plugins/geospatial/ip2geo/datasource/{name}: ["GET","PUT","DELETE"]
- [ ] _plugins/geospatial/ip2geo/datasource/{name}/_settings: ["PUT"]

VijayanB:
Tagging @vamshin for action

PR #893, opened 2025-05-12
Add API Spec for Geospatial plugin

weijia-aws:
### Description
This PR adds API Spec for [geospatial plugin](https://github.com/opensearch-project/geospatial)

### Tests
```
npm run test:spec -- --opensearch-insecure  --tests tests/default/geospatial 

> ts-node tools/src/tester/test.ts --opensearch-insecure --tests tests/default/geospatial

OpenSearch 2.19.0

PASSED  geojson/upload.yaml (tests/default/geospatial/geojson/upload.yaml)
PASSED  ip2geo/datasource.yaml (tests/default/geospatial/ip2geo/datasource.yaml)
PASSED  upload/stats.yaml (tests/default/geospatial/upload/stats.yaml)
PASSED  ip2geo/datasource/settings.yaml (tests/default/geospatial/ip2geo/datasource/settings.yaml)

Tested 8/611 paths.


```

```
npm run lint:spec

> opensearch_api_tools@1.0.0 lint:spec
> ts-node tools/src/linter/lint.ts

Validating /Users/zweijia/IdeaProjects/opensearch-api-specification/spec ...
No errors found.
```

### Issues Resolved
Closes [#238](https://github.com/opensearch-project/opensearch-api-specification/issues/238)
Closes https://github.com/opensearch-project/geospatial/issues/658 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -78,6 +78,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Added `searchOnly` to `NodeShard` ([#883](https://github.com/opensearch-project/opensearch-api-specification/pull/883))
 - Added `GET /_plugins/_neural/stats` ([#850](https://github.com/opensearch-project/opensearch-api-specification/pull/850))
 - Added `ToolAttributes` to `ml._common.yaml` ([#878](https://github.com/opensearch-project/opensearch-api-specification/pull/878)) 
+- Added `GET _plugins/geospatial/_upload/stats`, `PUT`, `POST _plugins/geospatial/geojson/_upload`, `GET _plugins/geospatial/ip2geo/datasource`, `GET`, `PUT`, `DELETE _plugins/geospatial/ip2geo/datasource/{name}` and `PUT _plugins/geospatial/ip2geo/datasource/{name}/_settings` ([#893](https://github.com/opensearch-project/opensearch-api-specification/pull/893))
 
 ### Removed
 - Removed unsupported `_common.mapping:SourceField`'s `mode` field and associated `_common.mapping:SourceFieldMode` enum ([#652](https://github.com/opensearch-project/opensearch-api-specification/pull/652))
--- a/spec/namespaces/geospatial.yaml
+++ b/spec/namespaces/geospatial.yaml
@@ -43,22 +43,34 @@ paths:
         '200':
           $ref: '#/components/responses/geospatial.get_upload_stats@200'
 
+  /_plugins/geospatial/ip2geo/datasource:
+    get:
+      operationId: geospatial.get_ip2geo_datasource.0
+      x-operation-group: geospatial.get_ip2geo_datasource
+      x-version-added: '2.11'
+      description: Get one or more IP2Geo data sources, defaulting to returning all if no names specified.
+      externalDocs:
+        url: 'https://docs.opensearch.org/docs/latest/ingest-pipelines/processors/ip2geo/#sending-a-get-request'
+      responses:
+        '200':
+          $ref: '#/components/responses/geospatial.get_ip2geo_datasource@200'
+
   /_plugins/geospatial/ip2geo/datasource/{name}:
     get:
-      operationId: geospatial.get_ip_to_geo_datasource.0
-      x-operation-group: geospatial.get_ip_to_geo_datasource
+      operationId: geospatial.get_ip2geo_datasource.1
+      x-operation-group: geospatial.get_ip2geo_datasource
       x-version-added: '2.11'
       description: Get one or more IP2Geo data sources, defaulting to returning all if no names specified.
       externalDocs:
         url: 'https://docs.opensearch.org/docs/latest/ingest-pipelines/processors/ip2geo/#sending-a-get-request'
       parameters:
-        - $ref: '#/components/parameters/geospatial.get_ip_to_geo_datasource::path.name'
+        - $ref: '#/components/parameters/geospatial.get_ip2geo_datasource::path.name'
       responses:
         '200':
-          $ref: '#/components/responses/geospatial.get_ip_to_geo_datasource@200'
+          $ref: '#/components/responses/geospatial.get_ip2geo_datasource@200'
     put:
-      operationId: geospatial.put_ip_to_geo_datasource.0
-      x-operation-group: geospatial.put_ip_to_geo_datasource
+      operationId: geospatial.put_ip2geo_datasource.0
+      x-operation-group: geospatial.put_ip2geo_datasource
       x-version-added: '2.11'
       description: |-
         Create a specific IP2Geo data source.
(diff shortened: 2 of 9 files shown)


Xtansia (changes requested):

CHANGELOG.md
```diff
@@ -78,6 +78,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Added `searchOnly` to `NodeShard` ([#883](https://github.com/opensearch-project/opensearch-api-specification/pull/883))
 - Added `GET /_plugins/_neural/stats` ([#850](https://github.com/opensearch-project/opensearch-api-specification/pull/850))
 - Added `ToolAttributes` to `ml._common.yaml` ([#878](https://github.com/opensearch-project/opensearch-api-specification/pull/878)) 
+- Added `GET _plugins/geospatial/_upload/stats`, `PUT`, `POST _plugins/geospatial/geojson/_upload`, `GET _plugins/geospatial/ip2geo/datasource`, `GET`, `PUT`, `DELETE _plugins/geospatial/ip2geo/datasource/{name}` and `PUT _plugins/geospatial/ip2geo/datasource/{name}/_settings` ([#238](https://github.com/opensearch-project/opensearch-api-specification/issues/238))
```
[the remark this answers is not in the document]
weijia-aws: Done

spec/namespaces/geospatial.yaml
```diff
@@ -0,0 +1,246 @@
...
+          $ref: '#/components/responses/geospatial.ip_to_geo_get_datasource@200'
+    put:
+      operationId: geospatial.ip_to_geo_put_datasource.0
+      x-operation-group: geospatial.ip_to_geo_put_datasource
+      x-version-added: '2.11'
+      description: |-
+        Create a specific IP2Geo data source.
+        Default values:
+          - endpoint: "https://geoip.maps.opensearch.org/v1/geolite2-city/manifest.json"
+          - update_interval_in_days: 3.
```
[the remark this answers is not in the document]
weijia-aws: Done

spec/namespaces/geospatial.yaml
```diff
@@ -0,0 +1,246 @@
...
+        $ref: '#/components/requestBodies/geospatial.geojson_upload_post'
+      responses:
+        '200':
+          $ref: '#/components/responses/geospatial.geojson_upload_post@200'
+        '400':
+          $ref: '#/components/responses/geospatial.geojson_upload_post@400'
+    put:
+      operationId: geospatial.geojson_upload_put.0
+      x-operation-group: geospatial.geojson_upload_put
+      x-version-added: '2.11'
```
Xtansia, line 26: As these appear to be functionally identical, I'd give these the same operation group to allow generated clients to collapse them into one operation. Also so that the grouping reads more like the action being performed I'd potentially rearrange the naming to `geospatial.upload_geojson`.

spec/namespaces/geospatial.yaml
```diff
@@ -0,0 +1,246 @@
...
+        - When type is `geo_shape`, all geometry types are allowed (Point, MultiPoint, LineString, MultiLineString, Polygon, MultiPolygon, GeometryCollection, Envelope).
+      requestBody:
+        $ref: '#/components/requestBodies/geospatial.geojson_upload_put'
+      responses:
+        '200':
+          $ref: '#/components/responses/geospatial.geojson_upload_put@200'
+  /_plugins/geospatial/_upload/stats:
+    get:
+      operationId: geospatial.upload_stats.0
+      x-operation-group: geospatial.upload_stats
```
Xtansia, line 39: Because `upload` is also an action, it may be clearer to prefix this with `get_`, to make it explicit this isn't for uploading stats.

spec/namespaces/geospatial.yaml
```diff
@@ -0,0 +1,246 @@
...
+      x-operation-group: geospatial.ip_to_geo_get_datasource
+      x-version-added: '2.11'
+      description: Get information for a specific IP2Geo data source.
+      externalDocs:
+        url: 'https://docs.opensearch.org/docs/latest/ingest-pipelines/processors/ip2geo/#sending-a-get-request'
+      parameters:
+        - $ref: '#/components/parameters/geospatial.ip_to_geo_get_datasource::path.datasource_name'
+      responses:
+        '200':
+          $ref: '#/components/responses/geospatial.ip_to_geo_get_datasource@200'
```
Xtansia, line 70: As these have identical response structures, the pattern used elsewhere in the spec is to give these the same operation group, and allow generated clients to make overloads and pick the path based on whether a datasource name was provided.

Again also preferring the action as the start of the name, in this case I'd prefer `geospatial.get_ip_to_geo_datasource`.

I'd also be fine with these using `get_ip2geo_datasource` for conciseness and matching the naming of the data format, however the regex [here](https://github.com/opensearch-project/opensearch-api-specification/blob/main/tools/src/linter/components/Operation.ts#L14) currently disallows digits, I think it'd be reasonable to allow digits as long as it isn't the first character of either part of the group. Thoughts @dblock @nhtruong ?

spec/namespaces/geospatial.yaml
```diff
@@ -0,0 +1,246 @@
...
+    put:
+      operationId: geospatial.ip_to_geo_put_datasource.0
+      x-operation-group: geospatial.ip_to_geo_put_datasource
```
Xtansia, line 73: As above regarding naming. ie. `put_ip_to_geo_datasource`/`put_ip2geo_datasource`.

spec/namespaces/geospatial.yaml
```diff
@@ -0,0 +1,246 @@
...
+      requestBody:
+        $ref: '#/components/requestBodies/geospatial.ip_to_geo_put_datasource'
+      responses:
+        '200':
+          $ref: '#/components/responses/geospatial.ip_to_geo_put_datasource@200'
+        '400':
+          $ref: '#/components/responses/geospatial.ip_to_geo_put_datasource@400'
+    delete:
+      operationId: geospatial.ip_to_geo_delete_datasource.0
+      x-operation-group: geospatial.ip_to_geo_delete_datasource
```
Xtansia, line 93: As above regarding naming.  ie. `delete_ip_to_geo_datasource`/`delete_ip2geo_datasource`.

spec/namespaces/geospatial.yaml
```diff
@@ -0,0 +1,246 @@
...
+        - $ref: '#/components/parameters/geospatial.ip_to_geo_delete_datasource::path.datasource_name'
+      responses:
+        '200':
+          $ref: '#/components/responses/geospatial.ip_to_geo_delete_datasource@200'
+        '404':
+          $ref: '#/components/responses/geospatial.ip_to_geo_delete_datasource@404'
+  /_plugins/geospatial/ip2geo/datasource/{datasource_name}/_settings:
+    put:
+      operationId: geospatial.ip_to_geo_datasource_settings.0
+      x-operation-group: geospatial.ip_to_geo_datasource_settings
```
Xtansia, line 108: In case there's a chance of a get variant in future, this should ideally be `put_ip_to_geo_datasource_settings`/`put_ip2geo_datasource_settings`.

spec/schemas/geospatial._common.yaml
```diff
@@ -0,0 +1,392 @@
...
+                - $ref: '#/components/schemas/Polygon'
+            - allOf:
+                - $ref: '#/components/schemas/GeoShapes'
+                - $ref: '#/components/schemas/MultiPolygon'
+            - allOf:
+                - $ref: '#/components/schemas/GeoShapes'
+                - $ref: '#/components/schemas/GeometryCollection'
+            - allOf:
+                - $ref: '#/components/schemas/GeoShapes'
+                - $ref: '#/components/schemas/Envelope'
```
Xtansia, line 231: This doesn't entirely make sense as the `allOf`s are generally duplicating what would already be checked by `GeoShapes`. Should this not instead just be `$ref: '#/components/schemas/Geometry`?

Xtansia (changes requested):

spec/namespaces/geospatial.yaml
weijia-aws: I tried to do that initially, but it violates lint:spec. Here's what it complains:

```
Validating /Users/zweijia/IdeaProjects/opensearch-api-specification/spec ...
[ERROR] Errors found:

{
  file: 'namespaces/geospatial.yaml',
  location: 'Operation Group: geospatial.geojson_upload',
  message: "2 'geospatial.geojson_upload' operations must have identical description property."
}
{
  file: 'namespaces/geospatial.yaml',
  location: 'Operation Group: geospatial.geojson_upload',
  message: "2 'geospatial.geojson_upload' operations must have an identical set of responses."
}
[ERROR]
```

spec/namespaces/geospatial.yaml
weijia-aws: Sounds good

spec/namespaces/geospatial.yaml
weijia-aws: > As these have identical response structures, the pattern used elsewhere in the spec is to give these the same operation group, and allow generated clients to make overloads and pick the path based on whether a datasource name was provided.

Although these have the same response structure, the request is different. For the first one, it doesn't require to pass in a {datasource_name} parameter , while the second one requires. Also the description is different as well. If we move to a same group, it will violate the lint rule again.

> Again also preferring the action as the start of the name, in this case I'd prefer geospatial.get_ip_to_geo_datasource.

Will do

> I'd also be fine with these using get_ip2geo_datasource for conciseness and matching the naming of the data format, however the regex [here](https://github.com/opensearch-project/opensearch-api-specification/blob/main/tools/src/linter/components/Operation.ts#L14) currently disallows digits, I think it'd be reasonable to allow digits as long as it isn't the first character of either part of the group. Thoughts @dblock @nhtruong ?

Yeah, I was using ip2geo_datasource first, and it violate the regex

spec/schemas/geospatial._common.yaml
weijia-aws: Yes, your are right!

spec/namespaces/geospatial.yaml
Xtansia: Ah, I just noticed the `operation will fail if index exists.` vs `regardless if index exists.`, that does mean they shouldn't actually be the same operation. Naming will be tricky to communicate that difference.

spec/namespaces/geospatial.yaml
Xtansia: > > As these have identical response structures, the pattern used elsewhere in the spec is to give these the same operation group, and allow generated clients to make overloads and pick the path based on whether a datasource name was provided.
> 
> Although these have the same response structure, the request is different. For the first one, it doesn't require to pass in a {datasource_name} parameter , while the second one requires. Also the description is different as well. If we move to a same group, it will violate the lint rule again.
> 

Based on the code https://github.com/opensearch-project/geospatial/blob/main/src/main/java/org/opensearch/geospatial/ip2geo/action/RestGetDatasourceHandler.java they're not really separate actions, as the `{datasource_name}` can also be a comma delimited array, so it's just "Get one or more IP2Geo data sources, defaulting to returning all if no names specified".

spec/namespaces/geospatial.yaml
```diff
@@ -0,0 +1,246 @@
...
+          $ref: '#/components/responses/geospatial.put_ip_to_geo_datasource_settings@404'
+
+components:
+  parameters:
+    geospatial.get_ip_to_geo_datasource::path.datasource_name:
+      name: datasource_name
+      in: path
+      required: true
+      schema:
+        type: string
```
Xtansia, line 130: Given this can be a comma delimited array, this should be:
```suggestion
        $ref: '../schemas/_common.yaml#/components/schemas/StringOrStringArray'
```

Xtansia (commented):
Additionally all the `{datasource_name}` path params can just be `{name}`

spec/namespaces/geospatial.yaml
weijia-aws: I see, then I can merge these two actions into one

spec/namespaces/geospatial.yaml
```diff
@@ -0,0 +1,228 @@
...
+    get:
+      operationId: geospatial.get_upload_stats.0
+      x-operation-group: geospatial.get_upload_stats
+      x-version-added: '2.11'
+      description: Retrieves statistics for all geospatial uploads.
+      responses:
+        '200':
+          $ref: '#/components/responses/geospatial.get_upload_stats@200'
+
+  /_plugins/geospatial/ip2geo/datasource/{name}:
```
Xtansia, line 46: You still need the non `{name}` path variant, just giving them the same operation-group, description and response $ref

Xtansia (commented):
I would update the [operation group regex](https://github.com/opensearch-project/opensearch-api-specification/blob/main/tools/src/linter/components/Operation.ts#L14C1-L14C70) to:
```typescript
const GROUP_REGEX = /^([a-z]+[a-z_]*[a-z]+\.)?([a-z]+[a-z0-9_]*[a-z0-9]+)$/
```

And rename all the `ip_to_geo`s to `ip2geo`

spec/namespaces/geospatial.yaml
weijia-aws: Done

tools/src/linter/components/Operation.ts
```diff
@@ -11,7 +11,7 @@ import { type OperationSpec, type ValidationError } from 'types'
 import _ from 'lodash'
 import ValidatorBase from './base/ValidatorBase'
 
-const GROUP_REGEX = /^([a-z]+[a-z_]*[a-z]+\.)?([a-z]+[a-z_]*[a-z]+)$/
+const GROUP_REGEX = /^([a-z]+[a-z_]*[a-z]+\.)?([a-z]+[a-z0-9_]*[a-z0-9]+)$/
```
Xtansia, line 14: Looks like this regex appears in error messages which are tested verbatim, so will need updating:
- https://github.com/opensearch-project/opensearch-api-specification/blob/main/tools/tests/linter/Operation.test.ts#L22
- https://github.com/opensearch-project/opensearch-api-specification/blob/main/tools/tests/linter/Operation.test.ts#L26

weijia-aws: Updated

Xtansia (approved):

PR #904, opened 2025-05-29
Add template to QueryContainer

nathaliellenaa:
### Description
Add `template` to `QueryContainer`

### Issues Resolved
Resolves https://github.com/opensearch-project/opensearch-java/issues/1583

### DIFF
--- a/spec/schemas/_common.query_dsl.yaml
+++ b/spec/schemas/_common.query_dsl.yaml
@@ -214,6 +214,9 @@ components:
           maxProperties: 1
         span_within:
           $ref: '#/components/schemas/SpanWithinQuery'
+        template:
+          description: Returns documents using a search template with placeholder values.
+          $ref: '#/components/schemas/TemplateQuery'
         term:
           description: |-
             Returns documents that contain an exact term in a provided field.
@@ -2094,6 +2097,28 @@ components:
           required:
             - big
             - little
+    TemplateQuery:
+      type: object
+      properties:
+        knn:
+          $ref: '#/components/schemas/TemplateKnnQuery'
+    TemplateKnnQuery:
+      type: object
+      properties:
+        text_embedding:
+          $ref: '_common.query_dsl.yaml#/components/schemas/VectorField'
+        vector_field:
+          $ref: '_common.query_dsl.yaml#/components/schemas/VectorField'
+      additionalProperties: true
+    VectorField:
+      type: object
+      properties:
+        vector:
+          $ref: '#/components/schemas/QueryVector'
+        k:
+          description: The total number of nearest neighbors to return as top hits.
+          type: integer
+          format: int32
     TermQuery:
       oneOf:
         - title: value
--- a/spec/schemas/_common.query_dsl.yaml
+++ b/spec/schemas/_common.query_dsl.yaml
@@ -215,8 +215,9 @@ components:
         span_within:
           $ref: '#/components/schemas/SpanWithinQuery'
         template:
-          description: Returns documents using a search template with placeholder values.
-          $ref: '#/components/schemas/TemplateQuery'
+          description: Return documents using a template query contains placeholders that are resolved by search request processors during query execution.
+          type: object
+          additionalProperties: true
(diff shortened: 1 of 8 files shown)


spec/schemas/_common.query_dsl.yaml
```diff
@@ -2094,6 +2097,28 @@ components:
           required:
             - big
             - little
+    TemplateQuery:
+      type: object
+      properties:
+        knn:
+          $ref: '#/components/schemas/TemplateKnnQuery'
+    TemplateKnnQuery:
```
ylwu-amzn, line 2105: I don't think we should add this `TemplateKnnQuery` , template query can support any opensearch query, knn is just one kind . https://docs.opensearch.org/docs/latest/search-plugins/search-relevance/template-query/

spec/schemas/_common.query_dsl.yaml
```diff
@@ -214,6 +214,9 @@ components:
           maxProperties: 1
         span_within:
           $ref: '#/components/schemas/SpanWithinQuery'
+        template:
+          description: Returns documents using a search template with placeholder values.
```
mingshl, line 218: search template is another feature, let's rephrase this sentence to be more clear.  

```
Return documents using a template query contains placeholders that are resolved by search request processors during query execution. 
```

spec/schemas/_common.query_dsl.yaml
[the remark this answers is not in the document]
nathaliellenaa: Yes, we will have `additionalProperties` set to `true` for `template` since we don't have fixed structure. ([reference](https://github.com/opensearch-project/opensearch-api-specification/pull/904#discussion_r2114922761))

nathaliellenaa:
@Xtansia The test is failing in `1.3.17` and `2.0.0` version because the register model API was introduced in `2.7` version. I added `version: '>= 2.19'` to the test prologues, but the endpoint is still being run in the previous versions. I tried moving the test to a different yaml file, but got an `Invalid path detected` warning. Should we update the version requirement to `2.7` in this test file, or would you recommend a different approach?
```
ERROR   _core/search/knn/search.yaml (tests/default/_core/search/knn/search.yaml)
    ERROR   PROLOGUES
        ERROR   PUT /_cluster/settings (persistent setting [plugins.ml_commons.jvm_heap_memory_threshold], not recognized)
        PASSED  PUT /movies
        PASSED  POST /movies/_doc
        ERROR   POST /_plugins/_ml/models/_register (Method Not Allowed
Expected to find non undefined value at `payload.task_id`.)
        ERROR   GET /_plugins/_ml/tasks/{task_id} (Fail to find task
Expected to find non undefined value at `payload.model_id`.)
        ERROR   POST /_plugins/_ml/models/{model_id}/_deploy (Bad Request
Expected to find non undefined value at `payload.task_id`.)
        ERROR   GET /_plugins/_ml/tasks/{task_id} (Fail to find task
Expected to find non undefined value at `payload.model_id`.)
        ERROR   PUT /_search/pipeline/ml_inference_pipeline (Bad Request)
    ERROR   EPILOGUES
        ERROR   POST /_plugins/_ml/models/{model_id}/_undeploy (Bad Request)
        PASSED  DELETE /_plugins/_ml/models/{model_id}
        PASSED  DELETE /movies
```

dblock (changes requested):
There's a lot going on here, maybe split these tests out, instead of `tests/default/_core/search/knn/search.yaml` put this into `tests/default/_core/search/knn/search/search_pipeline.yaml`? Maybe reorganize other tests too?

nathaliellenaa:
> There's a lot going on here, maybe split these tests out, instead of `tests/default/_core/search/knn/search.yaml` put this into `tests/default/_core/search/knn/search/search_pipeline.yaml`? Maybe reorganize other tests too?

I reorganize the tests under `tests/default/_core/search/knn/search.yaml` into different files under `tests/default/_core/search/knn/search` folder

dblock (approved):

PR #933, opened 2025-06-24
Fix integer overflow issue in DocStatus by changing Integer to Long

donggyu04:
### Description

Since version 3.0.0 when calling `OpenSearchClient.nodes().stats()`, the function call fails on certain clusters with an error such as 
```
java.lang.RuntimeException: Jackson exception: Numeric value (5069719939) out of range of int (-2147483648 - 2147483647)
```
While debugging, it appears that the overflow error occurs because the `2xx` value in `doc_status` on a specific coordinator node exceeds the range of an integer.

image

To resolve this issue, the type of the `2xx` field was changed to `Long`.

### Issues Resolved

- https://github.com/opensearch-project/opensearch-java/issues/1645


reta (approved):

Xtansia (changes requested):
Please also update the other `1xx`, `3xx`, `4xx` and `5xx` properties to `int64`. And also add an entry to the CHANGELOG.md

donggyu04:
@Xtansia I applied your request through https://github.com/opensearch-project/opensearch-api-specification/pull/933/commits/fbc4c58057d766e7d8a65ebda9fc0076ceb78435

Xtansia (approved):
Thanks @donggyu04!

Issue #636, opened 2024-10-22
[FEATURE] Add specifications for Derived Fields


Xtansia:
### Is your feature request related to a problem?
The unsupported "runtime fields" were removed from the spec in #634, these should be replaced with the specs for "derived fields".

### Do you have any additional context?
- https://github.com/opensearch-project/OpenSearch/issues/1133
- https://github.com/opensearch-project/OpenSearch/issues/12281
- https://opensearch.org/docs/latest/field-types/supported-field-types/derived/

PR #976, opened 2025-10-24
Add derived fields

chriswhite199:
### Description
Add in specification for derived fields:
* https://docs.opensearch.org/latest/mappings/supported-field-types/derived/

### Issues Resolved
Closes [#636]

### DIFF
--- a/tests/default/indices/index_template.yaml
+++ b/tests/default/indices/index_template.yaml
@@ -39,16 +39,6 @@ chapters:
             number_of_shards: 4
             number_of_replicas: 4
           mappings:
-            derived:
-              inlineDerived:
-                type: text
-                script:
-                  lang: painless
-                  source: emit('one')
-              storedDerived:
-                type: text
-                script:
-                  id: script-id
             properties:
               timestamp:
                 type: date


Xtansia (approved):

spec/namespaces/_core.yaml
```diff
@@ -2581,6 +2581,11 @@ components:
                   $ref: '../schemas/_common.aggregations.yaml#/components/schemas/AggregationContainer'
               collapse:
                 $ref: '../schemas/_core.search.yaml#/components/schemas/FieldCollapse'
+              derived:
+                x-version-added: '2.15'
+                type: object
+                additionalProperties:
+                  $ref: '../schemas/_common.mapping.yaml#/components/schemas/DerivedMapping'
```
Xtansia: This isn't needed here, it already exists at line 2722

Xtansia (changes requested):

spec/namespaces/_core.yaml
chriswhite199: Ok i've rebased to the upstream and updated to account for #981.

@Xtansia - can you re-review and comment?

Issue #992, opened 2025-11-19
[FEATURE] add semantic field type


zzfrankz:
[semantic field type](https://docs.opensearch.org/latest/mappings/supported-field-types/semantic/) is not exposed through java client yet.

### Is your feature request related to a problem?
[semantic field type](https://docs.opensearch.org/latest/mappings/supported-field-types/semantic/) is not exposed through java client yet.

### What solution would you like?
add it in the field mapping config, generate yaml configs and propagate it to all clients. 

### What alternatives have you considered?
N/A

### Do you have any additional context?
N/A

PR #995, opened 2025-11-20
semantic field type

zzfrankz:
### Description
[Semantic field type](https://docs.opensearch.org/latest/mappings/supported-field-types/semantic/) was introduced in OS since 3.1, the PR will enable the semantic field type in different OpenSearch clients.

### Issues Resolved
Closes [[FEATURE] add semantic field type](https://github.com/opensearch-project/opensearch-api-specification/issues/992)


tests/plugins/ml/mapping/semantic_advanced.yaml.tmp
```diff
@@ -0,0 +1,337 @@
+$schema: ../../../../json_schemas/test_story.schema.yaml
+
+description: Test advanced semantic field type mappings with all configuration options.
```
karenyrx: this tmp file can be removed, along with the `*.md` files

zzfrankz: removed .md files, they caused the yaml parsing error as well
https://github.com/opensearch-project/opensearch-api-specification/actions/runs/19584604338/job/56134611280?pr=995

spec/namespaces/indices.yaml
```diff
@@ -1026,6 +1026,10 @@ paths:
       responses:
         '200':
           $ref: '#/components/responses/indices.create@200'
+        '400':
```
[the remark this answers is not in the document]
zzfrankz: a good catch, the error is `500` if the mode is not loaded and the model id is invalid 

`{
  "error": {
    "root_cause": [
      {
        "type": "runtime_exception",
        "reason": "runtime_exception: Failed to fetch model [invalid-model-id]: Failed to find model with the provided model id: invalid-model-id"
      }
    ],
    "type": "runtime_exception",
    "reason": "runtime_exception: Failed to fetch model [invalid-model-id]: Failed to find model with the provided model id: invalid-model-id"
  },
  "status": 500
}`

PR #1003, opened 2025-12-04
Add array type to `fields` in `Highlight` and `default` in `HighlighterTagsSchema`

karenyrx:
### Description

1. tags schema is missing ["default" option ](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/search/fetch/subphase/highlight/HighlightBuilder.java#L237-L240)

2. Fields supports object or array type. If object type, then useExplicitFieldOrder is set to false. If array type, then true.  Code path: 
* https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/search/fetch/subphase/highlight/HighlightBuilder.java#L301 
-> https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/search/fetch/subphase/highlight/HighlightBuilder.java#L295 
->  https://github.com/opensearch-project/OpenSearch/blob/main/libs/core/src/main/java/org/opensearch/core/xcontent/ObjectParser.java#L498-L532 )
```
1. JSON parser encounters "fields": ...
2. Checks if it's { (object) or [ (array)
3. If [ (array) → calls orderedModeCallback → sets useExplicitFieldOrder(true)
5. If { (object) → callback never called → useExplicitFieldOrder stays false
```
### Issues Resolved

### DIFF
--- a/spec/schemas/_core.search.yaml
+++ b/spec/schemas/_core.search.yaml
@@ -819,12 +819,12 @@ components:
             $ref: '#/components/schemas/HighlightField'
         - type: array
           items:
-            - type: object
-              additionalProperties:
-                title: key
-                $ref: '#/components/schemas/HighlightField'
-              minProperties: 1
-              maxProperties: 1
+            type: object
+            additionalProperties:
+              title: key
+              $ref: '#/components/schemas/HighlightField'
+            minProperties: 1
+            maxProperties: 1
           x-protobuf-excluded: true
     HighlighterEncoder:
       type: string


spec/schemas/_core.search.yaml
```diff
@@ -819,7 +819,16 @@ components:
             $ref: '#/components/schemas/HighlightField'
         - type: array
           items:
-            $ref: '#/components/schemas/HighlightField'
+            $ref: '#/components/schemas/HighlightFieldsArrayItem'
+          x-protobuf-ignore: true
```
karenyrx: adding x-protobuf-ignore here for now, as we figure out how this can be supported in protobufs via tooling

spec/schemas/_core.search.yaml
```diff
@@ -809,11 +809,26 @@ components:
...
+          items:
+            $ref: '#/components/schemas/HighlightFieldsArrayItem'
+          x-protobuf-ignore: true
+    HighlightFieldsArrayItem:
+      - type: object
+        additionalProperties:
+          title: key
+          $ref: '#/components/schemas/HighlightField'
+        minProperties: 1
+        maxProperties: 1
```
karenyrx: move under `items`

Issue #998, opened 2025-11-28
[FEATURE]


sam-briggs-depop:
### Is your feature request related to a problem?

Yes. The OpenSearch Search API response for `matched_queries` is polymorphic, but the current specification defines it strictly as an array of strings.

When the `include_named_queries_score` parameter is set to `true`, the API returns `matched_queries` as an object of `string -> number`, where the value represents the score.
When the parameter is `false` (default), it returns an array of `string`.

Currently, `spec/schemas/_core.search.yaml` defines it solely as:

```yaml
matched_queries:
  type: array
  items:
    type: string
```

This strict definition prevents generated clients (like `opensearch-java`) from correctly generating code to handle the response when scores are requested, leading to deserialization failures or the inability to implement the feature in the client without manual overrides.

### What solution would you like?

I would like the specification for `matched_queries` to be updated to reflect the actual API behavior.

Since the specification supports OpenAPI 3.1 (which utilizes JSON Schema 2020-12), we should utilize **Union Types** (via `oneOf`) to define that this field can be either an Array or an Object.

I propose updating the schema in `spec/schemas/_core.search.yaml` to something similar to:

```yaml
matched_queries:
  oneOf:
    - type: array
      items:
        type: string
      description: "The list of named queries that matched (when include_named_queries_score is false)"
    - type: object
      additionalProperties:
        type: number
      description: "A map of named queries to their scores (when include_named_queries_score is true)"
```

### What alternatives have you considered?

**Manual Client Overrides**
We could leave the spec as-is and force client maintainers to manually patch the generated logic. However, this defeats the purpose of having a canonical API specification and increases the maintenance burden on all downstream clients (Java, Python, Rust, etc.).

**Loosening the type to `Any`**
We could define the field simply as an `object` (Any), but this removes type safety for the standard use case (List of Strings) and is bad practice for a strongly typed specification.

### Do you have any additional context?

*   **Current Definition:** [`spec/schemas/_core.search.yaml#L96`](https://github.com/opensearch-project/opensearch-api-specification/blob/61777cdbe5bc32204640e6e236daf7cf71aa871d/spec/schemas/_core.search.yaml#L96)
[...]

PR #1015, opened 2025-12-22
feature: Make matched_queries polymorphic in search response

sam-briggs-depop:
### Description

Add oneOf schema to matched_queries field in Hit schema to support
both array and object return types based on the include_named_queries_score
parameter.

When include_named_queries_score is false (default), matched_queries
returns an array of query names. When true, it returns an object mapping
query names to their scores.

This change improves API accuracy and enables proper client generation
for the polymorphic response behavior that already exists in OpenSearch.

Changes:
- Update matched_queries in spec/schemas/_core.search.yaml to use oneOf
- Add descriptions for both array (names) and object (scores) variants
- Use double format for score values to match _score field convention

### Issues Resolved
Closes https://github.com/opensearch-project/opensearch-api-specification/issues/998


lucy66hw:
@sam-briggs-depop The workflows are all passed now. Thanks for the PR..

PR #1040, opened 2026-02-16
Add missing ShardSearchFailure

karenyrx:
### Description
This fix resolves the issue where the OpenSearch Java client was failing with MissingRequiredPropertyException when deserializing search responses. The spec now correctly distinguishes between:
- Search failures (`ShardSearchFailure`) - no `primary` field in  [server code](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/action/search/ShardSearchFailure.java#L142-L157)
- Bulk/write failures (`ShardFailure`) - has required `primary` field in [server code](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/action/support/replication/ReplicationResponse.java#L304-L317)
 
### Issues Resolved
https://github.com/opensearch-project/opensearch-java/issues/1877

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -27,7 +27,7 @@
         "ajv-errors": "^3.0.0",
         "ajv-formats": "^3.0.1",
         "aws4-axios": "^3.3.7",
-        "axios": "^1.7.5",
+        "axios": "^1.13.5",
         "axios-mock-adapter": "^2.0.0",
         "cbor": "^9.0.2",
         "commander": "^12.0.0",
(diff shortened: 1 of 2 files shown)


karenyrx:
Proto compatibility check awaiting fix in https://github.com/opensearch-project/opensearch-protobufs/pull/388 before being rerun

karenyrx:
The failing CI is not related to this PR, as it is also failing on main branch: https://github.com/opensearch-project/opensearch-api-specification/actions/runs/22080607191/job/63804886957

Will debug separately

janhoy:
Question @karenyrx . The CHANGELOG entry for this change was added under the section  `## [0.2.0] - 2025-05-25`. Looks wrong, as that was released in may last year? What will be the next version of the API-spec, and what version will v3.7.0 of opensearch-java use, vs what version will v4.0.0 eventually use? Do we need any backport somewhere to make sure this lands in 3.7.0?

Issue #1094, opened 2026-04-16
[FEATURE] Add support for terms lookup by query feature from OpenSearch 3.2.0


kieran-pringle:
### Is your feature request related to a problem?
When trying to use the latest Java client, I found that the builders do not allow setting a query, despite this query working in raw JSON in my local OpenSearch instance:
https://docs.opensearch.org/latest/query-dsl/term/terms/#terms-lookup-by-query

### What solution would you like?
To add support for this part of this API

### What alternatives have you considered?
Continuing to use raw JSON :( 

### Do you have any additional context?
N/A

andrross:
[Catch All Triage](https://github.com/opensearch-project/.github/blob/main/TRIAGING.md) - [1](https://github.com/krisfreedain) [2](https://github.com/andrross)

kieran-pringle:
@andrross - I followed the link above and am not sure if there are more steps expected from me at this time? I do understand that this project is run on volunteer time, so if there is any time that I can volunteer to make it easier for the team, please let me know. 

The PR to resolve this issue (as best as I can tell as a first time contributor!) is here: https://github.com/opensearch-project/opensearch-api-specification/pull/1095

PR #1095, opened 2026-04-16
Add query to TermsLookup to support terms lookup by query feature

kieran-pringle:
### Description
This adds support for the terms lookup by query feature from OpenSearch 3.2.0

### Issues Resolved
https://github.com/opensearch-project/opensearch-api-specification/issues/1094

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Add title to `ml.predict_model_stream` and `ml.execute_agent_stream` requestBody ([#1072](https://github.com/opensearch-project/opensearch-api-specification/pull/1072))
 - Added `agentic` query type, `agentic_query_translator` request processor, and `agentic_context` response processor ([#1073](https://github.com/opensearch-project/opensearch-api-specification/pull/1073))
 - Added `allow_no_indices`, `ignore_unavailable`, and `ignore_throttled` query parameters to `create_pit` ([#1141](https://github.com/opensearch-project/opensearch-api-specification/pull/1141))
+- Added `query` to TermsLookup to support terms lookup by query
 
 ### Deprecated
 - Marked the plural `_aliases` URL forms of `put_alias` and `delete_alias` as deprecated; the singular `_alias` form is the canonical path ([#1131](https://github.com/opensearch-project/opensearch-api-specification/pull/1131))
--- a/spec/schemas/_common.query_dsl.yaml
+++ b/spec/schemas/_common.query_dsl.yaml
@@ -2166,6 +2166,8 @@ components:
           $ref: '_common.yaml#/components/schemas/Field'
         routing:
           $ref: '_common.yaml#/components/schemas/Routing'
+        query:
+          $ref: '#/components/schemas/QueryContainer'
         store:
           type: boolean
       required:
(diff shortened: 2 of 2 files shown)


kieran-pringle:
I have rebased to deal with the issues flagged by Mend

sean-:
Thanks for adding this! The `query` support looks right. I verified against the server source and have two suggestions before this merges.

### 1. `query` needs `x-version-added: 3.2.0`

The `query` field is version-guarded on the transport wire in [`TermsLookup.java`](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/indices/TermsLookup.java) — both read and write are gated on `Version.V_3_2_0`:

```java
if (in.getVersion().onOrAfter(Version.V_3_2_0)) {
    query = in.readOptionalWriteable(inStream -> inStream.readNamedWriteable(QueryBuilder.class));
}
```

`V_3_2_0` resolves to `3.2.0` in `Version.java`, which matches the linked issue. Without the annotation, the field leaks into specs extracted for older server versions. This follows the existing convention in this file (e.g. the `filter` property carries `x-version-added: '3.0'`).

```diff
         query:
           $ref: '#/components/schemas/QueryContainer'
+          x-version-added: 3.2.0
         store:
           type: boolean
```

### 2. `id` and `query` are mutually exclusive

The server requires **exactly one** of `id` or `query` — never both, never neither (`TermsLookup.java`, the `ConstructingObjectParser` validation):

```java
if (id == null && query == null) {
    throw new IllegalArgumentException("... requires specifying either the id or the query.");
}
if (id != null && query != null) {
    throw new IllegalArgumentException("... cannot specify both id and query.");
}
```

The current schema lists `id` as unconditionally `required`, which is now wrong for the query-lookup path. A `oneOf` over the two required fields captures the constraint. Note the `query` branch is itself gated with `x-version-added: 3.2.0` so it's pruned for pre-3.2.0 targets — otherwise the extracted older spec would require a `query` field it no longer defines.

```diff
       required:
-        - id
         - index
         - path
+      oneOf:
+        - required:
+            - id
+        - required:
+            - query
+          x-version-added: 3.2.0
```

I confirmed both behave correctly by running `merge --opensearch-version 3.1.0` (collapses to `oneOf: [- required: [id]]`, no `query`) vs `3.2.0` (full constraint), and `lint:spec` passes.

kieran-pringle:
Thanks for the comments @sean- . I have updated the PR to reflect your feedback

kieran-pringle:
Test fixed and rebased on main

karenyrx:
Please add a changelog @kieran-pringle

kieran-pringle:
Thanks @karenyrx - I have added a changelog and made changes so the tests against older versions pass

kieran-pringle:
The tests for `2.0.0` failed with this, which I believe is unrelated?
```
/home/runner/work/opensearch-api-specification/opensearch-api-specification/node_modules/axios/lib/core/AxiosError.js:77
    const axiosError = new AxiosError(error.message, code || error.code, config, request, response);
```
It looks like the entire suite failed with a transient issue.

The proto compatibility check CI step seems misconfigured as it returns this error:
```
Error: The actions actions/checkout@v4, actions/setup-node@v3, actions/setup-java@v3, actions/download-artifact@v4, and actions/github-script@v6 are not allowed in opensearch-project/opensearch-api-specification because all actions must be pinned to a full-length commit SHA.
```

karenyrx (approved):

acote-coveo:
@karenyrx I don't know if you were aware but it looks like this commit failed on the main branch CI. Just wanted to let you know in case there's some follow up needed! 

Thanks

karenyrx:
@acote-coveo Yes the CI on main has unfortunately been failing for other commits as well.  https://github.com/opensearch-project/opensearch-api-specification/commits/main/  and is unrelated to this PR. These shall be fixed separately

kieran-pringle:
Thanks for all the help @karenyrx !

PR #1099, opened 2026-05-02
Fix ltr CacheStatsResponse to use NodesResponseBase via allOf

sean-:
`CacheStatsResponse` is served by [`RestFeatureStoreCaches.getStats`](https://github.com/opensearch-project/opensearch-learning-to-rank-base/blob/4318b045a5466fa563df336edacfe18c97a07933/src/main/java/com/o19s/es/ltr/rest/RestFeatureStoreCaches.java#L80-L82), which wraps the response with [`NodesResponseRestListener`](https://github.com/opensearch-project/OpenSearch/blob/27333365da0d55a16bea2ebaf34f6f2a691f6d50/server/src/main/java/org/opensearch/rest/action/RestActions.java#L267-L278). That listener calls [`RestActions.nodesResponse`](https://github.com/opensearch-project/OpenSearch/blob/27333365da0d55a16bea2ebaf34f6f2a691f6d50/server/src/main/java/org/opensearch/rest/action/RestActions.java#L215-L227), which renders the [`_nodes` header via `buildNodesHeader`](https://github.com/opensearch-project/OpenSearch/blob/27333365da0d55a16bea2ebaf34f6f2a691f6d50/server/src/main/java/org/opensearch/rest/action/RestActions.java#L171-L179) and then delegates body rendering to [`CachesStatsNodesResponse.toXContent`](https://github.com/opensearch-project/opensearch-learning-to-rank-base/blob/4318b045a5466fa563df336edacfe18c97a07933/src/main/java/com/o19s/es/ltr/action/CachesStatsAction.java#L98-L115) (which itself extends [`BaseNodesResponse`](https://github.com/opensearch-project/opensearch-learning-to-rank-base/blob/4318b045a5466fa563df336edacfe18c97a07933/src/main/java/com/o19s/es/ltr/action/CachesStatsAction.java#L60)).

The previous schema declared `_nodes` as an inline property alongside `nodes`, but the server renders `_nodes` outside the response class via the `NodesResponseRestListener` framework. Replacing the inline property with an `allOf` composition against `NodesResponseBase` makes the schema accurately mirror the server's class hierarchy. This also resolves a name collision (`_nodes` and `nodes` normalize to the same identifier in typed code generators), but that is a consequence of the correctness fix, not the motivation.

This pattern matches `ltr.Stats` (same file), `nodes.info`, `nodes.stats`, `nodes.usage`, `cluster.stats`, and 6 other schemas already in the spec.

Present since the LTR plugin's first OpenSearch release (1.0.0-alpha1).

### DIFF
--- a/spec/schemas/ltr._common.yaml
+++ b/spec/schemas/ltr._common.yaml
@@ -160,22 +160,22 @@ components:
           $ref: '#/components/schemas/NodeStatsDetails'
 
     CacheStatsResponse:
-      type: object
-      properties:
-        _nodes:
-          $ref: '_common.yaml#/components/schemas/NodeStatistics'
-        cluster_name:
-          $ref: '_common.yaml#/components/schemas/Name'
-        all:
-          $ref: '#/components/schemas/CacheAllStats'
-        stores:
-          type: object
-          description: Cache statistics by store.
-        nodes:
-          type: object
-          description: Cache statistics per node.
-          additionalProperties:
-            $ref: '#/components/schemas/NodeDetails'
+      allOf:
+        - $ref: 'nodes._common.yaml#/components/schemas/NodesResponseBase'
+        - type: object
+          properties:
+            cluster_name:
+              $ref: '_common.yaml#/components/schemas/Name'
+            all:
+              $ref: '#/components/schemas/CacheAllStats'
+            stores:
+              type: object
+              description: Cache statistics by store.
+            nodes:
+              type: object
+              description: Cache statistics per node.
+              additionalProperties:
+                $ref: '#/components/schemas/NodeDetails'
 
     NotFoundResponse:
       type: object


andrross (approved):

spec/schemas/ltr._common.yaml
```diff
@@ -160,22 +160,22 @@ components:
...
-        stores:
-          type: object
-          description: Cache statistics by store.
-        nodes:
-          type: object
-          description: Cache statistics per node.
-          additionalProperties:
-            $ref: '#/components/schemas/NodeDetails'
+      allOf:
+        - $ref: 'nodes._common.yaml#/components/schemas/NodesResponseBase'
```
karenyrx, line 164: This PR seems like we're modifying the upstream source of truth (the spec) to cater to a specific downstream consumer (language client generators). Long term, it should be the client generators that handle this

sean-: This isn't a generator accommodation — it's a correctness fix to make the schema match the server's actual class hierarchy.

The server doesn't render `_nodes` inside [`CachesStatsNodesResponse.toXContent`](https://github.com/opensearch-project/opensearch-learning-to-rank-base/blob/4318b045a5466fa563df336edacfe18c97a07933/src/main/java/com/o19s/es/ltr/action/CachesStatsAction.java#L98-L115) at all. Instead, [`RestFeatureStoreCaches.getStats`](https://github.com/opensearch-project/opensearch-learning-to-rank-base/blob/4318b045a5466fa563df336edacfe18c97a07933/src/main/java/com/o19s/es/ltr/rest/RestFeatureStoreCaches.java#L80-L82) wraps the response with [`NodesResponseRestListener`](https://github.com/opensearch-project/OpenSearch/blob/27333365da0d55a16bea2ebaf34f6f2a691f6d50/server/src/main/java/org/opensearch/rest/action/RestActions.java#L267-L278), which calls [`RestActions.nodesResponse`](https://github.com/opensearch-project/OpenSearch/blob/27333365da0d55a16bea2ebaf34f6f2a691f6d50/server/src/main/java/org/opensearch/rest/action/RestActions.java#L215-L227) — that's where `_nodes` and `cluster_name` are rendered via [`buildNodesHeader`](https://github.com/opensearch-project/OpenSearch/blob/27333365da0d55a16bea2ebaf34f6f2a691f6d50/server/src/main/java/org/opensearch/rest/action/RestActions.java#L171-L179), *before* delegating to `toXContent` for the body fields (`all`, `stores`, `nodes`).

The previous schema declared `_nodes` as a direct inline property, which doesn't match this inheritance. [`CachesStatsNodesResponse` extends `BaseNodesResponse`](https://github.com/opensearch-project/opensearch-learning-to-rank-base/blob/4318b045a5466fa563df336edacfe18c97a07933/src/main/java/com/o19s/es/ltr/action/CachesStatsAction.java#L60), so using `allOf` with `NodesResponseBase` makes the schema structurally mirror the server's class hierarchy. This is the same pattern already used by `Stats` in this same file, plus `nodes.info`, `nodes.stats`, `cluster.stats`, and others.

The generator name collision (`_nodes`/`nodes` → same identifier) exposed the discrepancy, but the fix is making the spec match the server — not working around a generator limitation. This has been the server's structure since the LTR plugin's first OpenSearch release (`1.0.0-alpha1`).

Xtansia (approved):

PR #1101, opened 2026-05-05
Add missing properties to nodes.info settings schemas

sean-:
[`NodesInfoResponse.toXContent`](https://github.com/opensearch-project/OpenSearch/blob/27333365da0d55a16bea2ebaf34f6f2a691f6d50/server/src/main/java/org/opensearch/action/admin/cluster/node/info/NodesInfoResponse.java#L119-L121) serializes node settings via raw [`Settings.toXContent`](https://github.com/opensearch-project/OpenSearch/blob/27333365da0d55a16bea2ebaf34f6f2a691f6d50/server/src/main/java/org/opensearch/common/settings/Settings.java#L624-L635), which preserves the internal storage format. The following settings were missing from the spec:

- **`node.roles`** ([`NODE_ROLES_SETTING`](https://github.com/opensearch-project/OpenSearch/blob/27333365da0d55a16bea2ebaf34f6f2a691f6d50/server/src/main/java/org/opensearch/node/NodeRoleSettings.java#L50-L51)): `Setting<List<DiscoveryNodeRole>>`, serialized as a JSON array when configured via YAML list, or a comma-separated string when set via `-E`, env var, or flat YAML scalar.
- **`http.host`**, **`http.publish_host`** ([`SETTING_HTTP_HOST`, `SETTING_HTTP_PUBLISH_HOST`](https://github.com/opensearch-project/OpenSearch/blob/27333365da0d55a16bea2ebaf34f6f2a691f6d50/server/src/main/java/org/opensearch/http/HttpTransportSettings.java#L94-L102)): `Setting<List<String>>`, same polymorphic serialization.
- **`http.publish_port`** ([`SETTING_HTTP_PUBLISH_PORT`](https://github.com/opensearch-project/OpenSearch/blob/27333365da0d55a16bea2ebaf34f6f2a691f6d50/server/src/main/java/org/opensearch/http/HttpTransportSettings.java#L119)): `Setting<Integer>`.
- **`transport.host`** ([`HOST`](https://github.com/opensearch-project/OpenSearch/blob/27333365da0d55a16bea2ebaf34f6f2a691f6d50/server/src/main/java/org/opensearch/transport/TransportSettings.java#L62-L67)): `Setting<List<String>>`, same polymorphic serialization.

The string-or-array polymorphism occurs because [`Settings.fromXContent`](https://github.com/opensearch-project/OpenSearch/blob/27333365da0d55a16bea2ebaf34f6f2a691f6d50/server/src/main/java/org/opensearch/common/settings/Settings.java#L692-L717) is format-preserving: YAML arrays hit the [`START_ARRAY` branch](https://github.com/opensearch-project/OpenSearch/blob/27333365da0d55a16bea2ebaf34f6f2a691f6d50/server/src/main/java/org/opensearch/common/settings/Settings.java#L692) (stored as `List`, serialized as JSON array), while flat scalars hit the [`VALUE_STRING` branch](https://github.com/opensearch-project/OpenSearch/blob/27333365da0d55a16bea2ebaf34f6f2a691f6d50/server/src/main/java/org/opensearch/common/settings/Settings.java#L712) (stored as `String`, serialized as JSON string). Uses the existing `StringOrStringArray` ref per reviewer suggestion.

All settings present since OpenSearch 1.0.0.

### DIFF
--- a/spec/schemas/nodes.info.yaml
+++ b/spec/schemas/nodes.info.yaml
@@ -456,6 +456,8 @@ components:
         max_local_storage_nodes:
           type: string
           description: The maximum number of nodes allowed to store data locally
+        roles:
+          $ref: '_common.yaml#/components/schemas/StringOrStringArray'
       required:
         - name
     NodeInfoPath:
@@ -539,6 +541,12 @@ components:
           $ref: '_common.yaml#/components/schemas/StringifiedBoolean'
         port:
           $ref: '_common.yaml#/components/schemas/StringifiedInteger'
+        host:
+          $ref: '_common.yaml#/components/schemas/StringOrStringArray'
+        publish_host:
+          $ref: '_common.yaml#/components/schemas/StringOrStringArray'
+        publish_port:
+          $ref: '_common.yaml#/components/schemas/StringifiedInteger'
       required:
         - type
     NodeInfoSettingsHttpType:
@@ -573,6 +581,8 @@ components:
         type.default:
           type: string
           description: The default transport type.
+        host:
+          $ref: '_common.yaml#/components/schemas/StringOrStringArray'
       required:
         - type
     NodeInfoSettingsTransportType:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ### Deprecated
 - Marked the plural `_aliases` URL forms of `put_alias` and `delete_alias` as deprecated; the singular `_alias` form is the canonical path ([#1131](https://github.com/opensearch-project/opensearch-api-specification/pull/1131))
+- Add `host`, `publish_host`, and `publish_port` to `NodeInfoSettingsHttp`; `host` to `NodeInfoSettingsTransport`; `roles` to `NodeInfoSettingsNode`; and updated `NodeInfoSettingsNetwork.host` to `StringOrStringArray` ([#1101](https://github.com/opensearch-project/opensearch-api-specification/pull/1101))
 
 ### Removed
 - Remove unused cardinality aggregation execution hints - save_memory_heuristic/save_time_heuristic/segment_ordinals ([#970](https://github.com/opensearch-project/opensearch-api-specification/pull/970))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Add msearch `allow_partial_results` flag ([#1061](https://github.com/opensearch-project/opensearch-api-specification/pull/1061))
 - Add title to `ml.predict_model_stream` and `ml.execute_agent_stream` requestBody ([#1072](https://github.com/opensearch-project/opensearch-api-specification/pull/1072))
 - Added `agentic` query type, `agentic_query_translator` request processor, and `agentic_context` response processor ([#1073](https://github.com/opensearch-project/opensearch-api-specification/pull/1073))
+- Add `host`, `publish_host`, and `publish_port` to `NodeInfoSettingsHttp`; `host` to `NodeInfoSettingsTransport`; `roles` to `NodeInfoSettingsNode`; and updated `NodeInfoSettingsNetwork.host` to `StringOrStringArray` ([#1101](https://github.com/opensearch-project/opensearch-api-specification/pull/1101))
 
 ### Removed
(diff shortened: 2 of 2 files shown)


spec/schemas/nodes.info.yaml
```diff
@@ -456,6 +456,9 @@ components:
         max_local_storage_nodes:
           type: string
           description: The maximum number of nodes allowed to store data locally
+        roles:
+          type: string
+          description: A comma-separated list of roles assigned to the node.
```
andrross: Unfortunately I think we've got some weird behavior here and this might need to be something like the following:


```suggestion
        roles:
          description: |-
            The roles assigned to the node. Serialized as a comma-separated
            string when configured via a flat setting (env var, CLI `-E`,
            or flat YAML scalar), or as an array of strings when configured
            as a YAML list in `opensearch.yml`.
          oneOf:
            - type: string
            - type: array
              items:
                type: string
```

sean-: Confirmed by reading the server source. `Settings.toXContent()` serializes the raw internal map value.  When `node.roles` is loaded from a YAML list, `fromXContent` calls `builder.putList()` (stores as `List<String>`, serialized as a JSON array), but when set via `-E`, env var, or a flat YAML scalar, `putProperties`/`put(key, value)` stores it as a plain `String` (serialized as a comma-separated JSON string).

Applied your suggestion.  Thanks for catching this.

spec/schemas/nodes.info.yaml
```diff
@@ -539,6 +550,14 @@ components:
           $ref: '_common.yaml#/components/schemas/StringifiedBoolean'
         port:
           $ref: '_common.yaml#/components/schemas/StringifiedInteger'
+        host:
+          type: string
+          description: The bound host address for the HTTP server.
+        publish_host:
+          type: string
+          description: The host address published to other nodes for HTTP communication.
```
msfroh: I could be mistaken, but the occurrences of these settings in the code look like they're both `List<String>`. (In practice, I think the lists almost always have size 1, though.)

sean-: Oh!, interesting... yeah, you're right.  In my local setup when developing this, I was using a string, not an array.  Wild, thank you for pointing this out.

It turns out these settings are `Setting<List<String>>`, but the actual serialized response is **polymorphic** because the field can be either a string or an array depending on how the operator configured the setting.  Yuk.

The root cause is that `NodeInfo` serializes settings via the raw [`Settings.toXContent()`](https://github.com/opensearch-project/OpenSearch/blob/a70e8ee061c4158a43b7f31d62589952dd425627/server/src/main/java/org/opensearch/common/settings/Settings.java#L624-L635), which preserves the internal storage format rather than normalizing through the `Setting<List<String>>` type. [`Settings.fromXContent()`](https://github.com/opensearch-project/OpenSearch/blob/a70e8ee061c4158a43b7f31d62589952dd425627/server/src/main/java/org/opensearch/common/settings/Settings.java#L692-L717) is format-preserving:

- YAML array (`http.host: ["0.0.0.0"]`) hits the [`START_ARRAY` branch](https://github.com/opensearch-project/OpenSearch/blob/a70e8ee061c4158a43b7f31d62589952dd425627/server/src/main/java/org/opensearch/common/settings/Settings.java#L692-L707) calls `putList()`, which calls stored as `ArrayList`, which serializes as a JSON array
- YAML scalar / CLI `-E` / env var (`http.host: "0.0.0.0"`) hits the [`VALUE_STRING` branch](https://github.com/opensearch-project/OpenSearch/blob/a70e8ee061c4158a43b7f31d62589952dd425627/server/src/main/java/org/opensearch/common/settings/Settings.java#L712-L717) which calls `put()` and that stores the value as a `String`, which then gets serialized out as a JSON string

So despite the Java type being `List<String>`, the wire format depends on how `opensearch.yml` was written, meaning every API client has to handle both forms.  This is pretty gross, but here we are.

I've updated the spec to use `oneOf: [string, array]` for `http.host`, `http.publish_host`, `transport.host`, and `network.host` (same pattern as `settings.node.roles` in this PR).

spec/schemas/nodes.info.yaml
```diff
@@ -456,6 +456,17 @@ components:
         max_local_storage_nodes:
           type: string
           description: The maximum number of nodes allowed to store data locally
+        roles:
+          description: |-
+            The roles assigned to the node. Serialized as a comma-separated
+            string when configured via a flat setting (env var, CLI `-E`,
+            or flat YAML scalar), or as an array of strings when configured
+            as a YAML list in `opensearch.yml`.
+          oneOf:
```
Xtansia: You can use `$ref: '_common.yaml#/components/schemas/StringOrStringArray'` for these

PR #1102, opened 2026-05-06
Mark create_pit keep_alive query parameter as required

sean-:
The [`RestCreatePitAction`](https://github.com/opensearch-project/OpenSearch/blob/27333365da0d55a16bea2ebaf34f6f2a691f6d50/server/src/main/java/org/opensearch/rest/action/search/RestCreatePitAction.java#L43) parses `keep_alive` with a `null` default, passing it directly to `CreatePitRequest`. [`CreatePitRequest.validate()`](https://github.com/opensearch-project/OpenSearch/blob/27333365da0d55a16bea2ebaf34f6f2a691f6d50/server/src/main/java/org/opensearch/action/search/CreatePitRequest.java#L127-L133) then rejects the request with HTTP 400 ("keep alive not specified") when the value is null.

The validation has been present since the PIT API was first introduced (PR #2745 / #3921), shipped in OpenSearch 2.4.0 (2.x backport via #4616). All OpenSearch versions that support `create_pit` require `keep_alive`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Removed unused dependencies: `eslint-config-standard-with-typescript`, `eslint-plugin-import`, `eslint-plugin-n`, `eslint-plugin-promise`, `@eslint/eslintrc`
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


andrross (approved):

karenyrx (dismissed):

Xtansia (approved):

Xtansia (changes requested):
@sean- Please add a changelog entry

Xtansia (approved):

PR #1103, opened 2026-05-07
Add max_score to HitsMetadata required fields

sean-:
Add `max_score` to the `required` list of `HitsMetadata`.

The OpenSearch server **unconditionally** serializes `max_score` in every search response — emitting `null` when the value is `NaN`, or the float value otherwise.  This behavior has been present since the initial 1.0 fork and has never changed, but the spec incorrectly marked the field as optional.

**Server reference:** [`SearchHits.java#L247-L251`](https://github.com/opensearch-project/OpenSearch/blob/f1cbbba360e92384448104a14adcafc88ec6854b/server/src/main/java/org/opensearch/search/SearchHits.java#L247-L251)

```java
if (Float.isNaN(maxScore)) {
    builder.nullField(Fields.MAX_SCORE);
} else {
    builder.field(Fields.MAX_SCORE, maxScore);
}
```

The field was already declared as `nullable: true` with `type: number` in the spec but wasn't listed under `required`.

### DIFF
--- a/.github/workflows/analyze-pr-changes.yml
+++ b/.github/workflows/analyze-pr-changes.yml
@@ -144,8 +144,9 @@ jobs:
       - name: Generate API Changes Summary
         shell: bash -eo pipefail {0}
         run: |
-          if ! openapi-changes summary --no-logo --no-color --markdown $BEFORE_SPEC $AFTER_SPEC >output.md ; then
-            if ! grep -q 'breaking changes discovered' output.md ; then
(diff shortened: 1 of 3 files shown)


andrross (approved):

karenyrx (dismissed):

Xtansia (changes requested):
@sean- Please add a changelog entry

Xtansia (dismissed):

Xtansia (approved):

Xtansia (approved):

PR #1104, opened 2026-05-07
Add nested replication object to IndexSettings schema

sean-:

### DIFF
--- a/spec/schemas/indices._common.yaml
+++ b/spec/schemas/indices._common.yaml
@@ -158,11 +158,21 @@ components:
           $ref: '#/components/schemas/IndexSettingsStarTree'
         ingestion_source:
           $ref: '#/components/schemas/IngestionSource'
+        replication:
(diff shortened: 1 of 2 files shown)


spec/schemas/indices._common.yaml
```diff
@@ -158,11 +158,21 @@ components:
...
           $ref: '#/components/schemas/ReplicationType'
+          x-version-added: '1.2'
       additionalProperties:
         title: custom_settings
+    IndexSettingsReplication:
+      type: object
+      description: The replication settings for an index.
+      properties:
+        type:
+          $ref: '#/components/schemas/ReplicationType'
```
msfroh, line 175: Just for my own edification, does adding `replication` with `type` underneath it mean that it becomes possible to remove `replication.type` from line 164? 

I'm guessing something breaks if we remove it, but it feels like we *should* be able to remove it.

PR #1108, opened 2026-05-12
Add `node_id` property to `ErrorCause` schema (#1108)

sean-:
FailedNodeException serializes a `node_id` field via metadataToXContent that was missing from the spec. This surfaces the originating node in node-level failures such as `failed_node_exception`.


msfroh (commented):
It looks like there are several exception types that may add additional fields to a Protobuf response: https://github.com/opensearch-project/OpenSearch/blob/d866be82feeb9aab9cf399f64b25b82051275e20/modules/transport-grpc/src/main/java/org/opensearch/transport/grpc/proto/response/exceptions/opensearchexception/OpenSearchExceptionProtoUtils.java#L238-L250

For those, the Protobuf solution has been to model them under `ErrorCause`'s `metadata` property.

@sean-, are you working on a client that can just put this additional exception info under `metadata`? Or does it need to be modeled as a specific field in the `ErrorCause`?

sean-:
`node_id` is a top-level field in the JSON error response, not nested under `metadata`. The Protobuf `metadata` map is a gRPC-transport-only representation that does not participate in REST serialization. These are two independent pipelines.

To answer your question directly: yes, it needs to be a specific field in `ErrorCause`. I'm authoring a Go client
([opensearch-go#650](https://github.com/opensearch-project/opensearch-go/pull/650)) with a code generator that produces typed structs from the spec. Without `node_id` on `ErrorCause` in the spec, the generated struct has no field for it and the value is silently dropped during JSON deserialization. There's no `metadata` bag in the JSON wire format to fall back on.

The [`metadataToProto()` dispatch](https://github.com/opensearch-project/OpenSearch/blob/d866be82feeb9aab9cf399f64b25b82051275e20/modules/transport-grpc/src/main/java/org/opensearch/transport/grpc/proto/response/exceptions/opensearchexception/OpenSearchExceptionProtoUtils.java#L238-L250) lives entirely in `modules/transport-grpc` (a [gRPC transport plugin](https://github.com/opensearch-project/OpenSearch/blob/d866be82feeb9aab9cf399f64b25b82051275e20/modules/transport-grpc/build.gradle#L13-L16)). It packs exception-specific fields into a proto `metadata` `ObjectMap` because proto lacks the polymorphic override mechanism, its own [Javadoc calls it "similar to `metadataToXContent`"](https://github.com/opensearch-project/OpenSearch/blob/d866be82feeb9aab9cf399f64b25b82051275e20/modules/transport-grpc/src/main/java/org/opensearch/transport/grpc/proto/response/exceptions/opensearchexception/OpenSearchExceptionProtoUtils.java#L232-L233) (i.e. a parallel implementation, not an intermediary).

[`BytesRestResponse.build()`](https://github.com/opensearch-project/OpenSearch/blob/d866be82feeb9aab9cf399f64b25b82051275e20/server/src/main/java/org/opensearch/rest/BytesRestResponse.java#L169-L175) calls `generateFailureXContent()`, which calls [`innerToXContent()`](https://github.com/opensearch-project/OpenSearch/blob/d866be82feeb9aab9cf399f64b25b82051275e20/libs/core/src/main/java/org/opensearch/OpenSearchException.java#L405-L414), which in turn calls `metadataToXContent()` on the same `XContentBuilder` at the same nesting level as `type` and `reason`. [`FailedNodeException` overrides this to write `node_id` directly](https://github.com/opensearch-project/OpenSearch/blob/d866be82feeb9aab9cf399f64b25b82051275e20/server/src/main/java/org/opensearch/action/FailedNodeException.java#L71-L74):

```json
{ "type": "failed_node_exception", "reason": "...", "node_id": "abc123", "caused_by": { ... } }
```

So, no, no Protobuf anywhere in that chain.

msfroh:
Got it -- adding these various fields to `ErrorCause` should also make it possible to model them directly in `ErrorCause` for the Protobuf transport as well (in future), so we won't need the metadata hack.

While `node_id` is needed for `FailedNodeException`, the other properties needed to cover all exception types that have `metadataToXContent` are:
* `CircuitBreakingException`
    * `bytes_wanted`
    * `bytes_limit`
    * `durability`
* `ParsingException` and `SearchParseException`
    * `line`
    * `col`
* `ResponseLimitBreachedException`
    * `response_limit`
    * `limit_entity`
* `ScriptException`
    * `script_stack`
    * `script`
    * `lang`
* `SearchPhaseExecutionException`
    * `phase`
    * `grouped`
    * `failed_shards`
* `TooManyBucketsException`
    * `max_buckets`

sean-:
Done. I've expanded this PR to cover all exception subclasses that override `metadataToXContent()`. Every field was verified against the server source.

 | Exception | Fields added |
 |-----------|-------------|
 | `FailedNodeException` | `node_id` (already present) |
 | `CircuitBreakingException` | `bytes_wanted` (int64), `bytes_limit` (int64), `durability` (enum: `TRANSIENT`, `PERMANENT`) |
 | `ParsingException` / `SearchParseException` | `line` (int32), `col` (int32) |
 | `ResponseLimitBreachedException` | `response_limit` (int32), `limit_entity` (enum: `INDICES`, `SHARDS`) — both marked `x-version-added: '3.0'` |
 | `ScriptException` | `script_stack` (string[]), `script` (string), `lang` (string), `position` (object: `offset`, `start`, `end`) |
 | `SearchPhaseExecutionException` | `phase` (string), `grouped` (boolean), `failed_shards` (array of `ShardFailure`) |
 | `TooManyBucketsException` | `max_buckets` (int32) |

 `position` wasn't in your list but is emitted by [`ScriptException.Position.toXContent()`](https://github.com/opensearch-project/OpenSearch/blob/d866be82feeb9aab9cf399f64b25b82051275e20/server/src/main/java/org/opensearch/script/ScriptException.java#L221-L227) when a script compilation error includes source location.

 `ResponseLimitBreachedException` was introduced in 3.0.0 ([#15986](https://github.com/opensearch-project/OpenSearch/pull/15986)); all other exceptions are present since 1.0.0.

msfroh (approved):

PR #1133, opened 2026-05-27
Define typed enum schemas for `format` query parameter on cat, list, sql, ppl

sean-:
Today every `*::query.format` query parameter on cat, list, ppl, and sql operations is declared as a free-form `type: string` with no `enum`. This forces every code-generator to either accept an open string and surface invalid values at runtime, or special-case format validation per operation. It also obscures that these four operation families accept different sets of values and have different server-side defaults.

Add four typed enum schemas under `_common` and replace 35 inline schemas with `$ref`s to them. Each schema is scoped per surface so it accurately describes what its server actually accepts:

- `CatResponseFormat`  — `[cbor, json, smile, text, yaml]`, default `text`
- `ListResponseFormat` — `[cbor, json, smile, text, yaml]`, default `text`
- `SQLResponseFormat`  — `[cost, csv, extended, jdbc, json, raw, simple, standard, viz, yaml]`, default `jdbc`
- `PPLResponseFormat`  — same value set as SQL, default `jdbc`

Source-of-truth references:

- Cat formats: [`RestTable.MediaType.fromFormat`](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/rest/action/cat/RestTable.java#L83) — the omit-format default routes through `MediaType` to `text/plain`
- SQL/PPL formats: [`Format.java`](https://github.com/opensearch-project/sql/blob/main/protocol/src/main/java/org/opensearch/sql/protocol/response/format/Format.java) enumerates `JDBC`, `CSV`, `RAW`, `VIZ`, `JSON`, `YAML`, `SIMPLE`, `STANDARD`, `EXTENDED`, `COST`
- SQL/PPL default: [`PPLQueryRequestFactory`](https://github.com/opensearch-project/sql/blob/main/plugin/src/main/java/org/opensearch/sql/plugin/request/PPLQueryRequestFactory.java#L28) and [`SqlRequestParam`](https://github.com/opensearch-project/sql/blob/main/legacy/src/main/java/org/opensearch/sql/legacy/request/SqlRequestParam.java#L18) both define `DEFAULT_RESPONSE_FORMAT = "jdbc"`

Concrete downstream impact: in `opensearch-project/opensearch-go`, the regenerated `osapi/` package emits `Format string` for cat operations with no default. Calling `client.Cat.Indices(ctx, nil)` produces a `text/plain` response (the cat server's default) which then fails JSON parsing. The hand-written predecessor hardcoded `format=json` per cat op; the regenerated client lost that behavior because the spec did not encode the enum or the default. With typed enums and explicit defaults, generators can emit a typed Go enum (or equivalent) and pick a sensible client-side default that matches the response struct shape.

### DIFF
--- a/spec/schemas/_common.yaml
+++ b/spec/schemas/_common.yaml
@@ -828,65 +828,6 @@ components:
     DateFormat:
       type: string
       description: The date format pattern.
-    CatResponseFormat:
-      type: string
-      description: |-
-        The response format returned by `_cat` operations. Equivalent to the HTTP `Accept` header but expressed as a query parameter.
-        When the parameter is omitted the server responds with `text`.
-      default: text
-      enum:
-        - cbor
-        - json
-        - smile
-        - text
-        - yaml
-    ListResponseFormat:
-      type: string
-      description: |-
-        The response format returned by `_list` operations. Equivalent to the HTTP `Accept` header but expressed as a query parameter.
-        When the parameter is omitted the server responds with `text`.
-      default: text
-      enum:
-        - cbor
-        - json
-        - smile
-        - text
-        - yaml
-    SQLResponseFormat:
-      type: string
-      description: |-
-        The response format selector for SQL plugin operations. The valid set is the union of values accepted across `_plugins/_sql/*` endpoints:
-        query responses use `jdbc` (default), `csv`, `raw`, or `viz`; explain responses use `json`, `yaml`, `simple`, `standard`, `extended`, or `cost`.
-        Not every endpoint accepts every value.
-      default: jdbc
-      enum:
-        - cost
-        - csv
-        - extended
-        - jdbc
-        - json
-        - raw
-        - simple
-        - standard
-        - viz
-        - yaml
-    PPLResponseFormat:
-      type: string
-      description: |-
-        The response format selector for PPL plugin operations. The valid set is the union of values accepted across `_plugins/_ppl/*` endpoints;
-        defaults and per-endpoint subsets mirror the SQL plugin.
-      default: jdbc
-      enum:
-        - cost
-        - csv
-        - extended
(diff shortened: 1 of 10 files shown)


spec/schemas/_common.yaml
```diff
@@ -828,6 +828,65 @@ components:
     DateFormat:
       type: string
       description: The date format pattern.
+    CatResponseFormat:
```
Xtansia: These shouldn't be in `_common.yaml`, they should be in the relevant `{namespace}._common.yaml` files

Xtansia (changes requested):

spec/schemas/_common.yaml
sean-: Done.  Pushed five commits on top of `6557ec13`:

- `18a4fb6d` Move `CatResponseFormat` into `cat._common.yaml` as `ResponseFormat`
- `a189b7d9` Move `ListResponseFormat` into `list._common.yaml` as `ResponseFormat`
- `6d28174d` Move `SQLResponseFormat` into `sql._common.yaml` (existing file) as `ResponseFormat`
- `23c6a4ec` Move `PPLResponseFormat` into `ppl._common.yaml` as `ResponseFormat`
- `c695829e` Drop the four `*ResponseFormat` schemas from `_common.yaml`

Dropped the redundant `Cat`/`List`/`SQL`/`PPL` prefix on each schema since the filename already scopes it (matches the existing convention in `sql._common.yaml`). All 35 `*::query.format` `$ref`s in the namespace files now point at `{cat,list,sql,ppl}._common.yaml#/components/schemas/ResponseFormat`. `npm run lint:spec` and `npm run merge` are clean at every commit.

Xtansia (approved):

PR #1137, opened 2026-05-29
Add x-error-responses extension for partial-failure response shapes

sean-:
Introduce a per-operation `x-error-responses` vendor extension whose value is an array of `$ref` objects pointing at wrapper schemas in `spec/schemas/_common.errors.yaml`. Each wrapper describes one auxiliary failure shape an endpoint may carry alongside its primary 2xx response — including arity (array vs object vs map) and any parallel-array grouping (e.g. `task_failures[]` + `node_failures[]` together for the Tasks API) — using ordinary JSON Schema. Client generators walk an operation's `x-error-responses` array, look up each wrapper, and emit one typed handler per wrapper that detects and extracts the embedded failure content.

The new `_common.errors.yaml` defines 15 wrappers covering the distinct partial-failure encodings present in OpenSearch today: BulkItems, SearchShards, WriteShards, BroadcastShards, NodeFailures, BulkByScrollFailures, TaskFailures, MultiSearchItems, MultiDocItems, SnapshotCreateShardFailures, SnapshotGetShardFailures, SimulateDocFailures, RankEvalFailures, IngestionShardFailures, and PitNodeFailures. Each wrapper composes existing element schemas (`ShardSearchFailure`, `ShardFailure`, `NodeStatistics`, `TaskFailure`, `SnapshotShardFailure`, `IngestionStateShardFailure`, `BulkByScrollFailure`, etc.) — no new element types are added.

Annotate 115 operation variants across 10 namespace files (_core, asynchronous_search, cluster, dangling_indices, indices, ingest, ingestion, nodes, snapshot, tasks). `indices.recovery` and `indices.get_upgrade` are intentionally not annotated: their response classes extend `BroadcastResponse` but suppress the `_shards` block in their `toXContent` overrides.

Document the extension in DEVELOPER_GUIDE.md alongside the existing vendor extensions list.

Motivated by `opensearch-go`'s port of partial-failure error handling into its generated `v5preview/opensearchapi/` package, which needs a generator-driven mapping from operation to auxiliary failure types.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ### Added
 
+- Added `x-error-responses` vendor extension and `_common.errors.yaml` wrapper schemas catalog for declaring auxiliary partial-failure response shapes per operation ([#1137](https://github.com/opensearch-project/opensearch-api-specification/pull/1137))
 - Added apis for searching within search relevance objects such as search configurations, judgments, query sets, and experiments ([#1064](https://github.com/opensearch-project/opensearch-api-specification/pull/1064))
 - Added remaining APIs for LTR including store element operations, model management, feature set operations, routing support, and POST support for update operations ([#935](https://github.com/opensearch-project/opensearch-api-specification/pull/935))
 - Added specs for alert and finding endpoints of security_analytics plugin ([#907](https://github.com/opensearch-project/opensearch-api-specification/pull/907))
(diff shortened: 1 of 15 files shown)


msfroh (approved):
@sean- I would accept the suggestions from the `vale` bot around adding backticks when referencing APIs. 

I'm personally less fussed about it's suggestions around not using "vs" or other word choices.

Also, while this gets things up-to-date (or at least up-to-early-June), I'm a little scared that this is going to be extremely brittle. Do we have a good way of automatically detecting if/when the server breaks this? I might add that to a to-do list for tying the API spec to the server implementations.

PR #1161, opened 2026-07-10
Allow indexed_shape-only geo_shape and xy_shape queries

### LINKED ISSUE #1160: [BUG] `shape` should not be required in `GeoShapeQueryField` (blocks pre-indexed shape queries in generated clients)
### What is the bug?

The `GeoShapeQueryField` schema marks `shape` as required, but the OpenSearch geo_shape query supports a **pre-indexed shape** form that uses `indexed_shape` and no sibling `shape`. Because this spec is the source of truth for generated clients, the requiredness propagates downstream and makes it impossible to build a pre-indexed shape query with the generated classes.

Current schema (`spec/schemas/_common.query_dsl.yaml`, main):

```yaml
GeoShapeQueryField:
  type: object
  properties:
    indexed_shape:
      $ref: '#/components/schemas/FieldLookup'
    shape:
      $ref: '#/components/schemas/GeoShape'
    relation:
      $ref: '_common.yaml#/components/schemas/GeoShapeRelation'
  required:
    - shape
```

Per the docs, `shape` and `indexed_shape` are alternatives (one or the other, not both):
https://docs.opensearch.org/latest/query-dsl/geo-and-xy/geoshape/#using-a-pre-indexed-shape-definition

A pre-indexed shape query is valid against OpenSearch. This is the pre-indexed example from the docs (index `pre-indexed-shapes`, document id `search_triangle`, path `boundaries`):

```json
GET /testindex/_search
{
  "query": {
    "geo_shape": {
      "location": {
        "indexed_shape": {
          "index": "pre-indexed-shapes",
          "id": "search_triangle",
          "path": "boundaries"
        },
        "relation": "WITHIN"
      }
    }
  }
}
```

Validating this body against the current schema fails because `shape` is missing, even though the query has no inline shape by design.

### How did this happen? (timeline)

- PR #531 introduced the geo shape field schema (then named `GeoShapeField`) with `required: - shape`. At that time the schema had only `shape` and `relation`, with no `indexed_shape`, so requiring `shape` was consistent.
- `indexed_shape` was added to the schema later, but `required: - shape` was not revisited. That produced the current mismatch: `indexed_shape` is optional while `shape` stays mandatory.

(Origin PR: https://github.com/opensearch-project/opensearch-api-specification/pull/531)

### Downstream impact

Reported in the Java client: https://github.com/opensearch-project/opensearch-java/issues/2011

Maintainer confirmation on that issue (xluo-aws):
- either `shape` or `indexed_shape` can be used, but not both;
- the fix should be made in the API spec first, otherwise the change in the client repo is overwritten by regeneration.

Generated client repro (opensearch-java):

```java
new Query.Builder()
    .geoShape(new GeoShapeQuery.Builder()
        .field("some")
        .shape(shape -> shape.indexedShape(i -> i.index("some").id("id").path("path")))
        .build())
    .build();
// GeoShapeQueryField requires `shape`, so an indexed_shape-only query can't be built.
```

The schema should allow a `GeoShapeQueryField` that provides `indexed_shape` without `shape`, matching the documented pre-indexed shape query.

### Open questions for maintainers (how to model the contract)

I'd like to align on the intended schema shape before opening a PR:

1. Is the contract "exactly one of `shape` / `indexed_shape`" (mutually exclusive), or "at least one of them"? The docs read as one-or-the-other.
2. How should this be expressed in the spec: simply dropping `required: - shape`, or using `oneOf` / `minProperties` / another construct that the codegen and existing spec conventions support?
3. `XyShapeQueryField` has the same issue. It carries an identical `required: - shape` alongside an optional `indexed_shape`, so it likely needs the same treatment (it was introduced as `XyShapeField` in #531 and later renamed, same as the geo variant).
4. Which spec tests should be updated or added? There are existing tests under `tests/default/_core/search/query/geo_shape_*.yaml` from #531, but none cover the `indexed_shape` path.

### Additional context

- Docs source: https://github.com/opensearch-project/documentation-website/blob/main/_query-dsl/geo-and-xy/geoshape.md
- Current schema: https://github.com/opensearch-project/opensearch-api-specification/blob/main/spec/schemas/_common.query_dsl.yaml

Tejashribambal19:
## Summary

Fixes #1160.

The current API specification marks `shape` as required in both `GeoShapeQueryField` and `XyShapeQueryField`. However, OpenSearch supports pre-indexed shape queries using `indexed_shape` without an accompanying `shape`, as documented.

Because this specification is the source of truth for generated clients, the current requirement prevents clients from constructing valid `indexed_shape`-only queries.

This PR removes the `shape` requirement from both schemas, allowing both inline `shape` queries and pre-indexed `indexed_shape` queries.

## Changes

- Remove `required: [shape]` from `GeoShapeQueryField`
- Remove `required: [shape]` from `XyShapeQueryField`

## Motivation

The OpenSearch documentation describes `shape` and `indexed_shape` as alternative ways to define a geo/xy shape query. Requiring `shape` makes valid pre-indexed shape queries fail schema validation and propagates incorrect constraints to generated clients.

This change aligns the API specification with the documented query behavior.

## Testing

- Verified that the schema changes are limited to `GeoShapeQueryField` and `XyShapeQueryField`.
- Existing query behavior remains unchanged for inline `shape` queries.

## Notes

The current specification tests cover inline shape queries but do not include coverage for the indexed_shape path. This PR focuses on correcting the schema; regression tests for indexed_shape can be added separately if maintainers prefer additional coverage.

### DIFF
--- a/spec/schemas/_common.query_dsl.yaml
+++ b/spec/schemas/_common.query_dsl.yaml
@@ -806,8 +806,7 @@ components:
           $ref: '#/components/schemas/GeoShape'
         relation:
           $ref: '_common.yaml#/components/schemas/GeoShapeRelation'
-      required:
-        - shape
+      
     GeoShapeQuery:
       allOf:
         - $ref: '#/components/schemas/QueryBase'
@@ -2260,8 +2259,7 @@ components:
           $ref: '#/components/schemas/XyShape'
         relation:
           $ref: '_common.yaml#/components/schemas/GeoShapeRelation'
-      required:
-        - shape
+      
     XyShapeQuery:
       x-version-added: '2.4'
       allOf:


spec/schemas/_common.query_dsl.yaml
```diff
@@ -806,8 +806,7 @@ components:
           $ref: '#/components/schemas/GeoShape'
         relation:
           $ref: '_common.yaml#/components/schemas/GeoShapeRelation'
-      required:
```
lsh1215, line 809: Dropping `required` unblocks `indexed_shape`-only, but it also makes **both** `shape`+`indexed_shape` and **neither** valid. Per the docs and @xluo-aws (opensearch-java#2011), it should be exactly one — already modeled in this same file by `TermsLookup`:

```yaml
allOf:
  - oneOf:
      - type: object
        properties:
          indexed_shape: { $ref: '#/components/schemas/FieldLookup' }
        required: [indexed_shape]
      - type: object
        properties:
          shape: { $ref: '#/components/schemas/GeoShape' }
        required: [shape]
```

spec/schemas/_common.query_dsl.yaml
```diff
@@ -2260,8 +2259,7 @@ components:
           $ref: '#/components/schemas/XyShape'
         relation:
           $ref: '_common.yaml#/components/schemas/GeoShapeRelation'
-      required:
```
lsh1215, line 2263: Same here — `XyShapeQueryField` needs the same exactly-one `allOf`/`oneOf` treatment (with `XyShape` instead of `GeoShape`).

lsh1215 (commented):
Thanks for picking this up, @Tejashribambal19 — I opened the linked issue #1160. This unblocks the immediate case; I left two inline notes suggesting we model *exactly one* of `shape`/`indexed_shape` (via the existing `TermsLookup` `oneOf` pattern) instead of dropping `required`, since the contract is "one or the other, not both."

One more gap: there's no test coverage for the `indexed_shape` path yet — the existing `geo_shape_*` / `xy_shape_*` stories only exercise inline `shape` (open question #4 in the issue). I have a branch with the `oneOf` modeling plus `geo_shape_indexed_shape` / `xy_shape_indexed_shape` stories, validated against the merged schema (`lint:spec` / `merge` / unit all green). Happy to fold those into this PR or open an alternative — whichever you and the maintainers prefer.

PR #944, opened 2025-07-23
Add `false_allow_templates` as a dynamic mapping opton

bruce-hong-glean:
### Description
Add `false_allow_templates` as a dynamic mapping option

### Issues Resolved
https://github.com/opensearch-project/OpenSearch/issues/18617#issuecomment-3070419493

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #1178, opened 2026-07-23
Add false_allow_templates to DynamicMapping

Truta446:
### Description

Adds the `false_allow_templates` value to the `DynamicMapping` schema.

OpenSearch's `dynamic` mapping parameter accepts `false_allow_templates` (new fields are only indexed when they match a dynamic template, otherwise they are not indexed), but the spec's `DynamicMapping` enum was missing it — it only listed `false`, `strict`, `strict_allow_templates`, `true`, and boolean. As a result, generated clients reject the value at the type level even though it works at runtime.

The value was introduced in OpenSearch 3.3.0 (opensearch-project/OpenSearch#19065), so the added integration test case is gated with `version: '>= 3.3'`, mirroring the existing `strict_allow_templates` case.

Reported downstream in the JS client: .

### Issues Resolved

Fixes the missing `false_allow_templates` value in `DynamicMapping`.

### Changes

- `spec/schemas/_common.mapping.yaml`: add `false_allow_templates` to the `DynamicMapping` string enum.
- `tests/default/indices/mapping/mapping.yaml`: add an integration test setting `dynamic: false_allow_templates` (gated `>= 3.3`).
- `CHANGELOG.md`: changelog entry.

### Check List

- [x] All tests pass (`npm run lint:spec` passes; `npm run merge` succeeds and the bundled spec includes the value).
- [x] Updated the CHANGELOG.
- [ ] Commits are signed per the DCO using `--signoff`.


kanish671:
I just noticed that this is a duplicate of #944

PR #1192, opened 2026-08-15
remove incorrect model_id from SearchHitsHit

### LINKED ISSUE #1180: [BUG] `model_id` on `SearchHitsHit` in ML schema is incorrect — field never exists in API response
### Description

The `SearchHitsHit` schema in `spec/schemas/ml._common.yaml` declares a `model_id` property:

```yaml
SearchHitsHit:
  type: object
  properties:
    ...
    model_id:
      $ref: '_common.yaml#/components/schemas/Name'
    ...
```

However, the OpenSearch ML search APIs (`_plugins/_ml/models/_search`, `_plugins/_ml/agents/_search`, etc.) **never return `model_id` as a top-level field on search hits**. These APIs use standard OpenSearch `SearchResponse` — the document ID (which IS the model ID) is always in the standard `_id` field.

### Reproduction

```bash
# Search for ML models
curl -s http://localhost:9200/_plugins/_ml/models/_search \
  -H "Content-Type: application/json" \
  -d '{"query":{"match_all":{}},"size":1}'
```

Response hit structure:
```json
{
  "_index": ".plugins-ml-model",
  "_id": "6Wd_7p4BTcJJvPo5kauK",   // <-- this IS the model_id
  "_version": 1,
  "_seq_no": 0,
  "_primary_term": 1,
  "_score": 1.0,
  "_source": { ... }
}
```

Note: **no `model_id` field** at the hit level. The `_id` field contains the model ID.

### Impact

This causes downstream clients (e.g., [opensearch-js#1032](https://github.com/opensearch-project/opensearch-js/issues/1032)) to generate incorrect TypeScript types where `hit.model_id` appears valid but is always `undefined` at runtime.

### Root Cause

The `SearchModelTransportAction` in ML Commons simply delegates to OpenSearch's standard search handler (`MLSearchHandler.search()`), which returns a standard `SearchResponse`. No custom field injection occurs.

The `model_id` field correctly exists on the `Source` schema (inside `_source` for model chunks), but it should **not** be on `SearchHitsHit`.

### Suggested Fix

Remove `model_id` from the `SearchHitsHit` schema in `spec/schemas/ml._common.yaml` (around line 75):

```diff
 SearchHitsHit:
   type: object
   properties:
     _version:
       $ref: '_common.yaml#/components/schemas/VersionNumber'
     _seq_no:
       $ref: '_common.yaml#/components/schemas/SequenceNumber'
     _primary_term:
       type: integer
     _index:
       $ref: '_common.yaml#/components/schemas/IndexName'
     _id:
       $ref: '_common.yaml#/components/schemas/Id'
     _score:
       type: ['null', number]
     _source:
       $ref: '#/components/schemas/Source'
-    model_id:
-      $ref: '_common.yaml#/components/schemas/Name'
     sort:
       type: array
       items:
         type: number
         format: float
   required:
     - _score
```

### Verified On

- OpenSearch 3.6.0 (official tarball)
- OpenSearch 2.19 (reported by issue author in opensearch-js#1032)

The field has **never** been returned by the ML search API — this is not a regression but an incorrect spec definition since its introduction.

iprithv:
Removes model_id from the ML SearchHitsHit schema. ML search APIs (/_plugins/_ml/models/_search, /_plugins/_ml/agents/_search, /_plugins/_ml/tasks/_search, /_plugins/_ml/model_groups/_search, etc.) return a standard OpenSearch SearchResponse. The document ID is _id; there is no top-level model_id on the hit.

Fixes #1180.

SearchModelTransportAction (and the other ML search handlers) delegate to MLSearchHandler.search(), which returns a normal SearchResponse. No custom field is injected onto the hit.

The spec still declared model_id as a sibling of _id / _source on SearchHitsHit. That made generated clients (see opensearch-js#1032 (https://github.com/opensearch-project/opensearch-js/issues/1032)) expose hit.model_id, which is always undefined at runtime.

model_id is valid inside _source for some ML documents (e.g. model chunks). That field stays on the Source schema.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -60,6 +60,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Removed unused dependencies: `eslint-config-standard-with-typescript`, `eslint-plugin-import`, `eslint-plugin-n`, `eslint-plugin-promise`, `@eslint/eslintrc`
 
 ### Fixed
+- Removed incorrect `model_id` from ML `SearchHitsHit`; ML search APIs return a standard search hit and the document ID is `_id` ([#1180](https://github.com/opensearch-project/opensearch-api-specification/issues/1180))
 - Fixed `ml.predict_model` and `ml.predict_model_stream` request body schemas to match `RemoteInferenceMLInput` parser: corrected `parameters` type to `Map<String,String>`, removed incorrect `required` constraints, added `action_type`, `dlq`, `question`, and `context` fields, and added `PredictionActionType` enum ([#1188](https://github.com/opensearch-project/opensearch-api-specification/pull/1188))
 - Fixed stale and malformed OpenSearch documentation links in spec `externalDocs` and schema descriptions ([#1163](https://github.com/opensearch-project/opensearch-api-specification/pull/1163))
 - Fixed `DeletedPit` to mark `pit_id` and `successful` as required ([#1146](https://github.com/opensearch-project/opensearch-api-specification/pull/1146))
--- a/spec/schemas/ml._common.yaml
+++ b/spec/schemas/ml._common.yaml
@@ -72,8 +72,6 @@ components:
           description: The score.
         _source:
           $ref: '#/components/schemas/Source'
-        model_id:
-          $ref: '_common.yaml#/components/schemas/Name'
         sort:
           type: array
           items:


xluo-aws (approved):

wanglam (approved):

sean- (approved):

Hailong-am:
@sean- could you help to approve the CI workflow to make it start to run?

PR #1194, opened 2026-08-17
fix(_common): rename ShardFailure fields to match server response

gingeekrishna:
## Description

Rename the `ShardFailure` schema properties in `spec/schemas/_common.yaml` from `index`/`node`/`shard` to `_index`/`_node`/`_shard` to match what `ReplicationResponse.ShardInfo.Failure` on the OpenSearch server actually serializes (underscore-prefixed keys for `_index`, `_shard`, `_node`; no prefix for `reason`, `status`, `primary`).

An earlier version of this PR instead removed `shard` from `required`, on the assumption that the server sometimes omits it. As pointed out in review, the server always writes the field — the real problem was the name mismatch, so clients were looking up `shard` and finding nothing under that key while the actual value sat under `_shard`. This restores `_shard` as `required` and renames all three affected properties.

## Related Issues

Reported via [opensearch-java#1799](https://github.com/opensearch-project/opensearch-java/issues/1799) and the corresponding fix [opensearch-java#2037](https://github.com/opensearch-project/opensearch-java/pull/2037).

## Checklist

- [x] Ran YAML validation locally
- [x] `_shard` restored as `required`, matching server behavior; `_index`/`_node`/`_shard` now match the server's actual JSON keys


spec/schemas/_common.yaml
```diff
@@ -337,7 +337,6 @@ components:
       required:
         - primary
         - reason
-        - shard
```
iprithv, line 340: I looked at the server code, sorry I don't think removing shard from required is the right fix here.

ShardFailure schema is used inside ShardInfo.failures (line 315), which corresponds to ReplicationResponse.ShardInfo.Failure on the server side. that class always writes the shard field, it never omits it:

```
builder.field(_SHARD, shardId.id());
builder.field(_INDEX, shardId.getIndexName());
builder.field(_NODE, nodeId);
```

real problem is a field name mismatch. server writes _shard, _index, _node (with underscores), but this spec defines them as shard, index, node (without underscores). so the Java client looks for shard in the JSON, doesn't find it because the actual key is _shard, and throws MissingRequiredPropertyException.

making shard optional just hides this, client will silently get null instead of the actual shard ID. fix should be correcting the property names to match what the server actually returns.

Can you check the actual JSON response from the server that triggered the original issue? I'd expect the fields are there but named _shard, _index, _node.

iprithv (changes requested):

spec/schemas/_common.yaml
gingeekrishna: Good catch, thanks for checking the server code. You're right, removing `shard` from required just masked the real bug. Fixed in 646c39d: renamed `index`/`node`/`shard` to `_index`/`_node`/`_shard` to match what `ReplicationResponse.ShardInfo.Failure` actually writes, and restored `_shard` as required since the server always includes it. Also updated the PR title/description accordingly.

PR #1195, opened 2026-08-17
fix(_common.analysis): make PathHierarchyTokenizer fields optional

gingeekrishna:
## Description

Remove `buffer_size`, `delimiter`, `reverse`, and `skip` from the `required` list in the `PathHierarchyTokenizer` schema in `spec/schemas/_common.analysis.yaml`.

All four fields are optional in the OpenSearch API with documented defaults (`buffer_size=1024`, `delimiter=/`, `reverse=false`, `skip=0`). Marking them required caused `MissingRequiredPropertyException` when retrieving any index whose `PathHierarchyTokenizer` was configured with defaults, since OpenSearch omits default values from the response.

Only `type` is actually required. The four fields remain defined as properties and continue to be serialized/deserialized when present.

## Related Issues

Reported via [opensearch-java#1797](https://github.com/opensearch-project/opensearch-java/issues/1797) and the corresponding fix [opensearch-java#2038](https://github.com/opensearch-project/opensearch-java/pull/2038), where reviewer @reta requested this upstream spec change.


spec/schemas/_common.analysis.yaml
```diff
@@ -1758,10 +1758,6 @@ components:
             skip:
               $ref: '_common.yaml#/components/schemas/StringifiedInteger'
           required:
-            - buffer_size
```
iprithv, line 1761: this fixes path_hierarchy but AnalysisTokenizerPathHierarchyReverse has the exact same problem, it also marks buffer_size, delimiter, reverse, and skip as required. both types use the same factory class with the same defaults. 

Can you fix that one too in this? otherwise half the bug is still there.

iprithv (changes requested):

spec/schemas/_common.analysis.yaml
gingeekrishna: Thanks for the review! I searched the repo for `AnalysisTokenizerPathHierarchyReverse` (and for any other schema with the same `buffer_size`/`delimiter`/`reverse`/`skip` shape) and couldn't find a second definition anywhere - `PathHierarchyTokenizer` in `_common.analysis.yaml` is the only one, and this PR already drops all four from its `required` list.

I also checked the OpenSearch server source (`CommonAnalysisModulePlugin.java`): it registers `path_hierarchy` and the deprecated camelCase alias `PathHierarchy`, both backed by the same `PathHierarchyTokenizerFactory`. `reverse` is just a boolean setting on that one factory, there's no separate reverse-specific tokenizer type server-side.

Could you point me to the exact file/line (or a link) for `AnalysisTokenizerPathHierarchyReverse`? Want to make sure I fix the right thing rather than guessing.

PR #1196, opened 2026-08-17
fix(ism,flow_framework,sm): use int64 for timestamp fields

gingeekrishna:
## Description

Add `format: int64` to Unix epoch millisecond timestamp fields in the ISM, flow-framework, and snapshot management schemas.

Unix epoch timestamps in milliseconds overflow int32 (~2 billion ms ≈ only 24 days from epoch). Without `format: int64`, generated clients use 32-bit integer types, causing numeric overflow on any real timestamp.

Fields updated:
- `spec/schemas/ism._common.yaml`: `Policy.last_updated_time`, `IsmTemplate.last_updated_time`
- `spec/schemas/flow_framework.common.yaml`: `WorkflowTemplate.created_time`, `WorkflowTemplate.last_updated_time`
- `spec/schemas/sm._common.yaml`: `SnapshotManagementPolicy.last_updated_time`

## Related Issues

Reported via [opensearch-java#1717](https://github.com/opensearch-project/opensearch-java/issues/1717) and [opensearch-java#1898](https://github.com/opensearch-project/opensearch-java/issues/1898), fixed in [opensearch-java#2039](https://github.com/opensearch-project/opensearch-java/pull/2039), where reviewer @reta requested this upstream spec change.


spec/schemas/sm._common.yaml
```diff
@@ -84,6 +84,7 @@ components:
           type: boolean
         last_updated_time:
           type: integer
+          format: int64
           description: When the policy was last modified.
         enabled_time:
```
iprithv, line 89: enabled_time is also an epoch millis timestamp, same as last_updated_time. shouldn't it get format: int64 too? otherwise it'll still overflow in generated clients.

iprithv (changes requested):

spec/schemas/sm._common.yaml
gingeekrishna: Good catch, fixed in c29f1a0 — added `format: int64` to `enabled_time` as well.

PR #1155, opened 2026-07-03
Add mapper_type and mapper_settings to IngestionSource schema

imRishN:
## Summary
- Adds `IngestionSourceMapperType` enum with values `default`, `raw_payload`, `field_mapping`
- Adds `mapper_type` property to `IngestionSource` schema referencing the new enum
- Adds `mapper_settings` property to `IngestionSource` schema as an object with string additionalProperties (for field_mapping config keys: `id_field`, `version_field`, `op_type_field`, `op_type_field.delete_value`, `op_type_field.create_value`)

## Companion Server PRs
- opensearch-project/OpenSearch#20722 — Add `mapper_settings` support (merged)
- opensearch-project/OpenSearch#20729 — `FieldMappingIngestionMessageMapper` implementation (in review)

## Test plan
- [ ] Verify schema validates against a running OpenSearch instance with the server-side changes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1199, opened 2026-08-18
Add decoder_type and decoder_settings to IngestionSource index settings

### LINKED ISSUE #1198: [FEATURE] Add Support for ISM Simulate Policy API
  OpenSearch core recently added a pluggable decoder framework for pull-based (Kafka/Kinesis/file) ingestion, exposed via two new index settings:  index.ingestion_source.decoder_type and index.ingestion_source.decoder_settings.*. These settings aren't reflected in the IngestionSource schema in this repo (spec/schemas/indices._common.yaml), so API consumers/clients generated from this spec have no way to know these settings exist or validate/document them correctly.

Add decoder_type (string, selects the decoder implementation, e.g. xcontent or avro, defaults to xcontent) and decoder_settings (an open-ended object of decoder-specific settings, similar in shape to the existing param property) to the IngestionSource schema, alongside the other pull-based ingestion settings (pointer, poll, all_active, etc.).

### Do you have any additional context?
PR for the changes for add decoder in pull based ingestion: https://github.com/opensearch-project/OpenSearch/pull/22364

Tarun-kishore:
### Description
Add decoder_type and decoder_settings to IngestionSource index settings
PR for the changes in index settings: https://github.com/opensearch-project/OpenSearch/pull/22364

### Issues Resolved
Closes #1198 

### DIFF
--- a/spec/schemas/indices._common.yaml
+++ b/spec/schemas/indices._common.yaml
@@ -213,6 +213,13 @@ components:
         all_active:
           description: Indicates if all-active ingestion is enabled for this index.
           $ref: '_common.yaml#/components/schemas/StringifiedBoolean'
+        decoder_type:
+          type: string
+          description: Selects the decoder used to parse each ingested payload into a document (for example, `xcontent` or `avro`). Defaults to `xcontent`.
+        decoder_settings:
+          type: object
+          additionalProperties: true
+          description: Settings for the decoder selected by `decoder_type`. Supported keys depend on the decoder implementation.
     IngestionSourcePointer:
       type: object
       properties:
--- a/spec/schemas/indices._common.yaml
+++ b/spec/schemas/indices._common.yaml
@@ -182,6 +182,11 @@ components:
         - default
         - field_mapping
         - raw_payload
+    IngestionSourceDecoderType:
+      type: string
+      description: The type of decoder used to parse each ingested payload into a document.
+      enum:
+        - xcontent
     IngestionSource:
       type: object
       description: The pull-based ingestion settings controlling how data is read from streaming sources.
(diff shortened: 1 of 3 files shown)


spec/schemas/indices._common.yaml
```diff
@@ -220,6 +220,13 @@ components:
         all_active:
           description: Indicates if all-active ingestion is enabled for this index.
           $ref: '_common.yaml#/components/schemas/StringifiedBoolean'
+        decoder_type:
```
iprithv, line 228: mapper_type right below uses a $ref to an enum schema (IngestionSourceMapperType) that lists out the known values. shouldn't decoder_type do the same? can we define an IngestionSourceDecoderType enum with at least xcontent in it and reference it here. 
A plain type: string doesn't tell anyone what to actually put in this field.

iprithv, line 228: can we add version '3.8'? these settings don't exist before that version. other fields here already use version where appropriate (total_primary_shards_per_node has x-version-added: '3.0')

spec/schemas/indices._common.yaml
```diff
@@ -220,6 +220,13 @@ components:
...
+          type: string
+          description: Selects the decoder used to parse each ingested payload into a document (for example, `xcontent` or `avro`). Defaults to `xcontent`.
+        decoder_settings:
```
iprithv, line 232: same here, need version added!

iprithv (changes requested):
like in #1155 (tests/plugins/ingestion-fs/pull_based_ingestion/mapper_settings.yaml), can we add similar test that creates an index with decoder_type: xcontent and verifies it comes back in the GET settings response.

spec/schemas/indices._common.yaml
Tarun-kishore: Updated

Tarun-kishore: Added version check for 3.8

spec/schemas/indices._common.yaml
Tarun-kishore: updated version here too

PR #1214, opened 2026-08-25
Add model_selection to SemanticProperty mapping

yizheliu-amazon:
### Description

Adds the `model_selection` object to the `semantic` field mapping (`SemanticProperty`) and makes `model_id` optional.

The `semantic` field now supports resolving its model from cluster settings instead of a hard-coded `model_id`:

```json
{
  "mappings": {
    "properties": {
      "body": {
        "type": "semantic",
        "model_selection": { "language_option": "ENGLISH", "model_type": "SPARSE" }
      }
    }
  }
}
```

`model_type` (`SPARSE`/`DENSE`) + `language_option` (`ENGLISH`/`MULTILINGUAL`) map to the cluster setting
`plugins.neural_search.model_selection.model_id.<model_type>.<language_option>` (lower-cased), whose value is the resolved model id. On read-back the server returns both `model_selection` and the resolved `model_id`, so `model_id` is no longer required on input.

Changes:
- Added `model_selection` (`$ref` to new `SemanticModelSelection`) to `SemanticProperty`, `x-version-added: '3.9'`.
- Added the `SemanticModelSelection` schema (`language_option`, `model_type` enums).
- Removed `model_id` from `SemanticProperty.required` (optional when `model_selection` is supplied).

### Issues Resolved

Backing feature: opensearch-project/neural-search#1918 (implemented in opensearch-project/neural-search#1919).

### Validation

Ran the opensearch-java codegen against this spec change locally: it regenerates `SemanticProperty.java` (adds a nullable `modelSelection`, makes `modelId` nullable) and a new `SemanticModelSelection.java` with `languageOption`/`modelType`. Also verified end-to-end against a local `3.9.0-SNAPSHOT` cluster (create index with `model_selection` → server resolves and injects `model_id` from the cluster setting; ingest + neural query work).

### Check List
- [x] Updated the CHANGELOG.
- [x] All tests pass (spec YAML validated; repo linter/tests run in CI).

### DIFF
--- a/spec/schemas/_common.mapping.yaml
+++ b/spec/schemas/_common.mapping.yaml
@@ -982,6 +982,9 @@ components:
           type: string
         model_id:
           type: string
+        model_selection:
+          $ref: '#/components/schemas/SemanticModelSelection'
+          x-version-added: '3.9'
         search_model_id:
           type: string
         semantic_info_field_name:
@@ -1002,7 +1005,6 @@ components:
           type: boolean
           default: false
       required:
-        - model_id
         - type
     ShortNumberProperty:
       allOf:
@@ -1286,3 +1288,22 @@ components:
           type: string
         prune_ratio:
           type: number
+    SemanticModelSelection:
+      x-version-added: '3.9'
+      type: object
+      description: |-
+        Resolves the model to use for a semantic field from cluster settings instead of a hard-coded `model_id`.
+        The `model_type` and `language_option` pair maps to the cluster setting
+        `plugins.neural_search.model_selection.model_id.<model_type>.<language_option>` (lower-cased),
+        whose value is the resolved model id.
+      properties:
+        language_option:
+          type: string
+          enum:
+            - ENGLISH
+            - MULTILINGUAL
+        model_type:
+          type: string
+          enum:
+            - SPARSE
+            - DENSE
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ### Added
 
+- Added `model_selection` (`language_option`, `model_type`) to `SemanticProperty` and made `model_id` optional so the model id can be resolved from cluster settings ([#1214](https://github.com/opensearch-project/opensearch-api-specification/pull/1214))
 - Added `sparse_encoding`, `text_image_embedding`, and `text_chunking` ingest processor schemas with `ChunkingAlgorithm` (`fixed_token_length`, `delimiter`) support, and `neural_sparse` query DSL ([#1191](https://github.com/opensearch-project/opensearch-api-specification/pull/1191))
 - Added apis for searching within search relevance objects such as search configurations, judgments, query sets, and experiments ([#1064](https://github.com/opensearch-project/opensearch-api-specification/pull/1064))
 - Added remaining APIs for LTR including store element operations, model management, feature set operations, routing support, and POST support for update operations ([#935](https://github.com/opensearch-project/opensearch-api-specification/pull/935))


yizheliu-amazon:
Hi @harshavamsi , could you help review this PR? Thank you

spec/schemas/_common.mapping.yaml
```diff
@@ -1002,7 +1005,6 @@ components:
           type: boolean
           default: false
       required:
-        - model_id
```
iprithv, line 1005: `model_id` being dropped from `required` is the right direction so `model_selection` can replace it, but this is not “both optional” on the server.

In neural-search `SemanticMappingTransformer`, a semantic field with neither `model_id` nor `model_selection` is still rejected (`model_id is required for the semantic field at ...`). `model_id` is optional only when `model_selection` is present. If both are present they must resolve to the same id, which is how GET mapping → PUT mapping round-trips.

As written, this spec accepts `{ "type": "semantic" }`, which a 3.9 cluster rejects, and generated clients will allow constructing that payload.

Please encode “at least one of” with `anyOf` (not `oneOf`, GET mapping returns both the original `model_selection` and the resolved `model_id`):

```yaml
required:
  - type
anyOf:
  - required: [model_id]
  - required: [model_selection]

File: spec/schemas/_common.mapping.yaml [hunk 2]
```diff
@@ -982,6 +982,9 @@ components:
           type: string
         model_id:
           type: string
+        model_selection:
+          $ref: '#/components/schemas/SemanticModelSelection'
+          x-version-added: '3.9'
```
iprithv, line 987: This new field is not covered by spec tests. DEVELOPER_GUIDE.md requires tests for spec changes, and tests/plugins/ml/mapping/semantic_schema_validation.yaml already exists for this mapping.

Please add a chapter gated with version: '>= 3.9' that creates an index using model_selection without model_id:

```
type: semantic
model_selection:
  language_option: ENGLISH
  model_type: SPARSE
```

The spec tester will validate that payload against this schema. The cluster can still return 400 if plugins.neural_search.model_selection.model_id.sparse.english is unset, same pattern as the existing dummy-model_id chapters.

Do not add model_selection to the existing version: '>= 3.1' “all optional properties” chapter; on 3.1–3.8 that key is not a valid mapping parameter.

iprithv (changes requested):

spec/schemas/_common.mapping.yaml
yizheliu-amazon: sure. added

spec/schemas/_common.mapping.yaml
yizheliu-amazon: test added

yizheliu-amazon:
Hi @iprithv and @harshavamsi , could you review it?

yizheliu-amazon:
Hi @Hailong-am , could you also take a look at it? Thanks

tests/plugins/ml/mapping/semantic_schema_validation.yaml
```diff
@@ -81,3 +81,24 @@ chapters:
...
+        mappings:
+          properties:
+            selected:
+              type: semantic
+              model_selection:
+                language_option: ENGLISH
+                model_type: SPARSE
+    response:
+      status: [200, 400, 500]
+
```
Hailong-am, line 104: The test didn't actual run as the CI only run with OpenSearch 3.1, but it's good that we have a test.

Hailong-am (approved):

tests/plugins/ml/mapping/semantic_schema_validation.yaml
yizheliu-amazon: Right — it's gated `version: '>= 3.9'` so it's skipped on the 3.1 CI matrix and will start running once a 3.9+ distribution is added to `test-spec.yml`. Keeping it in now so the coverage lands with the feature version. Thanks for confirming the schema validation itself passes.

harshavamsi (dismissed):

PR #1226, opened 2026-09-03
ci: build coverage cluster on OpenSearch 3.8 from vendored plugin

Hailong-am:
## Problem / Motivation

The spec-coverage pipeline (`coverage.yml`, `analyze-pr-changes.yml`) dumps the live cluster's REST API surface using the `opensearch-api` plugin, installed in `coverage/Dockerfile` from a **release asset** of [dblock/opensearch-api](https://github.com/dblock/opensearch-api). That repo publishes exactly one release — `v2.12.0` — so `coverage/Dockerfile` is pinned to `OPENSEARCH_VERSION=2.12.0`. There is no 3.x release asset, so coverage cannot be run against any OpenSearch 3.x line, including the current 3.8.0.

## Why it matters

Coverage is the authoritative signal for spec gaps (the "Add Missing API Specs" checklist). Pinned to 2.12.0, it measures the spec against a cluster three major-minor lines behind current OpenSearch, so APIs added since 2.12 are invisible to the gap analysis.

## What changed (motivation -> approach -> change)

- **Root cause:** the pipeline depends on an *external release artifact* that upstream no longer cuts for 3.x.
- **Approach:** remove the external dependency by building the plugin from source in-tree. The plugin source already defaults to a 3.x line (`opensearch_version = "3.8.0"`), so it is effectively 3.x-ready — only a published artifact was missing.
- **Change:**
  - Vendored the plugin's buildable `main` source (Java + Gradle build + wrapper, no test suites) under `coverage/opensearch-api-plugin/`.
  - Rewrote `coverage/Dockerfile` as a two-stage build: stage 1 compiles the plugin with `-Dopensearch.version=${OPENSEARCH_VERSION} -Dbuild.snapshot=false`; stage 2 installs the resulting zip into `opensearchproject/opensearch:${OPENSEARCH_VERSION}`. Default `OPENSEARCH_VERSION=3.8.0`.
  - Narrowed the over-broad `/coverage/` `.gitignore` rule (the coverage tool writes to `build/`, not `coverage/`) so vendored source is tracked while plugin build artifacts (`build/`, `.gradle/`) stay ignored.
  - Added `coverage/**` to the `coverage.yml` triggers so plugin/Dockerfile edits re-run coverage.

Future version bumps are a one-line `OPENSEARCH_VERSION` arg change — no plugin edit needed.

## Tests

No automated tests (this is CI/build infrastructure). Verified by building and running the image locally — see Manual verification.

## Manual verification

Built `coverage/Dockerfile` locally with Docker:
- Plugin compiled cleanly against 3.8.0 -> `opensearch-api-3.8.0.0.zip`.
- `opensearch-plugin install` reported `-> Installed opensearch-api`.
- Container came up on OpenSearch `3.8.0`; `GET /_plugins/api` returned a valid OpenAPI 3.0.1 document (`"version":"3.8.0"`) listing the cluster's real REST paths (`/_filecache/prune`, `/_mapping`, …) — the exact endpoint `dump-cluster-spec` calls.

**Why no screenshot:** CI/build-only change (Dockerfile, vendored Java plugin, gitignore, workflow trigger); no user-visible UI.

## Open question for maintainers (needs a decision)

This vendors a compiled Java plugin into a spec repo. The alternative is cutting a proper 3.x release on the plugin's own repo and keeping `coverage/Dockerfile` on a download. I went with vendoring because it removes the external point of failure entirely, but whether the plugin should *live here* is a repo-character call I'd like your ruling on before this merges.

no linked issue: infra change surfaced while bumping coverage to 3.8


Hailong-am:
**Mend Security Check — rebutted (pre-existing, not introduced by this PR).**

All 6 vulnerabilities Mend flags are in the repo's existing **npm** tree under `/package.json`:

- `fast-uri` (transitive via `ajv`) — CVE-2026-76172 / -75975 / -75931 / -75899, already tracked as #1167.
- `qs` (direct) — CVE-2026-82417 / -82562, from the repo's existing `qs` version.

This PR changes **no npm dependencies** — `git diff origin/main...HEAD` touches neither `package.json` nor `package-lock.json`. The only `package.json` this PR adds is the vendored plugin's OpenSearch metadata file (`coverage/opensearch-api-plugin/src/main/java/org/opensearch/plugin/api/action/api/package.json`), which is not an npm manifest and carries no dependencies. Mend's own report confirms these are pre-existing ("Base branch total remaining vulnerabilities: 28").

Bumping `fast-uri`/`qs` here would be unrelated dependency work outside this PR's scope (a CI/build change), so I'm leaving the diff as-is. The npm CVEs should be resolved on their own track (e.g. #1167), not folded into this PR.

PR #1238, opened 2026-09-13
Add output_format to the date ingest processor schema

elgamal-ahmed:
### Description
Adds the `output_format` property to the `DateProcessor` schema. This is the format used when writing the parsed date to `target_field`. It is a supported option of the OpenSearch date ingest processor but was missing from the spec, so it never generated into the language clients.

### Issues Resolved
Surfaced while adding this option to the Java client (opensearch-project/opensearch-java#2150), which was pointed back here since the spec is the source of truth.

### By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.


elgamal-ahmed:
@reta following up from opensearch-project/opensearch-java#2150 (now closed), this adds `output_format` to the DateProcessor schema upstream as you suggested.

reta (approved):

Hailong-am (approved):

PR #1239, opened 2026-09-13
Make opensearch_version optional in PluginStats

elgamal-ahmed:
@reta following up from opensearch-project/opensearch-java#2151 (now closed), this makes `opensearch_version` optional in the PluginStats schema upstream as you suggested.


spec/schemas/_common.yaml
```diff
@@ -1735,7 +1735,6 @@ components:
         - has_native_controller
         - java_version
         - name
-        - opensearch_version
```
reta, line 1738: Just to clarify on that, `opensearch_version` is, in general, required field, but it was introduced in 2.14 (f I am not mistaken), so it is not present in the older versions.

reta (approved):

spec/schemas/_common.yaml
Hailong-am: check the code https://github.com/opensearch-project/OpenSearch/blob/1.0/server/src/main/java/org/opensearch/plugins/PluginInfo.java#L332 it presents from 1.0 @elgamal-ahmed which version are you used?

PR #1245, opened 2026-09-20
Add skip status to the knn.warmup response schema

clearlove10-c:
The k-NN warmup API now reports shards whose warmup was skipped: `_shards.skipped` is already covered by `ShardStatistics`, so extend the `knn.warmup` response with an optional `_shards.skip_reasons` array (`warm_tier_index`, `lucene_engine`). Update the warmup test story with a Lucene-engine index whose warmup is skipped.
### Description
The k-NN warmup API (`GET /_plugins/_knn/warmup/{index}`) now reports shards whose warmup was silently skipped (warm-tier indices, indices whose k-NN fields all use the Lucene engine). See the companion k-NN PR for the implementation.
Spec changes:
- `_shards.skipped` requires no schema change: the shared `ShardStatistics` already defines it (optional integer).
- `knn.warmup@200` is extended with `allOf` to add an optional `_shards.skip_reasons` string array, returned only when at least one shard was skipped. Supported values: `warm_tier_index`, `lucene_engine`.
Test story changes:
- Adds `version: '>= 3.9.0'` (the skip status ships with k-NN 3.9.0).
- Adds a Lucene-engine index to the prologues and a chapter asserting the skipped response (`_shards.skipped: 1`, `skip_reasons: ["lucene_engine"]`).
### Issues Resolved
Companion to opensearch-project/k-NN#3592 (feature request: opensearch-project/k-NN#3569)


Hailong-am (approved):
Merge after upstream change get merged.

tests/plugins/knn/knn/warmup.yaml
```diff
@@ -1,6 +1,7 @@
 $schema: ../../../../json_schemas/test_story.schema.yaml
 
 description: Load all the native library files for all the shards into native memory.
+version: '>= 3.9.0'
```
Hailong-am, line 4: Will this include in 3.9?

clearlove10-c: It depends on the feedback from the upstream PR.

